PR #28, opened 2021-05-28
Update sbt command in docs

skhatri:
Signed-off-by: Suresh Khatri <servlet2@msn.com>


zsxwing:
Thanks for your contribution. Could you describe your environment? Generally, `build/sbt` and `./build/sbt` should be the same.

zsxwing:
Closing as there is no response for a while.

PR #30, opened 2021-05-31
Fix AddFile to accept JSON which have no stats

dobachi:
According to [PROTOCOL](https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md)  definition, "stats" field of "File" is an optional field.
Therefore, "AddFile" and "from_json" method should accept JSON which have no "stats".

Signed-off-by: Masaru Dobashi <dobachi1983oss@gmail.com>

### DIFF
--- a/python/delta_sharing/protocol.py
+++ b/python/delta_sharing/protocol.py
@@ -160,7 +160,7 @@ class AddFile:
     id: str
     partition_values: Dict[str, str]
     size: int
-    stats: str = None
+    stats: Optional[str] = None
 
     @staticmethod
     def from_json(json) -> "AddFile":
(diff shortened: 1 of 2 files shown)


python/delta_sharing/protocol.py
```diff
@@ -160,7 +160,7 @@ class AddFile:
     id: str
     partition_values: Dict[str, str]
     size: int
-    stats: str
+    stats: str = None
```
ueshin, line 163: The type hint should be `Optional[str]`?

dobachi: Thank you! I fix it.

ueshin (approved):
LGTM. cc @zsxwing

ueshin:
Thanks! merged.

PR #32, opened 2021-05-31
Improve error handling in Delta Sharing Server

### LINKED ISSUE #27: The query endpoint should return a 400 on invalid input
Let's just say, hypothetically speaking some idiot were to pass the wrong type of JSON object for `predicateHints`, the API will 500.

```
curl -X POST "https://sharing.delta.io/delta-sharing/shares/delta_sharing/schemas/default/tables/COVID_19_NYT/query" -H  "accept: application/x-ndjson" -H  "Authorization: Bearer faaie590d541265bcab1f2de9813274bf233" -H  "Content-Type: a
pplication/json" -d "{\"predicateHints\":{},\"limitHint\":1000}"
```

Since I'm the idiot, I should be given a 400 error which means that the request was incorrectly formed.

zsxwing:
Currently Armeria is doing the request parsing work. When a request is invalid, such as invalid json (#27), or incorrect type (`maxResults` is not an int), it will throw an exception and report 500 Internal Error.

In this PR, we catch any unhandled exception when process a request, and convert it to `DeltaInternalException` when processing a request. Then we scope all exceptions that may contain sensitive information to a single type `DeltaInternalException`. Hence, `DeltaSharingServiceExceptionHandler` can safely expose the messages in the following exceptions to the user for better user experience and report 400 Bad Request:
- `scalapb.json4s.JsonFormatException`: Valid json string but invalid field type (#27).
- `com.fasterxml.jackson.databind.JsonMappingException`: Invalid json string.
- `java.lang.NumberFormatException`: `maxResults` is not an int.

Fixes #27

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -62,14 +62,17 @@ class DeltaSharingServiceExceptionHandler extends ExceptionHandlerFunction {
       // happens before `DeltaSharingService` receives the requests so these exceptions should never
       // contain sensitive information and should be okay to return their messages to the user.
       case _: scalapb.json4s.JsonFormatException =>
+        // valid json but may be incorrect field type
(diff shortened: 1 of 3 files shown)


ueshin (approved):
LGTM.

PR #36, opened 2021-06-05
Add more test tables and update integration tests

zsxwing:
Add two more tables and update existing integration tests accordingly:

- `share3.default.table4`: The table column order is not the same as parquet files.
- `share3.default.table5`: An empty table.


zsxwing:
@FX196 is fixing the empty table issue. I'm adding these tables so that he can write end-to-end tests for empty table.

FX196:
Thanks @zsxwing ! I was going to write the empty tables test myself

ueshin (approved):
LGTM.

PR #37, opened 2021-06-07
Add columns to empty table in load_as_pandas and fix issues for complex types

### LINKED ISSUE #35: load_as_pandas should return a DataFrame with correct schema for an empty table
When a table is empty, `load_as_pandas` will return a DataFrame without schema. We should create an empty DataFrame with the table schema instead. See: https://github.com/delta-io/delta-sharing/blob/v0.1.0/python/delta_sharing/reader.py#L65

FX196:
This PR fixes two issues:
- When a table is empty, `load_as_pandas` should return an empty Pandas DataFrame with the table schema. Fixes #35 
- When the type of a data column is a complex type, `load_as_pandas` should not fail. In a Delta table, only partition columns are not allowed to use complex types.

### DIFF
--- a/python/delta_sharing/converter.py
+++ b/python/delta_sharing/converter.py
@@ -21,6 +21,12 @@ import pandas as pd
 
 
 def _get_dummy_column(schema_type):
+    """
+    Return a dummy column with the data type specified in schema_type.
+    The dummy column is used to populate the dtype fields in empty tables.
+    :param schema_type: str or json representing a data type
+    :return: dummy pandas Series to be inserted into an empty table
+    """
     if schema_type == "boolean":
         return pd.Series([False])
     elif schema_type == "byte":
@@ -52,6 +58,13 @@ def _get_dummy_column(schema_type):
 
 
 def get_empty_table(schema_json: dict) -> pd.DataFrame:
+    """
+    For empty tables, we use dummy columns from `_get_dummy_column` and then
+    drop all rows to generate a table with the correct column names and
+    data types.
+    :param schema_json: json object representing the table schema
+    :return: empty table with columns specified in schema_json
+    """
     assert schema_json["type"] == "struct"
 
     dummy_table = pd.DataFrame(
@@ -67,6 +80,13 @@ def to_converters(schema_json: dict) -> Dict[str, Callable[[str], Any]]:
 
 
 def to_converter(schema_type) -> Callable[[str], Any]:
+    """
+    For types that support partitioning, a lambda to parse data into the
+    corresponding type is returned. For data types that cannot be partitioned
+    on, we return None. The caller is expected to check if the value is None before using.
+    :param schema_type: str or json representing a data type
+    :return: converter function or None
+    """
     if schema_type == "boolean":
         return lambda x: None if (x is None or x == "") else (x is True or x == "true")
     elif schema_type == "byte":
--- a/python/delta_sharing/converter.py
+++ b/python/delta_sharing/converter.py
@@ -50,9 +50,9 @@ def get_empty_table(schema_string: str) -> pd.DataFrame:
     schema_json = loads(schema_string)
     assert schema_json["type"] == "struct"
 
-    dummy_table = pd.DataFrame({field["name"]:
-                                _get_dummy_column(field["type"])
-                                for field in schema_json["fields"]})
+    dummy_table = pd.DataFrame(
+        {field["name"]: _get_dummy_column(field["type"]) for field in schema_json["fields"]}
+    )
     return dummy_table.iloc[0:0]
 
 
--- a/python/delta_sharing/converter.py
+++ b/python/delta_sharing/converter.py
@@ -21,31 +21,35 @@ import numpy as np
 import pandas as pd
 
 
-def _get_dummy_column(type_string: str):
-    if type_string == "boolean":
+def _get_dummy_column(schema_type):
+    if schema_type == "boolean":
         return pd.Series([False])
-    elif type_string == "byte":
+    elif schema_type == "byte":
         return pd.Series([0], dtype="int8")
-    elif type_string == "short":
+    elif schema_type == "short":
         return pd.Series([0], dtype="int16")
-    elif type_string == "integer":
+    elif schema_type == "integer":
         return pd.Series([0], dtype="int32")
-    elif type_string == "long":
+    elif schema_type == "long":
         return pd.Series([0], dtype="int64")
-    elif type_string == "float":
+    elif schema_type == "float":
         return pd.Series([0], dtype="float32")
-    elif type_string == "double":
+    elif schema_type == "double":
         return pd.Series([0], dtype="float64")
-    elif type_string.startswith("decimal"):
-        return pd.Series([Decimal("1.2")])
-    elif type_string == "string":
-        return pd.Series(["dummy"])
-    elif type_string == "date":
+    elif isinstance(schema_type, str) and schema_type.startswith("decimal"):
+        return pd.Series([0], dtype=np.dtype("O"))
+    elif schema_type == "string":
+        return pd.Series([0], dtype=np.dtype("O"))
+    elif schema_type == "date":
         return pd.Series([pd.Timestamp(0).date()])
-    elif type_string == "timestamp":
-        return pd.Series([pd.Timestamp(0)])
+    elif schema_type == "timestamp":
+        return pd.Series([pd.Timestamp(0).date()], dtype=np.dtype("datetime64[ns]"))
+    elif schema_type == "binary":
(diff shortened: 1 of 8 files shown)


python/delta_sharing/converter.py
```diff
@@ -21,6 +21,41 @@
...
+        return pd.Series([0], dtype="int16")
+    elif type_string == "integer":
+        return pd.Series([0], dtype="int32")
+    elif type_string == "long":
+        return pd.Series([0], dtype="int64")
+    elif type_string == "float":
+        return pd.Series([0], dtype="float32")
+    elif type_string == "double":
+        return pd.Series([0], dtype="float64")
+    elif isinstance(type_string, str) and type_string.startswith("decimal"):
```
ueshin, line 39: nit: Do we need to check `isinstance(type_string, str)`?

python/delta_sharing/converter.py
```diff
@@ -21,6 +21,41 @@
...
+        return pd.Series([Decimal("1.2")])
+    elif type_string == "string":
+        return pd.Series(["dummy"], dtype="string")
+    elif type_string == "date":
+        return pd.Series([pd.Timestamp(0).date()], dtype="datetime64[ns]")
+    elif type_string == "timestamp":
+        return pd.Series([pd.Timestamp(0)], dtype="datetime64[ns]")
+
```
ueshin, line 47: I guess we should return something like `pd.Series([None], dtype='object')` for other data tyeps?

python/delta_sharing/converter.py
```diff
@@ -21,6 +21,41 @@
...
+    elif type_string == "long":
+        return pd.Series([0], dtype="int64")
+    elif type_string == "float":
+        return pd.Series([0], dtype="float32")
+    elif type_string == "double":
+        return pd.Series([0], dtype="float64")
+    elif isinstance(type_string, str) and type_string.startswith("decimal"):
+        return pd.Series([Decimal("1.2")])
+    elif type_string == "string":
+        return pd.Series(["dummy"], dtype="string")
```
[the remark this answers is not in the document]
FX196: I thought it would be better to use the more specific dtype instead of the generic `object`. Perhaps we should get @zsxwing 's opinion?

python/delta_sharing/converter.py
```diff
@@ -21,6 +21,41 @@
...
+    elif type_string == "date":
+        return pd.Series([pd.Timestamp(0).date()], dtype="datetime64[ns]")
```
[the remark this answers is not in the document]
FX196: Kind of like the string case, `pd.Timestamp(0).date()` is treated as `object` unless we explicitly request pandas to use `datetime64[ns]`. pandas seems to cast all `datetime64` objects to `datetime64[ns]`, so I used it for both `date` and `timestamp`.

python/delta_sharing/converter.py
FX196: Ryan used this syntax for `to_converter` so I just ran with it. Slacked Ryan about this just now.

python/delta_sharing/converter.py
FX196: Ryan said we only need to support the dtypes that Delta Lake supports. But it probably won't hurt to add a case to default to.

python/delta_sharing/converter.py
FX196: Checked with Ryan, no need to check for `isinstance`

python/delta_sharing/converter.py
FX196: I added a ValueError at the end so the behavior is consistent with `to_converter`

python/delta_sharing/converter.py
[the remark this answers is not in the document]
FX196: Checked with Ryan, going with default behavior (`object`), likewise for DateType

python/delta_sharing/converter.py
```diff
@@ -14,44 +14,86 @@
...
-from json import loads
 from typing import Any, Callable, Dict
 
 import numpy as np
 import pandas as pd
 
 
-def to_converters(schema_string: str) -> Dict[str, Callable[[str], Any]]:
-    schema_json = loads(schema_string)
+def _get_dummy_column(schema_type):
```
zsxwing, line 23: nit: could you add a comment to explain what this method does.

python/delta_sharing/converter.py
```diff
@@ -14,44 +14,86 @@
...
+def get_empty_table(schema_json: dict) -> pd.DataFrame:
+    assert schema_json["type"] == "struct"
+
+    dummy_table = pd.DataFrame(
+        {field["name"]: _get_dummy_column(field["type"]) for field in schema_json["fields"]}
+    )
+    return dummy_table.iloc[0:0]
+
+
+def to_converters(schema_json: dict) -> Dict[str, Callable[[str], Any]]:
```
zsxwing, line 63: nit: could you add a comment to explain what this does and the meaning of the return value.

python/delta_sharing/converter.py
```diff
@@ -14,44 +14,86 @@
...
     assert schema_json["type"] == "struct"
 
     return {field["name"]: to_converter(field["type"]) for field in schema_json["fields"]}
 
 
-def to_converter(json) -> Callable[[str], Any]:
-    if json == "boolean":
+def to_converter(schema_type) -> Callable[[str], Any]:
```
zsxwing, line 69: Nit: missing comment and it's better to document the return type since it's not obvious what it means.

zsxwing:
@ueshin could you take another look?

python/delta_sharing/tests/test_reader.py
```diff
@@ -183,3 +186,54 @@ def list_files_in_table(
...
+            add_files: Sequence[AddFile] = []
+            return ListFilesInTableResponse(protocol=None, metadata=metadata, add_files=add_files)
+
+    reader = DeltaSharingReader(
+        Table("table_name", "share_name", "schema_name"), RestClientMock()  # type: ignore
+    )
+    pdf = reader.to_pandas()
+
+    reader = DeltaSharingReader(Table(name="table7", share="share1", schema="default"), rest_client)
+    expected = reader.to_pandas().iloc[0:0]
```
ueshin, line 237: Shouldn't the `expected` be created in local? Otherwise, this is trying to test `DeltaSharingReader` using `DeltaSharingReader`. I think it's weird.

The integration test should be at `test_load` in `test_delta_sharing.py`.

FX196: The objective was to make sure the empty table schema is consistent with the non-empty ones, so I added a reference table to compare with. If we make our own reference schema, we can't be sure our reference schema matches the real one when a table is actually read, hence we're using `DeltaSharingReader`.

Although we're using the class to test itself, we're using the non-empty table behavior to test the empty table behavior, so it's a different logical path.

FX196: I can make a local copy of the parquet file for the table so that this test isn't an integration test, but it's difficult to be sure we have the correct schema without actually reading a table.

python/delta_sharing/converter.py
```diff
@@ -14,44 +14,106 @@
...
         return lambda x: None if (x is None or x == "") else pd.Timestamp(x).date()
-    elif json == "timestamp":
+    elif schema_type == "timestamp":
         return lambda x: pd.NaT if (x is None or x == "") else pd.Timestamp(x)
+    elif schema_type == "binary":
+        return None  # partition on binary column not supported
+    elif isinstance(schema_type, dict) and schema_type["type"] in ("array", "struct", "map"):
+        return None  # partition on complex column not supported
 
     # TODO: binary
```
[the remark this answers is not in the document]
FX196: Removing these will actually break a lot of other stuff since other parts of the code use the keys of `converters` to iterate over the columns of the table. We change things to use `schema_json`, but it'll be a lot of work and Ryan said it wasn't worth it.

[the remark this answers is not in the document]
ueshin: I mean the comment `# TODO: binary`.

python/delta_sharing/tests/test_reader.py
ueshin: I see. I'm still feeling weird, though.
Anyway the behavior looks fine if the `table7` has a proper schema.
I'd leave it to @zsxwing for this.

ueshin (approved):
Otherwise, LGTM.

python/delta_sharing/tests/test_reader.py
zsxwing: I'm okay with this in the short term. Eventually we should make the server support local file system and use all of these tables locally.

zsxwing (approved):

PR #41, opened 2021-06-22
Check whether a table is a Delta table when querying table version

FX196:


server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -111,10 +121,7 @@ class DeltaSharedTable(
       limitHint: Option[Int]): (Long, Seq[model.SingleAction]) = withClassLoader {
     // TODO Support `limitHint`
     val snapshot = deltaLog.snapshot
-    if (snapshot.version < 0) {
-      throw new IllegalStateException(s"The table ${tableConfig.getName} " +
-        s"doesn't exist on the file system or is not a Delta table")
-    }
+    validateDeltaTable()
```
zsxwing, line 124: nit: Could you pass the snapshot to `validateDeltaTable`. It's better to use the same snapshot in a single request.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -100,8 +100,18 @@ class DeltaSharedTable(
     }
   }
 
+  /** Check if the table version in deltalog is valid */
+  def validateDeltaTable(): Unit = {
```
zsxwing, line 104: nit: add `private`

zsxwing (changes requested):
Left some nits. Otherwise, LGTM. Thanks!

zsxwing (approved):
LGTM

PR #45, opened 2021-07-03
Add conf/ to server classpath

zsxwing:
Add conf/ to server classpath so that users can put their hadoop configuration files in conf/ directly.

Manually checked the generated script and verified `conf/` is on the classpath.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -59,6 +59,7 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
   commonSettings,
   scalaStyleSettings,
   releaseSettings,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -59,6 +59,7 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
   commonSettings,
   scalaStyleSettings,
   releaseSettings,
+  scriptClasspath ++= Seq("../conf"),
```
zsxwing, line 62: This should be a path relative to `lib/`.

ueshin (approved):
LGTM.

PR #46, opened 2021-07-09
Dockerize Delta Sharing Server

### LINKED ISSUE #24: Dockerize Delta Sharing Server
We can dockerize Delta Sharing Server to simplify the server setup work.

FX196:
Closes #24

Allows the docker image to be built by running `build/sbt server/docker:publishLocal`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -118,18 +118,19 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
     scalapb.gen() -> (Compile / sourceManaged).value / "scalapb"
   ),
   daemonUser in Docker := "root",
+  daemonUserUid in Docker := None,
   // insert setcap to before entrypoint
   dockerCommands := {
     val commandLen = dockerCommands.value.length
-    dockerCommands.value.take(commandLen-2) ++ Seq(Cmd(
-      "RUN",
-      "setcap", """cap_net_bind_service+ep""",
-      "$(readlink -f /opt/docker/bin/delta-sharing-server)")) ++ dockerCommands.value.drop(commandLen-2)
-  },
-  // Remove command that changes USER to 1001:0
-  dockerCommands := dockerCommands.value.filterNot {
-    case Cmd("USER", args @ _*) => args.contains("1001:0")
-    case other => false
+    (dockerCommands.value.take(commandLen-2)
+      ++ Seq(Cmd(
+                "RUN",
+                "setcap", """cap_net_bind_service=+eip""",
+              "$(readlink -f /opt/docker/bin/delta-sharing-server)"),
+              Cmd("RUN", "mkdir /config", "&&", "touch",
+                "/config/delta-sharing-server-config.yaml")
+            )
+      ++ dockerCommands.value.drop(commandLen-2))
   }
 )
 
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,6 @@
  */
 
 import sbt.ExclusionRule
-import com.typesafe.sbt.packager.docker._
 
 ThisBuild / parallelExecution := false
 
@@ -116,8 +115,7 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
   ),
   Compile / PB.targets := Seq(
     scalapb.gen() -> (Compile / sourceManaged).value / "scalapb"
-  ),
-  daemonUserUid in Docker := None
+  )
 )
 
 /*
--- a/build.sbt
+++ b/build.sbt
@@ -15,6 +15,7 @@
  */
 
 import sbt.ExclusionRule
+import com.typesafe.sbt.packager.docker._
 
 ThisBuild / parallelExecution := false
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -115,7 +116,21 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
   ),
   Compile / PB.targets := Seq(
     scalapb.gen() -> (Compile / sourceManaged).value / "scalapb"
-  )
+  ),
+  daemonUser in Docker := "root",
```
krisgeus, line 120: The docker:publishLocal outputs a fine docker image that doesn't run as root. I agree with the enhancements in the documentation but don't understand the need to change the user, and add the setcap. Could you please clarify why the defaults don't work for you?

FX196: setcap was needed so that the server would be able to use privileged ports like 80 inside the container. But you're right that we don't need to keep the user as root after setcap.

build.sbt
```diff
@@ -115,7 +116,21 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
...
+  // insert setcap to before entrypoint
+  dockerCommands := {
+    val commandLen = dockerCommands.value.length
+    dockerCommands.value.take(commandLen-2) ++ Seq(Cmd(
+      "RUN",
+      "setcap", """cap_net_bind_service+ep""",
```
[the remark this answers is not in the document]
FX196: root is not needed after setcap, thanks for the catch!

FX196:
Turns out we still get permission denied with port 80 and a non-root user even after the setcap, so we'll just keep the user as root.

build.sbt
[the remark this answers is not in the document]
FX196: Turns out I didn't rebuild properly after I remove root, and we still need to be root for port 80

build.sbt
krisgeus: A clear. I chose port 8000 in my config file so the current docker container runs fine for me. We could do without the setcap if we make sure the default config file in the repo has a different port (not 80). That would be my preference to keep things as simple as possible. Just my 2cts

build.sbt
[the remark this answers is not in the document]
krisgeus: I chose port 8000 in my config file so the current docker container runs fine for me. We could do without the setcap if we make sure the default config file in the repo has a different port (not 80). That would be my preference to keep things as simple as possible. Just my 2cts

build.sbt
```diff
@@ -115,7 +116,22 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
...
+  ),
+  daemonUser in Docker := "root",
+  daemonUserUid in Docker := None,
+  // insert setcap to before entrypoint
+  dockerCommands := {
+    val commandLen = dockerCommands.value.length
+    (dockerCommands.value.take(commandLen-2)
+      ++ Seq(Cmd(
+                "RUN",
+                "setcap", """cap_net_bind_service=+eip""",
```
zsxwing, line 128: I found that `setcap` actually didn't work for shell scripts. The reason your change is working is because we switch to the root user (Removing setcap and running using root still works). Looks like what we need is just:

```
daemonUser in Docker := "root",
daemonUserUid in Docker := None
```

build.sbt
```diff
@@ -115,7 +116,22 @@ lazy val server = (project in file("server")) enablePlugins(JavaAppPackaging) se
...
+              "$(readlink -f /opt/docker/bin/delta-sharing-server)"),
+              Cmd("RUN", "mkdir /config", "&&", "touch",
```
zsxwing, line 130: We don't need to create this since we have asked the user to mount it.

build.sbt
zsxwing: @krisgeus Good point. I just did some research about [the potential risks](https://medium.com/jobteaser-dev-team/docker-user-best-practices-a8d2ca5205f4) of using root in a container. Totally agree that we should not use root.

@FX196 could you just change the server config template to use 8080 instead?

FX196:
@krisgeus @zsxwing Sounds good, I'll edit the template and remove the root and setcap.

build.sbt
FX196: Sorry, I thought I read on stackoverflow that the file needed to exist inside the container or else it will be created as a directory, but I just found out that it will only be created as a directory if it didn't exist on the host.

server/src/universal/conf/delta-sharing-server.yaml.template
```diff
@@ -18,8 +18,9 @@ shares:
       location: "s3a://<bucket-name>/<the-table-path>"
 # Set the host name that the server will use
 host: "localhost"
-# Set the port that the server will listen on
-port: 80
+# Set the port that the server will listen on, default is 8080 because
```
zsxwing, line 21: nit: our default value used in code is actually `80.

```
# Set the port that the server will listen on. Note: using ports below 1024 
# may require a privileged user in some operator systems.
```

FX196: I changed that just now in `ServerConfig.scala`. Maybe you didn't refresh?

server/src/main/scala/io/delta/sharing/server/config/ServerConfig.scala
```diff
@@ -65,7 +65,7 @@ case class ServerConfig(
       authorization = null,
       ssl = null,
       host = "localhost",
-      port = 80,
+      port = 8080,
```
zsxwing, line 68: Changing the default value would be a breaking change for people using a config without the port config. It's better to not change this.

FX196: I see. Do we need to say anything in the config template?

FX196: nvm, saw the comment above

server/src/universal/conf/delta-sharing-server.yaml.template
```diff
@@ -18,8 +18,9 @@ shares:
       location: "s3a://<bucket-name>/<the-table-path>"
 # Set the host name that the server will use
 host: "localhost"
-# Set the port that the server will listen on
-port: 80
+# Set the port that the server will listen on. Note: using ports below 1024 
+# may require a privileged user in some operator systems.
```
zsxwing, line 22: Sorry. I had a typo in my previous comment.
```suggestion
# may require a privileged user in some operating systems.
```

zsxwing (approved):
LGTM

PR #48, opened 2021-07-14
Test for non-default compression codec

FX196:
- Added a new table compressed with gzip to the test bucket.
- Added tests to make sure the client is able to read the table.

### DIFF
--- a/python/delta_sharing/tests/test_reader.py
+++ b/python/delta_sharing/tests/test_reader.py
@@ -238,8 +238,11 @@ def test_to_pandas_empty(rest_client: DataSharingRestClient):
 
     pd.testing.assert_frame_equal(pdf, expected)
 
+
 @pytest.mark.skipif(not ENABLE_INTEGRATION, reason=SKIP_MESSAGE)
 def test_load_gzip_table(rest_client: DataSharingRestClient):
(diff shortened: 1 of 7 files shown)


python/delta_sharing/tests/test_reader.py
```diff
@@ -237,3 +237,12 @@ def list_files_in_table(
     expected = reader.to_pandas().iloc[0:0]
 
     pd.testing.assert_frame_equal(pdf, expected)
+
+
+@pytest.mark.skipif(not ENABLE_INTEGRATION, reason=SKIP_MESSAGE)
+def test_load_gzip_table(rest_client: DataSharingRestClient):
```
zsxwing, line 243: It's better to add a test in `https://github.com/delta-io/delta-sharing/blob/main/python/delta_sharing/tests/test_delta_sharing.py#L112` instead.

zsxwing (changes requested):
Could you also add a test in https://github.com/delta-io/delta-sharing/blob/main/spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSuite.scala#L32 as well?

python/delta_sharing/tests/test_reader.py
FX196: Sure, I thought I'd put it here because it was more related to the reader.

zsxwing (approved):
LGTM

PR #49, opened 2021-07-20
Add retry with exponential backoff for Python client

FX196:
Added a retry function for DataSharingRestClient.
Added unit tests.

### DIFF
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -172,6 +172,10 @@ class DataSharingRestClient:
     def _post_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
         return self._request_internal(request=self._session.post, target=target, data=data)
 
+    def _request_internal_with_backoff(self, request, target: str, data: Optional[Dict[str, Any]]):
+        return self._retry_with_exponential_backoff(
+            lambda :self._request_internal(request=request, target=target, data=data))
+
     @contextmanager
     def _request_internal(self, request, target: str, data: Optional[Dict[str, Any]]):
         assert target.startswith("/"), "Targets should start with '/'"
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -167,18 +167,13 @@ class DataSharingRestClient:
         self._session.close()
 
     def _get_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
-        return self._request_internal_with_backoff(
+        return self._request_internal(
             request=self._session.get, target=target, data=data
         )
 
     def _post_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
-        return self._request_internal_with_backoff(
+        return self._request_internal(
             request=self._session.post, target=target, data=data
-        )
-
-    def _request_internal_with_backoff(self, request, target: str, data: Optional[Dict[str, Any]]):
-        return self._retry_with_exponential_backoff(
-            lambda: self._request_internal(request=request, target=target, data=data)
         )
 
     @contextmanager
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -213,13 +213,13 @@ class DataSharingRestClient:
     def _should_retry(self, error):
         if isinstance(error, HTTPError):
             error_code = error.response.status_code
-            if error_code == 429: # Too Many Requests
+            if error_code == 429:  # Too Many Requests
                 return True
-            elif 500 <= error_code < 600: # Internal Error
+            elif 500 <= error_code < 600:  # Internal Error
                 return True
             else:
                 return False
-        elif isinstance(error, ConnectionError): # Unable to connect to service
+        elif isinstance(error, ConnectionError):  # Unable to connect to service
             return True
         else:
             return False
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -67,6 +67,24 @@ class ListFilesInTableResponse:
     add_files: Sequence[AddFile]
 
 
+def retry_with_exponential_backoff(func):
+    def func_with_retry(self, *arg, **kwargs):
+        times_retried = 0
+        sleep_ms = 100
+        while True:
+            times_retried += 1
+            try:
+                return func(self, *arg, **kwargs)
+            except Exception as e:
+                if self._should_retry(e) and times_retried <= self._num_retries:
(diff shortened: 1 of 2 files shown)


python/delta_sharing/rest_client.py
```diff
@@ -180,3 +190,30 @@ def _request_internal(self, request, target: str, data: Optional[Dict[str, Any]]
...
+
+    def _should_retry(self, error):
+        if isinstance(error, HTTPError):
+            error_code = error.response.status_code
+            if error_code == 429:
+                return True
+            elif 500 <= error_code <= 600:
+                return True
+            else:
+                return False
```
FX196, line 217: I'm not sure what 
> case _: InterruptedException => false
      case _: InterruptedIOException => false
      case _: IOException => true
     
map to, so I just did HTTPError for now.

python/delta_sharing/rest_client.py
```diff
@@ -167,6 +172,11 @@ def _get_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
     def _post_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
         return self._request_internal(request=self._session.post, target=target, data=data)
 
+    def _request_internal_with_backoff(self, request, target: str, data: Optional[Dict[str, Any]]):
```
ueshin, line 175: Who will call this? Or is that out of scope of this PR?

python/delta_sharing/rest_client.py
```diff
@@ -180,3 +190,30 @@ def _request_internal(self, request, target: str, data: Optional[Dict[str, Any]]
...
             response.close()
+
+    def _retry_with_exponential_backoff(self, func):
+        times_retried = 0
+        sleep_ms = 100
+        while True:
+            times_retried += 1
+            try:
+                return func()
+            except Exception as e:
```
ueshin, line 201: I guess we can use `HTTPError` here if we re-throw the exception when it's not `HTTPError`?

ueshin: Then we can focus on only `HTTPError` in `_should_retry`.

python/delta_sharing/rest_client.py
FX196: Oops I forgot to change `_get_internal` and `_post_internal` to call this

python/delta_sharing/rest_client.py
FX196: The scala version also retries when an IOException is hit. I wasn't sure if we want to do the same for Python, so I asked in my comment above https://github.com/delta-io/delta-sharing/pull/49#discussion_r673571543

python/delta_sharing/rest_client.py
```diff
@@ -162,10 +167,19 @@ def close(self):
...
     def _post_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
-        return self._request_internal(request=self._session.post, target=target, data=data)
+        return self._request_internal_with_backoff(
+            request=self._session.post, target=target, data=data
+        )
+
+    def _request_internal_with_backoff(self, request, target: str, data: Optional[Dict[str, Any]]):
+        return self._retry_with_exponential_backoff(
+            lambda: self._request_internal(request=request, target=target, data=data)
+        )
```
ueshin, line 182: After a second look, I found this is not a good idea.

As `self._request_internal(request=request, target=target, data=data)` returns a generator, the `_retry_with_exponential_backoff` is only covering the creation of the generator which even doesn't trigger the request yet.

We should pass the whole process to it, e.g., https://github.com/delta-io/delta-sharing/blob/14336db83ad5aba81d4f2b18ad410d20a36e15bd/python/delta_sharing/rest_client.py#L155-L164

FX196: Good point. Do we want to use a decorator in this case?

ueshin: Do you mean decorating functions like `list_shares`, `list_schemas`, etc. with it?
That sounds good.

python/delta_sharing/rest_client.py
```diff
@@ -64,15 +67,37 @@ class ListFilesInTableResponse:
...
+            except Exception as e:
+                if self._should_retry(e) and times_retried <= self._num_retries:
+                    logging.info(f"Sleeping {sleep_ms} to retry because of error {e}")
+                    self._sleeper(sleep_ms)
+                    sleep_ms *= 2
+                else:
+                    raise e
+
+    return func_with_retry
+
```
FX196, line 87: Had to put this outside because I get errors if I do `retry_with_exponential_backoff(self, func)`
https://stackoverflow.com/questions/38524332/declaring-decorator-inside-a-class

python/delta_sharing/rest_client.py
FX196: Decorated list_shares, list_schemas, etc with a retry decorator. Can you take another look? Thanks!

python/delta_sharing/tests/test_rest_client.py
```diff
@@ -24,10 +27,54 @@
...
+
+        @retry_with_exponential_backoff
+        def fail_before_success(self):
+            if len(self.sleeps) < 4:
+                raise self.error
+            else:
+                return True
+
+    wrapper = TestWrapper()
+    assert wrapper.success()
```
ueshin, line 64: nit: we might want to check `wrapper.sleeps` here as well to confirm it succeeds without any error.

python/delta_sharing/rest_client.py
ueshin: I'm not sure there is a corresponding error from `requests` module.
I guess it wraps such errors in `HTTPError`.

ueshin (approved):
Otherwise, LGTM.

python/delta_sharing/tests/test_rest_client.py
FX196: Added :)

python/delta_sharing/rest_client.py
zsxwing: What's the error when the server is down? Could you try it? It would be great we can still retry so that even if we are restarting the server, the client can still survive.

zsxwing (commented):
Overall looks good. Could you check what are possible errors except HTTPError? It would be great if we can retry when someone just restarts the server.

FX196:
@zsxwing Anything else besides the "server down" error that we should retry?

zsxwing:
Looks like https://docs.python.org/3/library/exceptions.html#ConnectionError should cover the transient server errors.

FX196:
Added comments and ConnectionError

zsxwing (approved):
LGTM

PR #50, opened 2021-07-22
Fix KeyError in client when response contains empty items

FX196:
When the items field of the protobuf is empty, the resulting json won't contain the items field. This PR adds checks so that we won't have KeyErrors or NullPointerExceptions if we have empty responses for listSchema and listTables.

Added tests with empty shares and schemas.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -109,12 +109,12 @@ private[sharing] class DeltaSharingRestClient(
     val target = getTargetUrl("shares")
     val shares = ArrayBuffer[Share]()
     var response = getJson[ListSharesResponse](target)
-    shares ++= response.items
+    if (response != null && response.items != null) shares ++= response.items
     while (response.nextPageToken.nonEmpty) {
       val encodedPageToken = URLEncoder.encode(response.nextPageToken.get, "UTF-8")
       val target = getTargetUrl(s"/shares?pageToken=$encodedPageToken")
       response = getJson[ListSharesResponse](target)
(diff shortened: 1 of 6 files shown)


zhuansunxt (approved):
Thanks!

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -109,12 +109,12 @@ private[sharing] class DeltaSharingRestClient(
     val target = getTargetUrl("shares")
     val shares = ArrayBuffer[Share]()
     var response = getJson[ListSharesResponse](target)
-    shares ++= response.items
+    if (response != null && response.items != null) shares ++= response.items
```
ueshin, line 112: I'm not familiar with the code style here, but I'd prefer to always surround with `{` ... `}` for if-block. cc @zsxwing

ueshin (commented):
I assume the test for this fix is covered by `test_list_all_tables` in `test_delta_sharing.py`. Is that right?

FX196:
> I assume the test for this fix is covered by `test_list_all_tables` in `test_delta_sharing.py`. Is that right?

Yes, I added an empty share and a share with an empty schema in TestResource.scala, and the changes are implicitly tested by `test_list_all_tables`. Same for the scala client.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
FX196: Sure, I added the {} for the block.

ueshin (approved):
LGTM, pending tests.

zsxwing (approved):

PR #51, opened 2021-07-23
Add note about missing items

FX196:
Update PROTOCOL.md with note about the items field


PROTOCOL.md
```diff
@@ -68,6 +68,8 @@ Query Parameters | **maxResults** (type: Int, optional): The maximum number of r
 Response Header | Content-Type: application/json; charset=utf-8
 Response Body | <pre>{<br>  "items": [<br>    {<br>      "name": string<br>    }<br>  ],<br>  "nextPageToken": string,<br>}</pre>
 
+Note: the "items" field may be an empty array or missing when no results are found.
```
zsxwing, line 71: Could you move this into `Response Body`?

```suggestion
Note: the "items" field may be an empty array or missing when no results are found. The client should handle both cases.
```

zsxwing (approved):
LGTM

PR #55, opened 2021-08-24
Add limit pushdown rule

FX196:
This PR adds a custom catalyst rule for doing limit pushdown for Delta Sharing queries. When the rule is applied, the limit in SQL queries will be pushed down and sent to the server as `limitHints`. e.g. `SELECT * FROM table LIMIT 100`

Added tests.

### DIFF
--- a/python/delta_sharing/version.py
+++ b/python/delta_sharing/version.py
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-__version__ = "0.3.0.dev0"
+__version__ = ""
--- a/python/delta_sharing/version.py
+++ b/python/delta_sharing/version.py
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-__version__ = ""
+__version__ = "0.3.0.dev0"
--- a/python/delta_sharing/version.py
+++ b/python/delta_sharing/version.py
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-__version__ = "0.2.0.dev0"
+__version__ = "0.3.0.dev0"
--- a/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
@@ -27,12 +27,13 @@ import org.apache.spark.sql.{Column, Encoder, SparkSession}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.analysis.{Resolver, UnresolvedAttribute}
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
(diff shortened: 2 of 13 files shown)


spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSuite.scala
```diff
@@ -22,7 +22,7 @@ import scala.util.Random
 
 import org.apache.commons.io.IOUtils
 import org.apache.hadoop.fs.{FileSystem, Path}
-import org.apache.spark.sql.{QueryTest, Row}
+import org.apache.spark.sql.{QueryTest, Row, SparkSession}
```
zsxwing, line 25: nit: unused change

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -55,7 +56,11 @@ private[sharing] class RemoteDeltaLog(
   def createRelation(): BaseRelation = {
     val spark = SparkSession.active
     val snapshotToUse = snapshot
-    val fileIndex = new RemoteDeltaFileIndex(spark, this, path, snapshotToUse)
+    val fileIndex = new RemoteDeltaFileIndex(spark, this, path, snapshotToUse, None)
+    if (spark.sessionState.conf.getConfString(
+      "spark.delta.sharing.limitPushdown.enabled", "false").toBoolean) {
```
zsxwing, line 61: this should be enabled by default

spark/src/main/scala/io/delta/sharing/spark/perf/DeltaSharingLimitPushDown.scala
```diff
@@ -0,0 +1,56 @@
...
+object DeltaSharingLimitPushDown extends Rule[LogicalPlan] {
+
+  def setup(spark: SparkSession): Unit = synchronized {
+    if (!spark.experimental.extraOptimizations.contains(DeltaSharingLimitPushDown) ) {
+      spark.experimental.extraOptimizations ++= Seq(DeltaSharingLimitPushDown)
+    }
+  }
+
+  def apply(p: LogicalPlan): LogicalPlan = {
+    p transform {
```
zsxwing, line 36: let's also check `plan.conf.getConfString("spark.delta.sharing.limitPushdown.enabled", "true")`. If it's not enabled, we just return `p` directly.

python/delta_sharing/version.py
```diff
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-__version__ = "0.3.0.dev0"
+__version__ = ""
```
zsxwing, line 17: unnecessary change

zsxwing (commented):
One minor comment. Otherwise, LGTM

python/delta_sharing/version.py
FX196: Sorry don't know how that got in there. Reverted.

zsxwing (approved):

PR #56, opened 2021-09-09
Initial PR to add Support for Azure Data Lake Storage

### LINKED ISSUE #21: Support Azure Storage in Delta Sharing Server
Azure supports [shared access signature](https://docs.microsoft.com/en-us/rest/api/storageservices/delegate-access-with-shared-access-signature) which is similar to signed url in other storage systems. It's not hard to make Delta Sharing Server support it.

goodwillpunning:
Fixes #21 

This commit  adds support for sharing Delta tables on Azure Blob Storage and Data Lake Storage Gen2. In particular, this commit adds a new class `AzureFileSigner` which generates a pre-signed URL to Azure blob using a Shared Access Account Policy. Also, this commit automatically sets the Hadoop configuration with the `AZURE_STORAGE_ACCOUNT` and `AZURE_STORAGE_KEY` environment variables, which is used to generate the pre-signed blob URLs.

Signed-off-by: Will Girten <will.girten@databricks.com>

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
+++ b/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
@@ -22,6 +22,7 @@ import java.util.concurrent.TimeUnit.SECONDS
 
 import com.amazonaws.HttpMethod
 import com.amazonaws.services.s3.model.GeneratePresignedUrlRequest
+import com.microsoft.azure.storage.{CloudStorageAccount, SharedAccessAccountPolicy}
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.s3a.DefaultS3ClientFactory
 import org.apache.hadoop.util.ReflectionUtils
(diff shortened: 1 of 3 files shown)


server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -68,20 +70,46 @@ class DeltaSharedTable(
...
 
   private val deltaLog = withClassLoader {
     val tablePath = new Path(tableConfig.getLocation)
     val fs = tablePath.getFileSystem(conf)
-    if (!fs.isInstanceOf[S3AFileSystem]) {
-      throw new IllegalStateException("Cannot share tables on non S3 file systems")
+    if (!(fs.isInstanceOf[S3AFileSystem] ||
+      fs.isInstanceOf[NativeAzureFileSystem] ||
+      fs.isInstanceOf[AzureBlobFileSystem])) {
+      throw new IllegalStateException("Cannot share tables on non S3 or non Azure file systems")
```
dmoore247, line 90: Could this error message be a bit more scalable?
Something like `F" Filesystem {fs.type} is unsupported"`

dmoore247: Thinking ahead as this scales to multiple clouds, even on-prem object stores... should cloud specific implementations reside in their own class files.
Can one instance of the server support multiple both Azure and AWS file systems?

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
```diff
@@ -47,3 +48,44 @@ class S3FileSigner(
...
+
+    // Generate a Shared Access Policy with expiration
+    val sharedAccessAccountPolicy = new SharedAccessAccountPolicy()
+    sharedAccessAccountPolicy.setPermissionsFromString("racwdlup")
+    sharedAccessAccountPolicy.setSharedAccessStartTime(startDate)
+    sharedAccessAccountPolicy.setSharedAccessExpiryTime(expirationDate)
+    sharedAccessAccountPolicy.setResourceTypeFromString("sco")
+    sharedAccessAccountPolicy.setServiceFromString("bfqt")
+
+    val sasToken = cloudStorageAccount.generateSharedAccessSignature(sharedAccessAccountPolicy)
```
zsxwing, line 83: Is `sasToken` only allowed to access the given `bucket` and `objectKey`?  I don't see any place passing them to generate `sasToken`. Looks like `sasToken` has the same power as the long live storage key.

zsxwing:
@goodwillpunning Thanks for the contribution. As per our offline discussion, I will merge this and take over the rest of work.

PR #59, opened 2021-09-15
Fix Azure support for Delta Sharing Server

### LINKED ISSUE #21: Support Azure Storage in Delta Sharing Server

zsxwing:
This is a followup PR for #56. There are multiple changes:

- Add integration tests for Azure.
- Change the pre-signed url generation for Azure to make sure it's scoped down to the path only. We also have tests verifying this.
- Refactor the code to leverage hadoop-azure to get credentials rather than from the system env. The user can just follow the hadoop-azure doc to set up the credentials. See the updated README for more information.

Closes #21

### DIFF
--- a/python/dev/lint-python
+++ b/python/dev/lint-python
@@ -246,7 +246,7 @@ function black_test {
     fi
 
     echo "starting black test..."
-    BLACK_REPORT=$( ($BLACK_BUILD delta_sharing --line-length 100 --check ) 2>&1)
+    BLACK_REPORT=$( ($BLACK_BUILD delta_sharing --line-length 100 --check --diff) 2>&1)
     BLACK_STATUS=$?
 
     if [ "$BLACK_STATUS" -ne 0 ]; then
--- a/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
+++ b/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
@@ -171,8 +171,6 @@ object AbfsFileSigner {
       fs: AzureBlobFileSystem,
       uri: URI,
       preSignedUrlTimeoutSeconds: Long): CloudFileSigner = {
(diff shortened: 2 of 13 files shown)


python/dev/lint-python
```diff
@@ -246,7 +246,7 @@ function black_test {
     fi
 
     echo "starting black test..."
-    BLACK_REPORT=$( ($BLACK_BUILD delta_sharing --line-length 100 --check ) 2>&1)
+    BLACK_REPORT=$( ($BLACK_BUILD delta_sharing --line-length 100 --check --diff) 2>&1)
```
zsxwing, line 249: Update this to show the style details in CI

zsxwing:
cc @goodwillpunning

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
```diff
@@ -39,53 +44,147 @@ class S3FileSigner(
...
+    sharedAccessPolicy.setPermissions(java.util.EnumSet.of(SharedAccessBlobPermissions.READ))
+    sharedAccessPolicy.setSharedAccessExpiryTime(expiration)
+    sharedAccessPolicy
+  }
+
+  override def sign(path: Path): String = {
+    val containerRef = blobClient.getContainerReference(container)
+    val objectKey = objectKeyExtractor(path)
+    assert(objectKey.nonEmpty, s"cannot get object key from $path")
+    val blobRef = containerRef.getBlockBlobReference(objectKeyExtractor(path))
```
zhuansunxt, line 103: You can use `objectKey` here.

FX196 (approved):

PR #67, opened 2021-09-29
Update Python client to use listAllTables REST API

zhuansunxt:
Also fixed test failures caused by test setup change.


zsxwing (approved):
LGTM

PR #76, opened 2021-11-06
Add a limit parameter to 'load_as_pandas'

### LINKED ISSUE #62: Add a limit parameter to load_as_pandas
Currently `load_as_pandas` will load the entire table to memory. Sometimes, the user may just want to take some data from a big table to explore. We can add a `limit` parameter to `load_as_pandas` so that the user can just read small data from a shared table.

zsxwing:
This PR adds a new parameter `limit` to `load_as_pandas`. Users can use this parameter to limit the number of rows to be loaded into a pandas DataFrame. It can be used to explore a shared table without loading the entire table into the memory.

This PR also upgrades pyarrow to 4.0.0 so that we can use [dataset.head](https://issues.apache.org/jira/browse/ARROW-9731) to load only the first N rows.

Closes #62

### DIFF
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -33,7 +33,7 @@ class DeltaSharingReader:
         rest_client: DataSharingRestClient,
         *,
         predicateHints: Optional[Sequence[str]] = None,
-        limitHint: Optional[int] = None
+        limit: Optional[int] = None,
     ):
         self._table = table
         self._rest_client = rest_client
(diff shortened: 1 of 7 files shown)


python/delta_sharing/reader.py
```diff
@@ -33,7 +33,7 @@ def __init__(
         rest_client: DataSharingRestClient,
         *,
         predicateHints: Optional[Sequence[str]] = None,
-        limitHint: Optional[int] = None
+        limit: Optional[int] = None,
```
FX196, line 36: Are we using `limit` instead of `limitHint` because we will always return fewer than `limit`?

FX196 (approved):

python/delta_sharing/reader.py
zsxwing: Yep

PR #77, opened 2021-11-09
[SC-88052] Adds expirationTime as a field of DeltaSharingProfile

linzhou-db:
Adds expirationTime as a field of DeltaSharingProfile to serve as an informational field indicating when the token is going to expire.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -679,7 +679,7 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-
+expiration_time | The expiration time of the access token.
 Example:
 
 ```
@@ -687,5 +687,6 @@ Example:
   "shareCredentialsVersion": 1,
   "endpoint": "https://sharing.delta.io/delta-sharing/",
   "bearerToken": "<token>"
+  "expiration_time":"9999-12-31 23:59:59.999"
 }
 ```
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -679,7 +679,7 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-expirationTime | The expiration time of the bearer token.
+expirationTime | The expiration time of the bearer token. This field is optional.
 Example:
 
 ```
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -679,14 +679,14 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-expiration_time | The expiration time of the access token.
+expirationTime | The expiration time of the bearer token.
 Example:
 
 ```
 {
   "shareCredentialsVersion": 1,
   "endpoint": "https://sharing.delta.io/delta-sharing/",
-  "bearerToken": "<token>"
-  "expiration_time":"9999-12-31 23:59:59.999"
+  "bearerToken": "<token>",
+  "expirationTime": "9999-12-31 23:59:59.999"
 }
 ```
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -679,7 +679,7 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-expirationTime | The expiration time of the bearer token. This field is optional.
+expirationTime | The expiration time of the bearer token in [ISO 8601 format](https://www.w3.org/TR/NOTE-datetime). This field is optional.
 Example:
 
 ```
@@ -687,6 +687,6 @@ Example:
   "shareCredentialsVersion": 1,
   "endpoint": "https://sharing.delta.io/delta-sharing/",
   "bearerToken": "<token>",
-  "expirationTime": "9999-12-31 23:59:59.999"
+  "expirationTime": "2021-11-12T00:12:29.0Z"
 }
 ```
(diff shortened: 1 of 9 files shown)


PROTOCOL.md
```diff
@@ -679,13 +679,14 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-
+expiration_time | The expiration time of the access token.
```
zhuansunxt, line 682: I think we should stick to camelCase (not snake_case).

zhuansunxt, line 682: s/access token/bearer token to be consistent

linzhou-db: Done.

linzhou-db: Good catch, updated.
And sent out https://github.com/databricks/universe/pull/131640 to fix server side.

PROTOCOL.md
```diff
@@ -679,13 +679,14 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-
+expirationTime | The expiration time of the bearer token.
```
zsxwing, line 682:
```suggestion
expirationTime | The expiration time of the bearer token. This field is optional.
```

zsxwing (commented):
Could you add a test to verify we can still handle the case that `expirationTime` is missing?

PROTOCOL.md
linzhou-db: done

linzhou-db:
> Could you add a test to verify we can still handle the case that `expirationTime` is missing?

Added

python/delta_sharing/rest_client.py
```diff
@@ -274,3 +279,14 @@ def _should_retry(self, error):
             return True
         else:
             return False
+
+    def _error_on_expired_token(self, error):
+        token_expired = False
```
zsxwing, line 284: Maybe write it in this way to avoid parsing the expiration time when it's not 401
```
if isinstance(error, HTTPError) and error.response.status_code == 401:
  try:
            expiration_time = datetime.fromisoformat(self._profile.expiration_time)
            return datetime.now() > expiration_time
        except Exception:
            return False
else:
  return False
```
- `fromisotime` should be `fromisoformat`
- We should use `401 Unauthorized` rather than `403 Forbidden` .

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -255,8 +266,12 @@ class DeltaSharingRestClient(
 
         val statusCode = status.getStatusCode
         if (statusCode != HttpStatus.SC_OK) {
+          var additionalErrorInfo = ""
+          if (statusCode == HttpStatus.SC_FORBIDDEN && tokenExpired(profile)) {
```
zsxwing, line 270: SC_UNAUTHORIZED

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -228,6 +230,15 @@ class DeltaSharingRestClient(
     new HttpHost(url.getHost, port, protocol)
   }
 
+  private def tokenExpired(profile: DeltaSharingProfile): Boolean = {
+    try {
```
zsxwing, line 234: if (profile.expirationTime == null) return false

zsxwing (changes requested):
Left some comments. It would be great to use a standard format for `expirationTime` that includes the timezone information.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -255,8 +266,12 @@ class DeltaSharingRestClient(
 
         val statusCode = status.getStatusCode
         if (statusCode != HttpStatus.SC_OK) {
+          var additionalErrorInfo = ""
+          if (statusCode == HttpStatus.SC_FORBIDDEN && tokenExpired(profile)) {
+            additionalErrorInfo = s"It may be caused by expired token:${profile.expirationTime}"
```
zsxwing, line 271: `It may be caused by an expired token as it has expired at ${profile.expirationTime}`.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: done.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: updated.

zhuansunxt (approved):
LGTM

PROTOCOL.md
```diff
@@ -679,13 +679,14 @@ Field Name | Descrption
 shareCredentialsVersion | The file format version of the profile file. This version will be increased whenever non-forward-compatible changes are made to the profile format. When a client is running an unsupported profile file format version, it should show an error message instructing the user to upgrade to a newer version of their client.
 endpoint | The url of the sharing server.
 bearerToken | The [bearer token](https://tools.ietf.org/html/rfc6750) to access the server.
-
+expirationTime | The expiration time of the bearer token. This field is optional.
```
zsxwing, line 682:
```suggestion
expirationTime | The expiration time of the bearer token in [ISO 8601 format](https://www.w3.org/TR/NOTE-datetime). This field is optional.
```

PROTOCOL.md
```diff
@@ -679,13 +679,14 @@ Field Name | Descrption
...
 Example:
 
 ```
 {
   "shareCredentialsVersion": 1,
   "endpoint": "https://sharing.delta.io/delta-sharing/",
-  "bearerToken": "<token>"
+  "bearerToken": "<token>",
+  "expirationTime": "9999-12-31 23:59:59.999"
```
zsxwing, line 690:
```suggestion
  "expirationTime": "2021-11-12T00:12:29Z"
```

Could you also update other hard code strings in tests using ISO 8601 format?

zsxwing (commented):
Left two minor comments. Otherwise LGTM

PROTOCOL.md
linzhou-db: done

linzhou-db: done

PROTOCOL.md
linzhou-db: done.

python/delta_sharing/rest_client.py
```diff
@@ -86,6 +87,10 @@ def func_with_retry(self, *arg, **kwargs):
                     logging.info(f"Sleeping {sleep_ms} to retry because of error {e}")
                     self._sleeper(sleep_ms)
                     sleep_ms *= 2
+                elif self._error_on_expired_token(e):
+                    raise HTTPError(
+                        f"It may be caused by expired token:{self._profile.expiration_time}"
```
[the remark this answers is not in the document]
linzhou-db: updated.

zsxwing (approved):
LGTM

Issue #20, opened 2021-05-25
Support Google Cloud Storage in Delta Sharing Server


zsxwing:
Google Cloud Storage supports [Signed URLs](https://cloud.google.com/storage/docs/access-control/signed-urls). It's not hard to make Delta Sharing Server support it.

kohei-tosshy:
Hello.
I am trying to implement this by using GCP-connector for Hadoop 2.
Hadoop 2 API is also used for S3 and Azure in Delta Sharing.
But, there is a depencency problem between Spark and GCP-connector for Hadoop 2 related with Google Guava.
(https://cloud.google.com/dataproc/docs/guides/manage-spark-dependencies)

I think we can choose two way to solve this problem.

1. Use GCP-connector for Hadoop 2, then solve the depencency problem by sbt's shading 
2. Use GCP-connector for Hadoop 3. This cause the Hadoop API's difference among the cloud storages, but we don't need to use shading bacause no dependency problem occurs.

I want to ask which way is better.

zsxwing:
Can we exclude guava in build.sbt? I remember we don't use it. If it's not working, I prefer option 2. We can just upgrade Hadoop version to 3 in the server project.

zsxwing:
@kohei-tosshy Feel free to reach us if you hit any issue. I will try to get a GCS account in parallel so that we can set up integration tests when you have a PR ready.

zsxwing:
This is added by #81 81

PR #81, opened 2021-11-21
Support GCS for delta sharing Server

kohei-tosshy:
This is a PR for #20.

I implemented GCS support by using Google Cloud Client Library (https://cloud.google.com/storage/docs/reference/libraries) and Google Cloud Storage connector (https://cloud.google.com/dataproc/docs/concepts/connectors/cloud-storage).
Google Cloud Storage connector (GCS connector) has a dependency problem with Apache Spark releated with Google Guava, so I added this library's shaded jar in server/lib (SBT Unmanaged dependencies, not Managed dependencies in build.sbt).
I implemented this very forcefully, is this no problem?

To use GCP for delta sharing server, you have to do 2 steps.

1. In your terminal, Do `export GOOGLE_APPLICATION_CREDENTIALS=</path/to/your_gcp_service_acount_key.json>`
2. In the server config (e,g, delta-sharing-server.yaml), add a setting for GCP
```
shares:
- name: "share_gcp"
  schemas:
  - name: "schema_gcp"
    tables:
    - name: "table_gcp"
      location: "gs://delta-start/delta/"
```

If you want to coneect this server from Apache Spark,  execute spark-shell or spark-submit like below.

```
spark-shell \
--packages io.delta:delta-core_2.12:1.0.0,io.delta:delta-contribs_2.12:1.0.0,io.delta:delta-sharing-spark_2.12:0.2.0 \
--jars </path/to/gcs-connector-latest-hadoop2.jar> \
--conf spark.delta.logStore.oci.impl=io.delta.storage.GCSLogStore \
--conf spark.hadoop.fs.gs.impl=com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS \
--conf spark.hadoop.google.cloud.auth.service.account.enable=true 
--conf spark.hadoop.google.cloud.auth.service.account.json.keyfile=</path/to/your_gcp_service_acount_key.json>
```

I haven't implement integration test on GCP support yet.
If I understand correctly, what I should do is to add the almost same test with Azure support (https://github.com/delta-io/delta-sharing/blob/main/server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala#L510).
Is this correct?
If my understanding is correct, I'll add this integreation test later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 .pydevproject
 .scala_dependencies
 .settings
+/lib/
 R-unit-tests.log
 R/unit-tests.out
 R/cran-check.out
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@
 .pydevproject
 .scala_dependencies
 .settings
-/lib/
 R-unit-tests.log
 R/unit-tests.out
 R/cran-check.out
--- a/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
+++ b/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
@@ -22,6 +22,10 @@ import java.util.concurrent.TimeUnit.SECONDS
 
 import com.amazonaws.HttpMethod
 import com.amazonaws.services.s3.model.GeneratePresignedUrlRequest
+import com.google.cloud.storage.BlobId
+import com.google.cloud.storage.BlobInfo
+import com.google.cloud.storage.Storage
+import com.google.cloud.storage.StorageOptions
 import com.microsoft.azure.storage.{CloudStorageAccount, SharedAccessProtocols, StorageCredentialsSharedAccessSignature}
 import com.microsoft.azure.storage.blob.{SharedAccessBlobPermissions, SharedAccessBlobPolicy}
 import org.apache.hadoop.conf.Configuration
@@ -31,6 +35,7 @@ import org.apache.hadoop.fs.azurebfs.{AzureBlobFileSystem, AzureBlobFileSystemSt
 import org.apache.hadoop.fs.azurebfs.services.AuthType
 import org.apache.hadoop.fs.s3a.DefaultS3ClientFactory
 import org.apache.hadoop.util.ReflectionUtils
+
 
 trait CloudFileSigner {
   def sign(path: Path): String
@@ -186,5 +191,22 @@ object AbfsFileSigner {
       container,
       preSignedUrlTimeoutSeconds,
       getRelativePath(abfsStore, _))
+  }
+}
+
+class GCSFileSigner(
+    name: URI,
+    conf: Configuration,
+    preSignedUrlTimeoutSeconds: Long) extends CloudFileSigner{
+
+  override def sign(path: Path): String = {
+    val absPath = path.toUri
+    val bucketName = absPath.getHost
+    val objectName = absPath.getPath.stripPrefix("/")
+    val storage = StorageOptions.newBuilder.build.getService
+    val blobInfo = BlobInfo.newBuilder(BlobId.of(bucketName, objectName)).build
+    storage.signUrl(
+      blobInfo, preSignedUrlTimeoutSeconds, SECONDS, Storage.SignUrlOption.withV4Signature())
+      .toString
   }
 }
--- a/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
+++ b/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
@@ -197,13 +197,15 @@ object AbfsFileSigner {
 class GCSFileSigner(
     name: URI,
     conf: Configuration,
-    preSignedUrlTimeoutSeconds: Long) extends CloudFileSigner{
+    preSignedUrlTimeoutSeconds: Long) extends CloudFileSigner {
+
+  private val storage = StorageOptions.newBuilder.build.getService
 
   override def sign(path: Path): String = {
     val absPath = path.toUri
     val bucketName = absPath.getHost
     val objectName = absPath.getPath.stripPrefix("/")
-    val storage = StorageOptions.newBuilder.build.getService
+    assert(objectName.nonEmpty, s"cannot get object key from $path")
     val blobInfo = BlobInfo.newBuilder(BlobId.of(bucketName, objectName)).build
     storage.signUrl(
       blobInfo, preSignedUrlTimeoutSeconds, SECONDS, Storage.SignUrlOption.withV4Signature())
--- a/server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
+++ b/server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
@@ -552,7 +552,8 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
     assert(e.getMessage.contains("Server returned HTTP response code: 403")) // 403 Forbidden
   }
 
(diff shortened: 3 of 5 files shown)


zsxwing:
Thanks for the contribution. This is awesome!

> so I added this library's shaded jar in server/lib (SBT Unmanaged dependencies, not Managed dependencies in build.sbt).

Is Guava the only issue? If so, we can just exclude guava from other dependencies like this https://github.com/delta-io/delta-sharing/commit/a6bd550fd7248f8e667babeb21f946539d2c0ec3 This is better than adding a shaded jar to the project.

> In your terminal, Do `export GOOGLE_APPLICATION_CREDENTIALS=</path/to/your_gcp_service_acount_key.json>`

Does the GOOGLE_APPLICATION_CREDENTIALS env work for reading files using `GoogleHadoopFileSystem`? I saw GCS connector asks users to set a hadoop configuration `google.cloud.auth.service.account.json.keyfile`: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/INSTALL.md#configuring-hadoop Since this is the recommended way, can we just read the key file from the hadoop configuration? This will be similar to https://github.com/delta-io/delta-sharing#azure-data-lake-storage-gen2

> If I understand correctly, what I should do is to add the almost same test with Azure support 

Correct. I will do some manual test with your change for now. We can add the real integration tests later.

zsxwing:
@kohei-tosshy Any thoughts on my above questions?

kohei-tosshy:
Thank you for your comment.
And sorry for being late for reacting.

> Is Guava the only issue? If so, we can just exclude guava from other dependencies like this a6bd550 This is better than adding a shaded jar to the project.

I see. I'll try.

> Does the GOOGLE_APPLICATION_CREDENTIALS env work for reading files using GoogleHadoopFileSystem? I saw GCS connector asks users to set a hadoop configuration google.cloud.auth.service.account.json.keyfile: https://github.com/GoogleCloudDataproc/hadoop-connectors/blob/master/gcs/INSTALL.md#configuring-hadoop Since this is the recommended way, can we just read the key file from the hadoop configuration? This will be similar to https://github.com/delta-io/delta-sharing#azure-data-lake-storage-gen2

To get a presigned URL, I used Google Google Cloud Client Library.
I only used `GoogleHadoopFileSystem` only to changing a path.
GOOGLE_APPLICATION_CREDENTIALS is needed to use Google Google Cloud Client Library.

> Correct. I will do some manual test with your change for now. We can add the real integration tests later.

I see. I'll try.

kohei-tosshy:
@zsxwing

Sorry for being late for revising.
I revised Google Guava's dependency problem and added an integration test for GCS support.

.gitignore
```diff
@@ -21,7 +21,6 @@
 .pydevproject
 .scala_dependencies
 .settings
-/lib/
```
zhuansunxt, line 24: Revert this change?

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
```diff
@@ -188,3 +193,20 @@ object AbfsFileSigner {
...
+
+class GCSFileSigner(
+    name: URI,
+    conf: Configuration,
+    preSignedUrlTimeoutSeconds: Long) extends CloudFileSigner{
+
+  override def sign(path: Path): String = {
+    val absPath = path.toUri
+    val bucketName = absPath.getHost
+    val objectName = absPath.getPath.stripPrefix("/")
```
zhuansunxt, line 205: Would be great to have a check `assert(objectName.nonEmpty, s"cannot get object key from $path")` similar to AWS and Azure signer

zhuansunxt (approved):
The code change LGTM.

@zsxwing Do you want to have another look? It'll be nice to include this in the upcoming release.

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
```diff
@@ -188,3 +193,20 @@ object AbfsFileSigner {
       getRelativePath(abfsStore, _))
   }
 }
+
+class GCSFileSigner(
+    name: URI,
+    conf: Configuration,
+    preSignedUrlTimeoutSeconds: Long) extends CloudFileSigner{
```
zsxwing, line 200: nit
```suggestion
    preSignedUrlTimeoutSeconds: Long) extends CloudFileSigner {
```

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
```diff
@@ -188,3 +193,20 @@ object AbfsFileSigner {
...
+
+  override def sign(path: Path): String = {
+    val absPath = path.toUri
+    val bucketName = absPath.getHost
+    val objectName = absPath.getPath.stripPrefix("/")
+    val storage = StorageOptions.newBuilder.build.getService
```
zsxwing, line 206: Is `storage` thread-safe? If so, we can save it as a val in `GCSFileSigner` to avoid loading the credentials for each url.

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -551,4 +551,34 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
     }
     assert(e.getMessage.contains("Server returned HTTP response code: 403")) // 403 Forbidden
   }
+
+  integrationTest("gcs support") {
```
zsxwing, line 555: nit: Could you use `ignore` for this test for now? We will set up the credentials and enable it later.

zsxwing:
> To get a presigned URL, I used Google Google Cloud Client Library.
I only used GoogleHadoopFileSystem only to changing a path.
GOOGLE_APPLICATION_CREDENTIALS is needed to use Google Google Cloud Client Library.

Google Google Cloud Client Library also supports other ways to config credentials. But it seems hard to mimic what GoogleHadoopFileSystem does to support various of ways to config credentials. `GOOGLE_APPLICATION_CREDENTIALS` env is a good start right now. We can support other ways in future if people are asking.

Left some minor comments. Otherwise LGTM.

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
kohei-tosshy: I added it.

.gitignore
kohei-tosshy: I reverted it.

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
kohei-tosshy: I revised this.

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
kohei-tosshy: I used `ignore`  instead of `integrationTest`.

server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
kohei-tosshy: I think `storage` is thread-safe.
One instance of `GCSFileSigner` is created in each request for client, so each thread for a request has a dedicated instance of `GCSFileSigner` and won't touch other `GCSFileSigner` instances.
And, `storage` is used only for reading, so it will be OK if some threads for creating a presignied-URL are processed in paralle.

zsxwing (approved):
LGTM! Thanks!

PR #84, opened 2021-11-23
Fix Python connector GET Requests to work with maxResults and nextPageToken

### LINKED ISSUE #83: Python Connector GET Requests does not respect maxResults and nextPageToken
https://github.com/delta-io/delta-sharing/blob/main/python/delta_sharing/rest_client.py#L259

For get requests, the json kwarg does not work. It should be using the params kwarg instead.

wchau:
Fixes #83


zhuansunxt (approved):
LGTM

PR #85, opened 2021-11-24
Enhance protocol documentation

wchau:
Added edge case details with non-negative and 0 max results, and existence of page token.
Fixed JSON formatting.
Identified int types.

https://github.com/delta-io/delta-sharing/blob/c44815eb6b1984e59216e9b7e88b3eeb1b7631db/PROTOCOL.md

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -61,21 +61,56 @@ Here are the list of APIs used by Delta Sharing Protocol. All of the REST APIs u
 
 This is the API to list shares accessible to a recipient.
 
-HTTP Parameters | Value
+HTTP Request | Value
 -|-
-HTTP Method | GET
-Header | Authorization: Bearer {token}
-URL | {prefix}/shares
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -60,21 +60,164 @@ Here are the list of APIs used by Delta Sharing Protocol. All of the REST APIs u
...
+<tr>
+<td>Header</td>
+<td>Content-Type: application/json</td>
+</tr>
+<tr>
+<td>Body</td>
+<td>
+
+```json
+{
```
zsxwing, line 125: Do you plan to add this response format to OSS Delta Sharing server soon? If not, it's better to remove it from the protocol for now.

wchau: Yes, will prepare a PR to add now.

wchau: https://github.com/delta-io/delta-sharing/pull/89

I have also changed error_code to errorCode in the protocol. I think it is theoretically possible to change the behaviour inside DB's delta sharing servers to output camel case.

zhuansunxt:
I think the change overall looks pretty good. Two suggestions:
1. Can we split the change to two PRs, one for error code and one for documentation improvement?
2. Do you think it's better to add a dedicated response code section compared to having it for every API method?

wchau:
> 1. Can we split the change to two PRs, one for error code and one for documentation improvement?
Done
> 2. Do you think it's better to add a dedicated response code section compared to having it for every API method?
I am creating a poor man's version of DB's official API docs which follow the same format to try to be somewhat consistent. eg: https://docs.databricks.com/dev-tools/api/latest/account.html#operation/get-credential-configs

linzhou-db (commented):
Overall LGTM, do we also want to cover the status code in the http response?

linzhou-db (approved):

PR #87, opened 2021-11-29
Fix nextPageToken in server

### LINKED ISSUE #86: Server returns next_page_token instead of nextPageToken
As defined in the protocol.

wchau:
Fixes #86

### DIFF
--- a/server/src/main/protobuf/protocol.proto
+++ b/server/src/main/protobuf/protocol.proto
@@ -49,22 +49,22 @@ message QueryTableRequest {
 
 message ListSharesResponse {
     repeated Share items = 1;
-    optional string next_page_token = 2;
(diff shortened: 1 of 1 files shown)


server/src/main/protobuf/protocol.proto
```diff
@@ -49,22 +49,22 @@ message QueryTableRequest {
 
 message ListSharesResponse {
     repeated Share items = 1;
-    optional string next_page_token = 2;
+    optional string nextPageToken = 2;
```
zsxwing, line 52: This is correct in oss delta sharing server. You can take a look at the result by running

```
curl -XGET https://sharing.delta.io/delta-sharing/shares/delta_sharing/schemas/default/tables?maxResults=1
```

PR #75, opened 2021-11-03
Add the User-Agent request header to Delta Sharing client

zsxwing:
This PR adds the User-Agent request header to Delta Sharing client so that the server can use this information to track client metrics and it's also easier to debug issues. Here are examples of the header:

- Spark connector

> Delta-Sharing-Spark/0.3.0-SNAPSHOT Spark/3.1.1 Hadoop/3.2.0 Mac_OS_X/10.14.6 Java_HotSpot(TM)_64-Bit_Server_VM/25.171-b11 java/1.8.0_171 scala/2.12.10 java_vendor/Oracle_Corporation

- Pandas connector

> Delta-Sharing-Python/0.3.0.dev0 pandas/1.2.4 PyArrow/4.0.0 Python/3.9.5 System/macOS-10.14.6-x86_64-i386-64bit

Closes #57

PR #88, opened 2021-11-30
Add list_all_tables backward compatibility

zsxwing:
Currently Python list_all_tables API doesn't work with servers using 0.2.0 and below. As per some user feedback, this PR makes it work with old versions by adding a fallback logic.

We don't need to update spark connector. Its DeltaSharingRestClient is private and built mainly for testing the server logic, so no users will be affected.

This PR adds a PR to test the fallback logic. I also manually verified `list_all_tables` using a local server running 0.2.0.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -64,7 +64,7 @@ private[sharing] case class ListAllTablesResponse(
     nextPageToken: Option[String]) extends PaginationResponse
 
 /** A REST client to fetch Delta metadata from remote server. */
-class DeltaSharingRestClient(
+private[spark] class DeltaSharingRestClient(
     profileProvider: DeltaSharingProfileProvider,
     timeoutInSeconds: Int = 120,
     numRetries: Int = 10,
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -64,7 +64,7 @@ private[sharing] case class ListAllTablesResponse(
     nextPageToken: Option[String]) extends PaginationResponse
 
 /** A REST client to fetch Delta metadata from remote server. */
-class DeltaSharingRestClient(
+private[spark] class DeltaSharingRestClient(
```
zsxwing, line 67: I removed this accidentally in #75. Just adding it back.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -295,7 +295,7 @@ class DeltaSharingRestClient(
   }
 }
 
-object DeltaSharingRestClient extends Logging {
+private[spark] object DeltaSharingRestClient extends Logging {
```
zsxwing, line 298: ditto

zhuansunxt (approved):

linzhou-db (approved):

PR #89, opened 2021-11-30
Add support to return errors in JSON to adhere to protocol

wchau:


linzhou-db (approved):

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -55,28 +62,57 @@ class DeltaSharingServiceExceptionHandler extends ExceptionHandlerFunction {
...
-        HttpResponse.of(HttpStatus.INTERNAL_SERVER_ERROR)
+        // Hide message in response
+        HttpResponse.of(
+          HttpStatus.INTERNAL_SERVER_ERROR,
+          MediaType.PLAIN_TEXT_UTF_8,
+          JsonUtils.toJson(
+            Map(
+              "errorCode" -> ErrorCode.INTERNAL_ERROR,
+              "message" -> "")))
       case _ =>
```
linzhou-db, line 116: What will be returned in this case?

wchau: I will lump it with the DeltaInternalException case. There really is no reason why we need a separate Exception, since the fallthrough should be yielding an ISE anyways. It makes the DeltaInternalException class obsolete, but I will keep it around in case we need to distinguish them later.

PR #92, opened 2021-12-07
Block managed table creation for Delta Sharing

zsxwing:
Currently users can run `create table foo using deltaSharing` without location to create a Delta Sharing in Hive Metastore. However, such managed tables are useless because users won't be able to read the table. The default location of managed tables will be pointed to a file system path such as `/user/hive/warehouse/foo`, so the user will see an invalid path error. The path format of Delta Sharing must be `profile_file + "#<share-name>.<schema-name>.<table-name>"`.

This PR updates the code to block managed table creation, and adds a test for this. Also manually tested the following two cases to make sure we don't break external tables:
- Use an old version to create an **external** table and use the jar built by this PR to read it.
- Use the jar built by this PR to create an **external** table and use an old version to read it.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
@@ -28,29 +28,15 @@ import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
 /** A DataSource V1 for integrating Delta into Spark SQL batch APIs. */
-private[sharing] class DeltaSharingDataSource extends TableProvider with RelationProvider
-  with DataSourceRegister {
-
-  override def inferSchema(options: CaseInsensitiveStringMap): StructType = new StructType()
-
-  override def getTable(
-      schema: StructType,
-      partitioning: Array[Transform],
-      properties: java.util.Map[String, String]): Table = {
-    // Return a Table with no capabilities so we fall back to the v1 path.
-    new Table {
-      override def name(): String = s"V1FallbackTable"
-      override def schema(): StructType = new StructType()
-      override def capabilities(): java.util.Set[TableCapability] = Collections.emptySet()
-    }
-  }
+private[sharing] class DeltaSharingDataSource extends RelationProvider with DataSourceRegister {
 
   override def createRelation(
       sqlContext: SQLContext,
       parameters: Map[String, String]): BaseRelation = {
     DeltaSharingDataSource.setupFileSystem(sqlContext)
-    val path = parameters.getOrElse(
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
```diff
@@ -28,29 +28,15 @@ import org.apache.spark.sql.types.StructType
...
-      partitioning: Array[Transform],
-      properties: java.util.Map[String, String]): Table = {
-    // Return a Table with no capabilities so we fall back to the v1 path.
-    new Table {
-      override def name(): String = s"V1FallbackTable"
-      override def schema(): StructType = new StructType()
-      override def capabilities(): java.util.Set[TableCapability] = Collections.emptySet()
-    }
-  }
+private[sharing] class DeltaSharingDataSource extends RelationProvider with DataSourceRegister {
```
zsxwing, line 31: `TableProvider` was added to support advanced features such as time travel. Since we don't support these features today, I just remove it so that we will block managed tables properly.

FX196: To make sure I'm understanding correctly, this blocks managed tables because we will no longer be able to fall back to the v1 path, so only paths in the form of `profile_file + "#<share-name>.<schema-name>.<table-name>"` can be created?

zhuansunxt: @FX196 that's my understanding. Then what happens if we need to add `TableProvider` back to support advanced features?

zhuansunxt (approved):
LGTM

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
zsxwing: Removing TableProvider will make Spark to call `createRelation` below and verify the path. 

> Then what happens if we need to add TableProvider back to support advanced features?

We can add it back in future if we need. Since we have a test for blocking managed tables, adding it back in future would need to fix the test, such as load the table in `getTable` I removed in this PR.

zsxwing: Removing TableProvider will make Spark to call `createRelation` below and verify the path. 

> Then what happens if we need to add TableProvider back to support advanced features?

We can add it back in future if we need. Since we have a test for blocking managed tables, adding it back in future would need to fix the test, such as load the table in `getTable` I removed in this PR.

FX196: Makes sense. If we want to add it back in the future we can check the path in `getTable`?

FX196 (approved):

PR #93, opened 2021-12-07
Extend Delta Sharing Protocol to include Share ID and Table ID

zhuansunxt:
This PR extends the Delta Sharing Protocol to give Share and Table unique IDs. Unique IDs of Share and its tables inside will help the data recipient or data consumer disambiguate the name of datasets as time passes. This is especially useful when data recipient is a large organization and wants to apply access control on the shared dataset within their organization.

For example:
- at time 0 a data provider shares a Share with name `foo`
- at time 1 the data provider removes that Share `foo` to the recipient
- at time 2 the data provider creates another Share also with name `foo` and shares with the recipient again, this time with different tables inside.

As an admin on the data recipient side, I would expect to be able to distinguish the old and new Share in this case and decide with whom to further share the data within my organization. Table ID serve the same purpose.

Note that the IDs are optional to Share and Tables for backward compatibility.


linzhou-db (approved):

wchau:
What you mentioned in the PR about the need for the ID would be good to actually copy into the doc.

zsxwing:
It would be great to mention the requirements of `id` in the protocol, such as `it should be unique for the same service and restarting the service should not change it`.

Do we want to require uniqueness for all share/schema/tables, for example, an id of a share should not be the same as a table id?

zhuansunxt:
@wchau My impression is that the protocol.md serves as a specification without explaining too much on the design rationales. I scanned through the document but did not find a great place to insert that explanation. Maybe creating an issue can help here for the documentation purpose?

@zsxwing I added more documentation emphasizing the uniqueness and immutability requirement. PTAL. Also I did not add Schema ID in this PR since in my view Schema under a Share is a namespace grouping structure rather than a data resource object.

zsxwing (approved):

zhuansunxt:
Closes #94

PR #105, opened 2022-01-07
Integration test for Google GCS hosted table

zhuansunxt:
- Integration test.
- Library version upgrade. The previous version has bug when reading the file content.

The python integration test does not work yet. Still looking.


python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -280,6 +282,12 @@ def list_all_tables(
             pd.DataFrame({"c1": ["foo bar"], "c2": ["foo bar"],}),
             id="Azure Data Lake Storage Gen2",
         ),
+        pytest.param(
```
zhuansunxt, line 285: @ueshin This test currently fails with below error. It looks like fsspec is not handling GCS file correctly. I'm not quite familiar with this part of the code. Do you happen to come across this issue before?

```
self = <pyarrow.fs.FSSpecHandler object at 0x126890070>, path = 'https://storage.googleapis.com/delta-sharing-dev/delta-sharing-test/table1/c2%3Dfoo%20bar/part-00000-3287e08f-94ef-44...3a8be15a0bd00c41b863a16e8f117ab911f99f10b09eeb8ec10b499f7d58b64bd2cbde845b3ab8604f4e1d83e0331d6ec44336cb45f6529538d90d'

    def open_input_file(self, path):
        from pyarrow import PythonFile
    
        if not self.fs.isfile(path):
>           raise FileNotFoundError(path)
E           FileNotFoundError: https://storage.googleapis.com/delta-sharing-dev/delta-sharing-test/table1/c2%3Dfoo%20bar/part-00000-3287e08f-94ef-44fe-9837-4d8eee39363f.c000.snappy.parquet?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=delta-sharing-dev-svc%40databricks-dev.iam.gserviceaccount.com%2F20220107%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20220107T234922Z&X-Goog-Expires=3600&X-Goog-SignedHeaders=host&X-Goog-Signature=0f78839681431f0996043fe7a6698c56ea2e3a12f6597f6fab572bb8a613eb37e33b26c393ad1802f1df2a1cabaaed4eb5486bb9f242e3bfa5c860c3a0b0f0ee542112c7ec3f311bed542a81577da6db230b7abb921dcb21df9eebb3a14bd14118f285c348bfeef8cccece063b45f6fceed8fc5f4ece4ed104639cedf7a0d313381c21968c6e6ad67a40700196fadddc38f6b278d621f740f4aa1eb9159f9e0b9df95afec1be601288171bd937fe5e11943c0cf0c134c25268d10900e6c078767e316699803a8be15a0bd00c41b863a16e8f117ab911f99f10b09eeb8ec10b499f7d58b64bd2cbde845b3ab8604f4e1d83e0331d6ec44336cb45f6529538d90d
```

zsxwing: I dig the code and found a workaround for this file:

```
diff --git a/python/delta_sharing/reader.py b/python/delta_sharing/reader.py
index 6ab26fe..393b140 100644
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -25,6 +25,12 @@ from delta_sharing.converter import to_converters, get_empty_table
 from delta_sharing.protocol import AddFile, Table
 from delta_sharing.rest_client import DataSharingRestClient
 
+try:
+    from yarl import URL
+    from yarl._quoting import _Quoter
+    URL._PATH_REQUOTER = _Quoter(safe="@:", protected="/+=")
+except:
+    pass
 
 class DeltaSharingReader:
     def __init__(
```
This means we will require yarl 1.6.0 (`_PATH_REQUOTER` is not in old versions).

The root cause is our client receives the raw url to send the requests. But when we pass it to fsspec, it will forward the url to aiohttp. There is a workaroud in https://github.com/aio-libs/aiohttp/issues/6413#issuecomment-993393969 to specify our url is an encoded url but we cannot change fsspec code.

The issue will cause we sent an incorrect url
```
https://storage.googleapis.com/delta-sharing-dev/delta-sharing-test/table1/c2=foo%20bar/part-00000-3287e08f-94ef-44fe-9837-4d8eee39363f.c000.snappy.parquet?...
```
rather than
```
https://storage.googleapis.com/delta-sharing-dev/delta-sharing-test/table1/c2%3Dfoo%20bar/part-00000-3287e08f-94ef-44fe-9837-4d8eee39363f.c000.snappy.parquet?...
```
(The difference of two urls is `c2=foo%20bar` and `c2%3Dfoo%20bar`)

This issue doesn't happen in AWS and Azure because they accept both urls, but GCS. will reject `c2=foo%20bar`.

The workaround I posted above is just making yarl.URL not change `%3D` to `=`.

`FileNotFoundError` is caused by https://github.com/fsspec/filesystem_spec/issues/873 . The real error dropped by pyarrow is
```
aiohttp.client_exceptions.ClientResponseError: 403, message='Forbidden', url=URL('https://storage.googleapis.com/delta-sharing-dev/delta-sharing-test/table1/c2=foo%20bar/part-...')
```

README.md
```diff
@@ -281,7 +281,15 @@ The server is using `hadoop-azure` to read Azure Data Lake Storage Gen2. We supp
 ```
 `YOUR-ACCOUNT-NAME` is your Azure storage account and `YOUR-ACCOUNT-KEY` is your account key.
 
-More cloud storage supports will be added in the future.
+### Google Cloud Storage
+
+We support using [Service Account](https://cloud.google.com/iam/docs/service-accounts) to read Google Cloud Storage. You can find more details in [GCP Authentication Doc](https://cloud.google.com/docs/authentication/getting-started).
+
+```
```
zsxwing, line 288: To set up the credentials, you can specify the environment `GOOGLE_APPLICATION_CREDENTIALS` like this before starting the Delta Sharing Server.

zsxwing (commented):
- Could you also add a quick example in https://github.com/delta-io/delta-sharing/blob/main/server/src/universal/conf/delta-sharing-server.yaml.template#L14 ?
- For the test, I think we also need to write code to generate the file that `GOOGLE_APPLICATION_CREDENTIALS` points to.

zhuansunxt:
> * Could you also add a quick example in https://github.com/delta-io/delta-sharing/blob/main/server/src/universal/conf/delta-sharing-server.yaml.template#L14 ?

Added.

> * For the test, I think we also need to write code to generate the file that `GOOGLE_APPLICATION_CREDENTIALS` points to.

Why can not we just setup the file path?

zsxwing:
> Why can not we just setup the file path?

GitHub Actions runs in a container, but we don't want to build a public docker image that includes the google key file. To be clear, I meant the following steps:
- Store the key file content in GitHub secrets.
- Add `GOOGLE_APPLICATION_CREDENTIALS` to point to a file path, such as `/tmp/google_key.json`.
- Read the key file content from env and write it into the file `/tmp/google_key.json`.

zhuansunxt:
@zsxwing I added code to write temp file with key body loaded from github secret in integration test and it is verified to work (see https://github.com/delta-io/delta-sharing/tree/integration-test). However there is another issue of `"Type[URL]" has no attribute "_PATH_REQUOTER"` in the Github Action even if yarl is installed with desirable version (tried both 1.6.0 and 1.7.2).

zsxwing:
> However there is another issue of `"Type[URL]" has no attribute "_PATH_REQUOTER"` in the Github Action even if yarl is installed with desirable version (tried both 1.6.0 and 1.7.2).

We can ignore the mypy check error by adding a comment like this:
```
URL._PATH_REQUOTER = _Quoter(safe="@:", protected="/+=")  # type: ignore
```

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSuite.scala
```diff
@@ -33,6 +33,18 @@ class DeltaSharingSuite extends QueryTest with SharedSparkSession with DeltaShar
 
   import testImplicits._
 
+  private def buildGoogleServiceAccountKeyFile: Unit = {
```
zsxwing, line 36: ditto

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -38,7 +39,21 @@ import io.delta.sharing.server.util.JsonUtils
...
+
+  private def buildGoogleServiceAccountKeyFile: Unit = {
+    // Only write tmp file if the key content is defined, otherwise the key is loaded directly from
+    // the key file path in GOOGLE_APPLICATION_CREDENTIALS.
+    if (sys.env.get("GOOGLE_SERVICE_ACCOUNT_KEY").exists(_.length > 0)) {
+      val serviceAccountKey = sys.env("GOOGLE_SERVICE_ACCOUNT_KEY")
+      val fileWriter = new FileWriter(new File("/tmp/google_service_account_key.json"))
+      fileWriter.write(serviceAccountKey)
+      fileWriter.flush()
+      fileWriter.close()
```
zsxwing, line 55: We can use [FileUtils.writeStringToFile](https://github.com/delta-io/delta-sharing/blob/7dbf108f7cd2f866392d461f743d82144f9d4c17/server/src/test/scala/io/delta/sharing/server/config/ServerConfigSuite.scala#L32) to save multiple lines.

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -38,7 +39,21 @@ import io.delta.sharing.server.util.JsonUtils
 class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
 
   def shouldRunIntegrationTest: Boolean = {
-    sys.env.get("AWS_ACCESS_KEY_ID").exists(_.length > 0)
+    sys.env.get("AWS_ACCESS_KEY_ID").exists(_.length > 0) &&
+      sys.env.get("AZURE_TEST_ACCOUNT_KEY").exists(_.length > 0) &&
+      sys.env.get("GOOGLE_APPLICATION_CREDENTIALS").exists(_.length > 0)
+  }
+
+  private def buildGoogleServiceAccountKeyFile: Unit = {
```
zsxwing, line 47: Could we move this to `TestResource.setupTestTables`? Then it will work for all tests.

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -38,7 +39,21 @@ import io.delta.sharing.server.util.JsonUtils
...
+    // Only write tmp file if the key content is defined, otherwise the key is loaded directly from
+    // the key file path in GOOGLE_APPLICATION_CREDENTIALS.
+    if (sys.env.get("GOOGLE_SERVICE_ACCOUNT_KEY").exists(_.length > 0)) {
+      val serviceAccountKey = sys.env("GOOGLE_SERVICE_ACCOUNT_KEY")
+      val fileWriter = new FileWriter(new File("/tmp/google_service_account_key.json"))
```
zsxwing, line 52: nit: Since we set the env `GOOGLE_APPLICATION_CREDENTIALS`, it's better to read it from sys.env rather than using the hard code file path.

zsxwing (commented):
Left some minor style comments. Otherwise LGTM

zhuansunxt:
@zsxwing All comments addressed and integration tests are running correctly for server, spark and python client. Please take a final look.

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSuite.scala
```diff
@@ -16,7 +16,7 @@
 
 package io.delta.sharing.spark
 
-import java.io.EOFException
+import java.io.{EOFException, File, FileWriter}
```
zsxwing, line 19: nit: unused new imports

zsxwing (approved):
LGTM

PR #107, opened 2022-01-10
Make DeltaSharingProfile classes public

zhuansunxt:
The implementation of the DeltaSharingProfileProvider might not live in the same package as io.delta.sharing.spark. Marking the trait as public so that it becomes accessible in other packages.

Also rephrase the documentation a bit in PROTOCOL.md about recipient profile.


zsxwing (approved):
LGTM

PR #108, opened 2022-01-12
Trigger the hack for yarl only if the urls are GCS pre-signed urls

zsxwing:
This PR moves the hack for yarl into `_yarl_patch.py` and import it only if the urls are GCS pre-signed urls. This will avoid modifying yarl when using S3 or Azure.


zhuansunxt (approved):
LGTM, thanks!

Issue #110, opened 2022-01-18
Python 3.10


jccaleroe:
I can't install on Python 3.10, it seems this version of Python only fixed a bug, so I do not see the problem of restricting to version < 3.10.
I changed the restriction and install it on Python 3.10 and all worked perfectly splendid

zsxwing:
Good catch. Feel free to submit a PR to remove the restriction.

zsxwing:
Fixed by #111

PR #111, opened 2022-01-20
Add query_table_version to the rest client

### LINKED ISSUE #82: Expose table version in metadata to make polling for [data] updates faster & easier
Looking for table version # (or timestamp) so that we can build an airflow sensor to easily detect if a new version of the data is now available.

image

nkvuong:
This is proposed in #82 

Also remove the restriction for Python 3.10 #110 

Closes #82 and #110


.gitignore
```diff
@@ -106,3 +106,12 @@ spark-warehouse/
...
 *.egg-info
+
+# For VSCode
+*.vscode
+
+# For Metals
+*.metals
+
+# For venv
+*.venv
```
zhuansunxt, line 117: pls have a new line at EOF

python/delta_sharing/rest_client.py
```diff
@@ -253,6 +266,15 @@ def _get_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
     def _post_internal(self, target: str, data: Optional[Dict[str, Any]] = None):
         return self._request_internal(request=self._session.post, target=target, json=data)
 
+    def _head_internal(self, target: str):
+        assert target.startswith("/"), "Targets should start with '/'"
```
zhuansunxt, line 270: Can we do what other two methods are doing, e.g `return self._request_internal(request=self._session.head, target=target)`

python/delta_sharing/tests/test_rest_client.py
```diff
@@ -205,6 +205,14 @@ def test_query_table_metadata_partitioned_different_schemas(rest_client: DataSha
     )
 
 
+@pytest.mark.skipif(not ENABLE_INTEGRATION, reason=SKIP_MESSAGE)
+def test_query_table_version(rest_client: DataSharingRestClient):
+    response = rest_client.query_table_version(
+        Table(name="table1", share="share1", schema="default")
+    )
+    assert response.delta_table_version == 28
```
zhuansunxt, line 213: This hardcoded value is decided by the current version of the table and if that table is updated this test will break, so maybe just assert some integer value is defined.

python/delta_sharing/rest_client.py
nkvuong: `_request_internal` returns only the request content, but we only need the headers from the request, hence a new function

python/delta_sharing/tests/test_rest_client.py
nkvuong: added int assertion and that table version is positive

python/setup.py
```diff
@@ -40,7 +40,7 @@
     packages=[
         'delta_sharing',
     ],
-    python_requires='>=3.6,<3.10',
+    python_requires='>=3.6,<=3.10',
```
zsxwing, line 43: Could you remove `,<=3.10` to make us support future Python versions?

python/delta_sharing/rest_client.py
```diff
@@ -219,6 +224,14 @@ def query_table_metadata(self, table: Table) -> QueryTableMetadataResponse:
                 metadata=Metadata.from_json(metadata_json["metaData"]),
             )
 
+    @retry_with_exponential_backoff
+    def query_table_version(self, table: Table) -> QueryTableVersionResponse:
+        headers = self._head_internal(
+            f"/shares/{table.share}/schemas/{table.schema}/tables/{table.name}"
+        )
+        table_version = int(headers.get("delta-table-version", None) or -1)
```
zsxwing, line 232: Could you raise an error when the header is missing? This is a required header defined in the protocol. If a server doesn't return it, it should be a bug in the server and it's better to raise an error.

nkvuong: if the table does not exist on the server, should the server still return a delta-table-version in the header? Happy to raise a LookupError when the header is missing

.gitignore
nkvuong: done

python/delta_sharing/rest_client.py
zsxwing: Good point. This reminds me that we are missing `response.raise_for_status()` in `_head_internal`. We should fail if the server turns non 200 response. Could you also add a test for this?

zsxwing (approved):
LGTM. Thanks!

Issue #71, opened 2021-10-27
Improve the error message in the client when a share/schema/table doesn't exist


zsxwing:
When a share/schema/table doesn't exist, we currently report http 404 error. It would be great if our client can wrap the error with a user friendly message such as `[Share/Schema/Table] XXX does not exist, please contact your share provider for further information`.

PR #120, opened 2022-03-12
Improve the error message for the /share/schema/table APIs

kohei-tosshy:
Hello.

This is a PR for #71.

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -64,10 +64,7 @@ class DeltaSharingServiceExceptionHandler extends ExceptionHandlerFunction {
       case _: DeltaSharingNoSuchElementException =>
         if (req.method().equals(HttpMethod.HEAD)) {
           HttpResponse.of(
-            ResponseHeaders.builder(HttpStatus.NOT_FOUND)
-              .set("error-code", ErrorCode.RESOURCE_DOES_NOT_EXIST)
-              .set("message", cause.getMessage)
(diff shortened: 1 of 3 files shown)


server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -62,13 +62,22 @@ class DeltaSharingServiceExceptionHandler extends ExceptionHandlerFunction {
...
-          MediaType.JSON_UTF_8,
-          JsonUtils.toJson(
-            Map(
-              "errorCode" -> ErrorCode.RESOURCE_DOES_NOT_EXIST,
-              "message" -> cause.getMessage)))
+        if (req.method().equals(HttpMethod.HEAD)) {
+          HttpResponse.of(
+            ResponseHeaders.builder(HttpStatus.NOT_FOUND)
+              .set("error-code", ErrorCode.RESOURCE_DOES_NOT_EXIST)
+              .set("message", cause.getMessage)
```
zsxwing, line 69: perhaps not change these codes? I think it's not worth to add these non standard headers for this small change.

zsxwing (commented):
Thanks for doing this! Left one minor comment.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
kohei-tosshy: Thank you for you comment.
I think that your opinion is correct.
I removed this two headers.

zsxwing (approved):
LGTM. Thanks!

PR #121, opened 2022-03-13
Fix image link in README

alexott:
also, format examples a bit to avoid horizontal scrolling


README.md
```diff
@@ -13,7 +13,7 @@
 With Delta Sharing, a user accessing shared data can directly connect to it through pandas, Tableau, Apache Spark, Rust, or other systems that support the open protocol, without having to deploy a specific compute platform first. Data providers can share a dataset once to reach a broad range of consumers, while consumers can begin using the data in minutes.
 
 <p align="center">
-  <img src="https://delta.io/wp-content/uploads/2021/12/revised-sharing-hero@2x-1.png" width="85%"/>
+  <img src="https://delta.io/static/hero-desktop-372eb254dd239bf9c8ed259d0e909a02.png" width="85%"/>
```
zsxwing, line 16: Thanks for catching this!

@dennyglee Could you comment whether this is the right link?

dennyglee: Yes, this is the right link.

zsxwing: @dennyglee Will this link become invalid if you update the image next time? Is there a static link for https://github.com/delta-io/website/blob/main/src/components/pages/sharing/HeroSection/hero-desktop.png ?

dennyglee: Oh good point - in GitHub, we should perhaps use the link you used as we will continue to simply update that image.

The `https://delta.io/static/hero-desktop-372eb254dd239bf9c8ed259d0e909a02.png` image is generated by Netlify automatically every time we update the image, it appends the GUID for tracking/caching purposes.  So any images on delta.io itself will have URLs that change.  The github image name itself would not change though.

alexott:
Sure, let me do that

alexott:
@zsxwing done

zsxwing (approved):
LGTM. Thanks!

PR #124, opened 2022-03-28
Include response body in HTTPError exception

wchau:
The protocol includes the error message in the response body, which is not forwarded through the `HTTPError` exception. The exception format would be similar to how the Databricks CLI handled it.

Previously, the exception error message looks like:
`requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: <redacted>`

Now, the exception error message looks like:
```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: <redacted>
 Response from server: 
 { 'errorCode': 'PERMISSION_DENIED',
  'message': '<redacted>'}
```

### DIFF
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -21,6 +21,7 @@ from typing import Any, ClassVar, Dict, Generator, Optional, Sequence
 from urllib.parse import urlparse
 import time
 import logging
+import pprint
 from datetime import datetime
 
 import requests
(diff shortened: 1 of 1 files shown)


zsxwing:
I think `HTTPError.response.text` has the body. I'm not a Python expert so I'm not sure what's the convention in Python to handle `HTTPError`. But it seems weird to modify `HTTPError` when it has already the information.

wchau:
> I think HTTPError.response.text has the body. I'm not a Python expert so I'm not sure what's the convention in Python to handle HTTPError. But it seems weird to modify HTTPError when it already has the information.

https://github.com/databricks/databricks-cli/blob/master/databricks_cli/sdk/api_client.py#L143

Following this. The problem is that the default message does not include the response body, so even you crash, if you don't catch the request and extract the message manually, you would not why it failed. This provides better error messaging, especially for Python shell users importing the library.

wchau:
Note that the request and response fields would still be available in HTTPError for downstream exception handlers to use. It's just that the message that is modified.

zsxwing:
Yep I understand your change. My question was more about why Python requests doesn't put the response body into `HTTPError` message. I guess it cannot because it doesn't know what the response body could be. For our case, it should be fine.

zsxwing (approved):
LGTM

PR #135, opened 2022-05-10
OSS Delta Sharing Server: Adds api to accept cdf query

linzhou-db:
OSS Delta Sharing Server: Adds api to accept cdf query

- @Get("/shares/{share}/schemas/{schema}/tables/{table}/changes")
- Parse url parameters and construct the cdfoptions map
- Add classes DeltaErrors and DeltaDataSource for some exceptions and constants   
- Add `DeltaSharedTable.queryCDF` and return not implemented exception


zhuansunxt:
Hi @linzhou-db , could you have a standalone PR on https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md with the new API's specification? It's better to agree on specification before adding its implementation on the server IMO.

linzhou-db:
> Hi @linzhou-db , could you have a standalone PR on https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md with the new API's specification? It's better to agree on specification before adding its implementation on the server IMO.

sg, will do

linzhou-db:
> > Hi @linzhou-db , could you have a standalone PR on https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md with the new API's specification? It's better to agree on specification before adding its implementation on the server IMO.
> 
> sg, will do

Sent out https://github.com/delta-io/delta-sharing/pull/137

server/src/main/scala/io/delta/standalone/internal/DeltaErrors.scala
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone.internal
+
```
zhuansunxt, line 18: Should we put those inside a `DeltaErrors` object and make those functions?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -241,6 +244,37 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+      includeFiles = true,
+      getCdfOptionsMap(
+        Option(startingVersion),
+        Option(endingVersion),
+        Option(startingTimestamp),
+        Option(endingTimestamp)
+      )
+    )
+    logger.info(s"Took ${System.currentTimeMillis - start} ms to load the table cdf " +
+      s"and sign ${actions.length - 2} urls for table $share/$schema/$table")
```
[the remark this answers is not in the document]
linzhou-db: yes, this is copied from L243

server/src/main/scala/io/delta/standalone/internal/DeltaErrors.scala
linzhou-db: yes, will update.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -333,6 +368,72 @@ object DeltaSharingService {
...
+    }
+  }
+
+  private[server] def getCdfOptionsMap(
+    startingVersion: Option[String],
+    endingVersion: Option[String],
+    startingTimestamp: Option[String],
+    endingTimestamp: Option[String]): Map[String, String] = {
+    checkCDFOptionsValidity(startingVersion, endingVersion, startingTimestamp, endingTimestamp)
+    (if (startingVersion.isDefined) {
```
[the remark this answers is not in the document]
linzhou-db: I'll try.

server/src/main/scala/io/delta/standalone/internal/DeltaErrors.scala
linzhou-db: updated.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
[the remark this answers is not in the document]

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -343,6 +376,62 @@ object DeltaSharingService {
...
+    }
+    if (endingVersion.isDefined && endingTimestamp.isDefined) {
+      throw DeltaCDFErrors.multipleCDFBoundary("ending")
+    }
+    if (startingVersion.isEmpty && startingTimestamp.isEmpty) {
+      throw DeltaCDFErrors.noStartVersionForCDF
+    }
+    if (startingVersion.isDefined) {
+      try {
+        startingVersion.get.toLong
```
chakankardb, line 396: Should we validate timestamp?
If it gets validate elsewhere, perhaps add a comment.

server/src/main/scala/io/delta/standalone/internal/DeltaCDFErrors.scala
```diff
@@ -0,0 +1,29 @@
...
+package io.delta.standalone.internal
+
+object DeltaCDFErrors {
+  def multipleCDFBoundary(position: String): Throwable = {
+    new IllegalArgumentException(s"Multiple $position arguments provided for CDF read. Please " +
+      s"provide one of either ${position}Timestamp or ${position}Version."
+    )
+  }
+
+  def noStartVersionForCDF: Throwable = {
```
chakankardb, line 26: Should we call it noStartParamForCDF?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
linzhou-db: Added a comment, same as the code in our repo, it's validated in the cdc reader.

server/src/main/scala/io/delta/standalone/internal/DeltaCDFErrors.scala
linzhou-db: I more prefer to keep it the same as DeltaErrors in our code base.

chakankardb: sg.

chakankardb (approved):

linzhou-db:
@zhuansunxt you still want to take a look?

zhuansunxt (approved):
LGTM

Basem-Gaber:
Why is the concept of table changes coupled with the CDF ?

Isn't it possible to extract the info of the difference between a version and another by only knowing which files were added and which were removed ?

The API can return this metadata which should be sufficient for the client to use to replicate the changes, correct ?

linzhou-db:
Hi Basem, 
Thanks for your interest and question.


> Why is the concept of table changes coupled with the CDF ?

CDF is short for Change Data Feed, it IS the concept of table changes.


> Isn't it possible to extract the info of the difference between a version and another by only knowing which files were added and which were removed ?
> 
> The API can return this metadata which should be sufficient for the client to use to replicate the changes, correct ?

Technically yes, while CDF is a well developed, organized, and well maintained feature for this purpose, so good to just use it.

PR #137, opened 2022-05-10
Extends Delta Sharing PROTOCOL to support change data feed

linzhou-db:
Extends Delta Sharing PROTOCOL to support change data feed:

- Adds API: GET `{prefix}/shares/{share}/schemas/{schema}/tables/{table}/changes` 
- Extends API: POST `{prefix}/shares/{share}/schemas/{schema}/tables/{table}/query` with a new parameter `version`
- Extends Metadata with the `configuration` field, which is a map containing the configuration options of the table.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1944,9 +1944,12 @@ This is the API for clients to read change data feed from a table.
 
 The API supports a start parameter and and an end parameter. The start/end parameter can either be a version or a timestamp. The start parameter must be provided. If the end parameter is not provided, the API will use the latest table version for it. The parameter range is inclusive in the query.
 
-You specify a version as a long and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
+You specify a version as a Long and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
 
-More related info can be found in [Delta Change Data Feed](https://docs.databricks.com/delta/delta-change-data-feed.html).
+The change data feed represents row-level changes between versions of a Delta table. It records change data for UPDATE, DELETE, and MERGE operations. If you leveage this OSS connector to read chagne data feed, it contains three metadata columns that identify the type of change event, in addition to the data columns:
+- _change_type (type: String): There are four values: insert, update_preimage, update_postimage, delete. preimage is the value before the udpate, postimage is the value after the update.
+- _commit_version (type: Long): The table version containing the change.
+- _commit_timestamp (type: String): The timestap associated when the commit of the change was created, in the format "yyyy-mm-dd hh:mm:ss".
 
 
 HTTP Request | Value
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1887,7 +1887,7 @@ Example (See [Table Metadata Format](#table-metadata-format) for more details ab
     "date <= '2021-01-31'"
   ],
   "limitHint": 1000,
-  "version": 1
+  "version": 123
 }
 ```
 
@@ -1940,7 +1940,9 @@ delta-table-version: 123
 ```
 
 ### Read Change Data Feed from a Table
-This is the API for clients to read change data feed from a table. You can provide either version or timestamp for the start and end. The start and end versions and timestamps are inclusive in the queries. To read the changes from a particular start version to the latest version of the table, specify only the starting version or timestamp.
+This is the API for clients to read change data feed from a table.
+
+You can provide either version or timestamp for the start and end. The start and end versions and timestamps are inclusive in the queries. To read the changes from a particular start version to the latest version of the table, specify only the starting version or timestamp.
 
 You specify a version as an integer and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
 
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -1869,6 +1872,8 @@ The request body should be a JSON string containing the following two optional f
 - **limitHint** (type: Int32, optional): an optional limit number. It’s a hint from the client to tell the server how many rows in the table the client plans to read. The server can use this hint to return only some files by using the file stats. For example, when running `SELECT * FROM table LIMIT 1000`, the client can set `limitHint` to `1000`.
   - Applying `limitHint` is **BEST EFFORT**. The server may return files containing more rows than the client requests.
 
+- **version** (type: Int32, optional): an optional version number. If set, will return files as of the specified version of the table. This is only supported on tables with change data feed (cdf) enabled.
```
zhuansunxt, line 1875: I remember in your implementation PR this is a `long` (int64).

PROTOCOL.md
```diff
@@ -1933,6 +1939,283 @@ delta-table-version: 123
 }
 ```
 
+### Read Change Data Feed from a Table
+This is the API for clients to read change data feed from a table.
+
+You can provide either version or timestamp for the start and end. The start and end versions and timestamps are inclusive in the queries. To read the changes from a particular start version to the latest version of the table, specify only the starting version or timestamp.
```
zhuansunxt, line 1945: "for the start and end of change data feed"

PROTOCOL.md
```diff
@@ -1933,6 +1939,283 @@ delta-table-version: 123
...
+You specify a version as an integer and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
+
+
+HTTP Request | Value
+-|-
+Method | `GET`
+Header | `Authorization: Bearer {token}`
+URL | `{prefix}/shares/{share}/schemas/{schema}/tables/{table}/changes`
+URL Parameters | **{share}**: The share name to query. It's case-insensitive.<br>**{schema}**: The schema name to query. It's case-insensitive.<br>**{table}**: The table name to query. It's case-insensitive.
+Query Parameters | **startingVersion** (type: Int32, optional): The starting version of the query, inclusive. <br> **startingTimestamp** (type: String, optional): The starting timestamp of the query, will be converted to a version created greater or equal to this timestamp. <br> **endingVersion** (type: Int32, optional): The ending version of the query, inclusive. <br> **endingTimestamp** (type: String, optional): The ending timestamp of the query, will be converted to a version created earlier or equal to this timestamp.
```
zhuansunxt, line 1956: Similarly, we'd need to decide on int32 and int64. I did not find its specification in the delta lake protocol. @zsxwing

PROTOCOL.md
zhuansunxt, line 1945: Mention that one of either starting version or starting timestamp must be provided? All those query parameters are marked as optional below.

PROTOCOL.md
```diff
@@ -1933,6 +1939,283 @@ delta-table-version: 123
...
+
+Example (See [Table Metadata Format](#table-metadata-format) for more details about the format):
+
+`GET {prefix}/shares/vaccine_share/schemas/acme_vaccine_data/tables/vaccine_patients/changes?startingVersion=0&endingVersion=2`
+
+
+```
+HTTP/2 200 
+content-type: application/x-ndjson; charset=utf-8
+delta-table-version: 123
```
zhuansunxt, line 2156: Question: are we going to set this version header in the HTTP response?

PROTOCOL.md
```diff
@@ -2023,6 +2310,91 @@ Example (for illustration purposes; each JSON object must be a single line in th
 }
 ```
 
+### CDF Files
+
+#### Add File
```
zhuansunxt, line 2315: What's the difference between an Add file vs CDF file? By reading this spec it does not give me that information and the two looks identical.

PROTOCOL.md
linzhou-db: updated.

`limitHint` is actually also int64 in protocol. I was following it.

But I see that `Query Table Version` is returning table version as long, so int64 makes sense.

PROTOCOL.md
linzhou-db: updated.

linzhou-db: Added.

PROTOCOL.md
linzhou-db: All are int64, as of all other `table version` parameters.
cdf is not open source yet, so there's no reference.

PROTOCOL.md
linzhou-db: So I removed it as this is not needed, and according to the nature of the query, there's not a single version to return.

PROTOCOL.md
linzhou-db: Good question. For the fields, there are no differences, including `RemoveFile`.

The difference is for the connector to tell the [_change_type](https://docs.databricks.com/delta/delta-change-data-feed.html#change-data-event-schema) in the response. It would be `insert` for AddFile and `delete` for RemoveFile, and the value is included in the CDFFile for each row.

PROTOCOL.md
```diff
@@ -1933,6 +1939,282 @@ delta-table-version: 123
 }
 ```
 
+### Read Change Data Feed from a Table
+This is the API for clients to read change data feed from a table.
+
+You can provide either version or timestamp for the start and end of change data feed. The start and end versions and timestamps are inclusive in the queries. To read the changes from a particular start version to the latest version of the table, specify only the starting version or timestamp. Either startingVersion or startingTimestamp must be provided. For either starting or ending parameter, only one parameter (version or timestamp) is supported.
```
zhuansunxt, line 1945: "Either startingVersion or startingTimestamp must be provided." this reads a bit strange. How about "A starting parameter, either startingVersion or startingTimestamp, must be provided."?

PROTOCOL.md
zhuansunxt: Just for my education: connector will decide change_type based on "add" and "remove". How would the "cdf" file be interpreted? I did not quite follow what `and the value is included in the CDFFile for each row.` mean.

PROTOCOL.md
chakankardb: How about:
The API supports a start parameter and and an end parameter. The start/end parameter can either be a version or a timestamp. The start parameter must be provided. If the end parameter is not provided, the API will use the latest table version for it. The parameter range is inclusive in the query.

PROTOCOL.md
```diff
@@ -1933,6 +1939,282 @@ delta-table-version: 123
...
+
+You specify a version as an long and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
```
chakankardb, line 1947: an long -> a Long ?

PROTOCOL.md
```diff
@@ -1933,6 +1939,282 @@ delta-table-version: 123
...
+</tr>
+<tr>
+<td>Headers</td>
+<td>
+
+`Content-Type: application/x-ndjson; charset=utf-8`
+
+`Delta-Table-Version: {version}`
+
+**{version}** is a long value which represents the current table version.
```
chakankardb, line 1974: long value -> Long?

PROTOCOL.md
```diff
@@ -1933,6 +1939,284 @@ delta-table-version: 123
 }
 ```
 
+### Read Change Data Feed from a Table
+This is the API for clients to read change data feed from a table.
+
+The API supports a start parameter and and an end parameter. The start/end parameter can either be a version or a timestamp. The start parameter must be provided. If the end parameter is not provided, the API will use the latest table version for it. The parameter range is inclusive in the query.
+
+You specify a version as a long and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
```
chakankardb, line 1947: long -> Long?

chakankardb (approved):

PROTOCOL.md
```diff
@@ -1942,9 +1942,11 @@ delta-table-version: 123
...
 This is the API for clients to read change data feed from a table.
 
-You can provide either version or timestamp for the start and end of change data feed. The start and end versions and timestamps are inclusive in the queries. To read the changes from a particular start version to the latest version of the table, specify only the starting version or timestamp. Either startingVersion or startingTimestamp must be provided. For either starting or ending parameter, only one parameter (version or timestamp) is supported.
+The API supports a start parameter and and an end parameter. The start/end parameter can either be a version or a timestamp. The start parameter must be provided. If the end parameter is not provided, the API will use the latest table version for it. The parameter range is inclusive in the query.
 
-You specify a version as an long and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
+You specify a version as a long and a timestamps as a string in the format "yyyy-mm-dd hh:mm:ss[.fffffffff]".
+
+More related info can be found in [Delta Change Data Feed](https://docs.databricks.com/delta/delta-change-data-feed.html).
```
zhuansunxt, line 1949: I don't think we should have any link pointing to Databricks specific documentation. At least at this point that documentation still says CDF is only available on Databricks Runtime. It could make ppl think this feature is also somewhat proprietary but it's actually not.

chakankardb: This is an interesting point.
If Delta has not open sourced it yet, is it ok for us to open-source this feed via sharing?
Do we need to seek their permission?

PROTOCOL.md
linzhou-db: updated.

PROTOCOL.md
linzhou-db: We've already checked with @zsxwing, CDF will be open source in Q2/Q3.

Good point. I removed the link and added some explanation on the file content in the `CDF File` section.

chakankardb: sg!

chakankardb (approved):

linzhou-db:
@zhuansunxt you want to take another look when you get a time? no hurry.

zhuansunxt (approved):
LGTM

PR #138, opened 2022-05-11
Extends Delta Sharing Metadata with field: configuration

linzhou-db:
Extends Delta Sharing Metadata with field: configuration. It's a string to string map, which contains table properties.


zhuansunxt (approved):
LGTM. Seems like this overlaps with https://github.com/delta-io/delta-sharing/pull/135. Please resolve conflicts before merging.

linzhou-db:
> LGTM. Seems like this overlaps with #135. Please resolve conflicts before merging.

Good catch. Will do, thanks!

PR #141, opened 2022-05-12
Check TableConfig.cdfEnabled when queryTable has version parameter

linzhou-db:
Check TableConfig.cdfEnabled when queryTable has version parameter

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -238,8 +238,8 @@ class DeltaSharingService(serverConfig: ServerConfig) {
     val start = System.currentTimeMillis
     val tableConfig = sharedTableManager.getTable(share, schema, table)
     if (queryTableRequest.version.isDefined && !tableConfig.cdfEnabled) {
-      throw new DeltaSharingIllegalArgumentException("reading table by version is not enabled " +
-        s"on table $share.$schema.$table")
+      throw new DeltaSharingIllegalArgumentException("reading table by version is not supported" +
+        s" because change data feed is not enabled on table: $share.$schema.$table")
     }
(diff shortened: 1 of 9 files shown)


zhuansunxt:
Should this be part of the implementation of supporting `version` parameter in QueryTable API?

linzhou-db:
> Should this be part of the implementation of supporting `version` parameter in QueryTable API?

You are right, but that one is already merged...

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -237,6 +237,10 @@ class DeltaSharingService(serverConfig: ServerConfig) {
 
     val start = System.currentTimeMillis
     val tableConfig = sharedTableManager.getTable(share, schema, table)
+    if (queryTableRequest.version.isDefined && !tableConfig.cdfEnabled) {
+      throw new DeltaSharingIllegalArgumentException("reading table by version is not enabled " +
```
zhuansunxt, line 241: Better to explain the root cause here: e.g. reading table by version is not supported because the table is not enabled with change data feed.

zhuansunxt (approved):
LGTM

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
chakankardb: +1

chakankardb (approved):

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala

Basem-Gaber:
Why is reading table be versions limited only to tables with CDF enabled ?


Isn't it possible to extract the info of the difference between a version and another by only knowing which files were added and which were removed ?
Supposedly replicating the metadata I am mentioning should lead to the same of effect of the CDF, correct ?

PR #145, opened 2022-05-19
[SC-99477] Support readChangeFeed in DeltaSharingDataSource

linzhou-db:
Support readChangeFeed in DeltaSharingDataSource, with a new case class RemoteDeltaCDFRelation, DeltaSharingClient.getCDFFiles is not implemented yet.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
@@ -18,6 +18,9 @@ package io.delta.sharing.spark
 
 import java.util.Collections
 
+import scala.collection.JavaConverters._
+import scala.collection.mutable
+
 import org.apache.spark.SparkEnv
 import org.apache.spark.delta.sharing.PreSignedUrlCache
 import org.apache.spark.sql.{SparkSession, SQLContext}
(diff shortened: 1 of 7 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
```diff
@@ -37,13 +40,37 @@ private[sharing] class DeltaSharingDataSource extends RelationProvider with Data
...
+          DeltaSharingDataSource.CDF_START_TIMESTAMP_KEY)
+      }
+      if (caseInsensitiveParams.containsKey(DeltaSharingDataSource.CDF_END_VERSION_KEY)) {
+        cdfOptions(DeltaSharingDataSource.CDF_END_VERSION_KEY) = caseInsensitiveParams.get(
+          DeltaSharingDataSource.CDF_END_VERSION_KEY)
+      }
+      if (caseInsensitiveParams.containsKey(DeltaSharingDataSource.CDF_END_TIMESTAMP_KEY)) {
+        cdfOptions(DeltaSharingDataSource.CDF_END_TIMESTAMP_KEY) = caseInsensitiveParams.get(
+          DeltaSharingDataSource.CDF_END_TIMESTAMP_KEY)
+      }
```
chakankardb, line 63: Do we need to validate these, or is it expected that they are already validated?

chakankardb (approved):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
linzhou-db: do you mean the unit test? There's no unit test for this class only, I will add integration test once all code are in place.

chakankardb: No I meant validating the options (for example, only one of startVersion/startTimestamp is set).

linzhou-db: I see. I didn't add it because I was thinking of leaving this check to the server. Should we check it on the connector as well? Maybe we can skip?

chakankardb: Ok, let the server do the checks.

PR #146, opened 2022-05-20
Implement DeltaSharingClient.getCDFFiles

linzhou-db:
- Implement DeltaSharingClient.getCDFFiles
- Extends Metadata with configuration
- Exposes DeltaStandaloneException from the OSS Delta Sharing Server as client error

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -30,6 +30,7 @@ import com.linecorp.armeria.internal.server.ResponseConversionUtil
 import com.linecorp.armeria.server.{Server, ServiceRequestContext}
 import com.linecorp.armeria.server.annotation.{ConsumesJson, Default, ExceptionHandler, ExceptionHandlerFunction, Get, Head, Param, Post, ProducesJson}
 import com.linecorp.armeria.server.auth.AuthService
+import io.delta.standalone.exceptions.DeltaStandaloneException
 import io.delta.standalone.internal.DeltaCDFErrors
 import io.delta.standalone.internal.DeltaCDFIllegalArgumentException
 import io.delta.standalone.internal.DeltaDataSource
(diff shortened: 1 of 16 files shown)


spark/src/main/scala/io/delta/sharing/spark/model.scala
```diff
@@ -75,14 +89,65 @@ private[sharing] case class Protocol(minReaderVersion: Int) extends Action {
...
 
+sealed abstract class AddFileBase(
+    url: String,
+    id: String,
+    @JsonInclude(JsonInclude.Include.ALWAYS)
+    partitionValues: Map[String, String],
+    size: Long,
+    @JsonRawValue
+    stats: String = null)
+    extends Action {}
```
chakankardb, line 100: There is some conflict here with changes in my PR:
https://github.com/delta-io/delta-sharing/pull/149

Perhaps wait for that PR before we merge this one?

linzhou-db: sg, will do

chakankardb (approved):

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -94,6 +95,14 @@ class DeltaSharingServiceExceptionHandler extends ExceptionHandlerFunction {
             Map(
               "errorCode" -> ErrorCode.INVALID_PARAMETER_VALUE,
               "message" -> cause.getMessage)))
+      case _: DeltaStandaloneException =>
+        HttpResponse.of(
+          HttpStatus.BAD_REQUEST,
```
chakankardb, line 100: Is it safe to always map delta standalone exceptions to be bad requests?

linzhou-db: In this case it is a bad request instead of an internal error. If we do not expose all `DeltaStandaloneException`, then we have to do a special match for this case, which also works, but don't feel it's obviously better than the current solution.

Will quickly sync with Ryan tomorrow to get more context.

chakankardb: I added a more specific check in: https://github.com/delta-io/delta-sharing/pull/150

PR #150, opened 2022-05-24
Python rest client changes for CDF.

chakankardb:
This PR has the following changes:
1) Implemented list_table_changes in the python rest client (made the method a POST on server side).
2) Added integ test.

### DIFF
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -281,12 +281,11 @@ class DataSharingRestClient:
         # We do not validate the CDF options here since the server will perform validations anyways.
         if cdfOptions.starting_version is not None:
             query_str += f"startingVersion={cdfOptions.starting_version}"
-        elif cdfOptions.starting_timestamp is not None:
+        if cdfOptions.starting_timestamp is not None:
             query_str += f"startingTimestamp={cdfOptions.starting_timestamp}"
-
         if cdfOptions.ending_version is not None:
             query_str += f"&endingVersion={cdfOptions.ending_version}"
-        elif cdfOptions.ending_timestamp is not None:
+        if cdfOptions.ending_timestamp is not None:
             query_str += f"&endingTimestamp={cdfOptions.ending_timestamp}"
 
         with self._get_internal(query_str) as lines:
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -18,7 +18,7 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 import json
 from typing import Any, ClassVar, Dict, Generator, List, Optional, Sequence
-from urllib.parse import urlparse
+from urllib.parse import quote, urlparse
 import time
 import logging
 import pprint
@@ -279,14 +279,16 @@ class DataSharingRestClient:
 
         # Add cdf options.
         # We do not validate the CDF options here since the server will perform validations anyways.
+        params = []
         if cdfOptions.starting_version is not None:
-            query_str += f"startingVersion={cdfOptions.starting_version}"
+            params.append(f"startingVersion={cdfOptions.starting_version}")
         if cdfOptions.starting_timestamp is not None:
-            query_str += f"startingTimestamp={cdfOptions.starting_timestamp}"
+            params.append(f"startingTimestamp={quote(cdfOptions.starting_timestamp)}")
         if cdfOptions.ending_version is not None:
-            query_str += f"&endingVersion={cdfOptions.ending_version}"
+            params.append(f"endingVersion={cdfOptions.ending_version}")
         if cdfOptions.ending_timestamp is not None:
-            query_str += f"&endingTimestamp={cdfOptions.ending_timestamp}"
+            params.append(f"endingTimestamp={quote(cdfOptions.ending_timestamp)}")
+        query_str += "&".join(params)
 
         with self._get_internal(query_str) as lines:
             protocol_json = json.loads(next(lines))
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -17,7 +17,7 @@ import collections
 from contextlib import contextmanager
 from dataclasses import dataclass
 import json
-from typing import Any, ClassVar, Dict, Generator, Optional, Sequence
+from typing import Any, ClassVar, Dict, Generator, List, Optional, Sequence
 from urllib.parse import urlparse
 import time
 import logging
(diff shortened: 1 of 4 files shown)


server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -263,7 +263,7 @@ class DeltaSharingService(serverConfig: ServerConfig) {
     streamingOutput(Some(version), actions)
   }
 
-  @Get("/shares/{share}/schemas/{schema}/tables/{table}/changes")
+  @Post("/shares/{share}/schemas/{schema}/tables/{table}/changes")
```
chakankardb, line 266: This needs to be a POST for the streaming output to work.

python/delta_sharing/rest_client.py
```diff
@@ -273,8 +273,34 @@ def list_files_in_table(
...
             )
 
+    @retry_with_exponential_backoff
     def list_table_changes(self, table: Table, cdfOptions: CdfOptions) -> ListTableChangesResponse:
-        raise HTTPError("API not implemented yet!")
+        query_str = f"/shares/{table.share}/schemas/{table.schema}/tables/{table.name}/changes?"
+
+        # Add cdf options.
+        # We do not validate the CDF options here since the server will perform validations anyways.
+        if cdfOptions.starting_version is not None:
```
linzhou-db, line 282: L282 to L290 is still assuming that one and only one starting parameter will be set, are you intended to do so?

I prefer we do not make such assumption? 

Also for starting_timestamp and ending_timestamp, we need to encode them for the server to be able to parse, something I found on web:

```
import urllib.parse
urllib.parse.quote(query)
```

python/delta_sharing/tests/test_rest_client.py
```diff
@@ -385,11 +387,95 @@ def test_list_files_in_table_partitioned_different_schemas(
...
-    try:
-        rest_client.list_table_changes(Table(name="x", share="y", schema="z"), CdfOptions())
-        assert False
-    except Exception as e:
-        assert isinstance(e, HTTPError)
+    # The following table query will return all types of actions.
+    cdf_table = Table(name="cdf_table_with_partition", share="share1", schema="default")
+    response = rest_client.list_table_changes(
+        cdf_table,
+        CdfOptions(starting_version=0, ending_version=3)
```
linzhou-db, line 398: Better to add a test case with `starting_timestamp`, to just ensure it's correctly encoded(so that the server is able to parse it). 

But one thing to note is that, seems that when running tests locally the server timezone is PST, and when github runs tests the server timezone is GMT. So the same string `startingTimestamp` will be parsed as different timestamp (long), which will map to different version of the table, thus results in different testing behavior.

One trick I did is use a very large startingTimestamp and expect an error message: in https://github.com/delta-io/delta-sharing/pull/146 , DeltaSharingRestClientSuite:`getCDFFiles with Timestamp`,

linzhou-db:
The error is transient, it's gone..

python/delta_sharing/rest_client.py
chakankardb: Removed the assumption.

I tried your suggestion about urllib, but the server does not like it and it returns an error.

python/delta_sharing/tests/test_rest_client.py
chakankardb: Added a test for start timestamp.

python/delta_sharing/rest_client.py
linzhou-db: hmmm, I think we need to do sth like this, so that `&` is correctly added.

```
params_list = []
if cdfOptions.starting_version is not None:
  params_list.append(f"startingVersion={cdfOptions.starting_version}")
query_str = "&".join(params_list))
```

For the urllib, I think we need and only need to apply `urllib.parse.quote` on cdfOptions.starting_timestamp, so that it doesn't encode `=`, and it encodes the space ` `. 
This is due the the server behavior, verified in both Delta Sharing OSS server and our rpc parser: https://github.com/databricks/universe/pull/174415

And needs to add a corresponding test case to test the encoding behavior is correct in our client.

python/delta_sharing/tests/test_rest_client.py
linzhou-db: As mentioned, needs to add a corresponding test case to test the encoding behavior is correct in our client, could you separate it out, and expect an error, similar to https://github.com/delta-io/delta-sharing/pull/146?

Though agree with you, we need to discuss whether it's safe to always expose `DeltaStandaloneException`, but in this case, we have to expose it because it's not internal error, and for the user/client to know that they need to set correct timestamp.

python/delta_sharing/rest_client.py
chakankardb: Done... good catch on the params!

python/delta_sharing/tests/test_rest_client.py
chakankardb: Done.
Also added a more specific exception mapping in the reader, which will be safer.

server/src/main/scala/io/delta/standalone/internal/DeltaSharingCDCReader.scala
```diff
@@ -139,13 +140,28 @@ class DeltaSharingCDCReader(val deltaLog: DeltaLogImpl, val conf: Configuration)
...
     }
   }
 
+  /**
+   * A wrapper around DeltaHistory manager method.
+   * The method implicitly validates the timestamp arg, so we remap its exception to CDF
+   * illegal arg exception.
+   */
+   private def getActiveCommitAtTime(ts: Timestamp): DeltaSharingHistoryManager.Commit = {
+     val commit = try {
```
linzhou-db, line 156: If you sync with the head, you don't need this change? The main concern is that history.getActiveCommitAtTime(ts) may throw other exceptions which may not be `DeltaCDFIllegalArgumentException`.

Or do you still prefer this change over what I did here? https://github.com/delta-io/delta-sharing/blob/main/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala#L98

Neither is obviously better since both are not precisely catching and throwing the error.

linzhou-db (approved):
LGTM with a minor comment on the server exception handling.

server/src/main/scala/io/delta/standalone/internal/DeltaSharingCDCReader.scala
chakankardb: The scope of getActiveCommitAt is a lot smaller, and looking at the code it will most commonly only throw timestamp related exception. So it is safer imho.

The other one seems too broad, so we should perhaps remove that one until we find a better solution.

PR #152, opened 2022-05-26
spark connector changes to hook up CDF relation to the file indices.

chakankardb:
This PR has the following changes:
1) Hook up CDF relation to the file indices.
2) Integration tests.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
@@ -19,27 +19,55 @@ package io.delta.sharing.spark
 import scala.collection.mutable.ListBuffer
 
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.{DataFrame, Row, SQLContext}
+import org.apache.spark.sql.{DataFrame, DeltaSharingScanUtils, Row, SparkSession, SQLContext}
+import org.apache.spark.sql.execution.LogicalRDD
+import org.apache.spark.sql.execution.datasources.{HadoopFsRelation, LogicalRelation}
 import org.apache.spark.sql.sources.{BaseRelation, Filter, PrunedFilteredScan}
 import org.apache.spark.sql.types.StructType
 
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
```diff
@@ -19,27 +19,55 @@ package io.delta.sharing.spark
...
 case class RemoteDeltaCDFRelation(
-    schema: StructType,
-    sqlContext: SQLContext,
+    spark: SparkSession,
+    snapshotToUse: RemoteSnapshot,
     client: DeltaSharingClient,
     table: DeltaSharingTable,
     cdfOptions: Map[String, String]) extends BaseRelation with PrunedFilteredScan {
 
+  override def schema: StructType = DeltaTableUtils.addCdcSchema(snapshotToUse.schema)
```
linzhou-db, line 37: nit: Not sure where this schema is used. 
We are using both the `snapshotToUse.schema` and the `metadata.schemaString`(L66) in this class, which implicitly assumes that they are the same.

linzhou-db (approved):

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
chakankardb: During debugging, I saw that it is called. 
It is called before the changes are fetched.

I'm not sure if it is actually used to create the DF, since we the fileIndex also has the schema.
My guess is that it may not be, and the fileIndex's schema should be used (since we scan that).
I followed what the CDC reader in DBR does here.

linzhou-db: sg, thanks!

chakankardb: Ok, I just verified. It is actually used.
If I do not add the cdc columns to the schema, spark will simply filter those out.
We should use the versionOf to fetch the right schema.

PR #153, opened 2022-05-26
Support load_table_changes_as_pandas in python connector

linzhou-db:
Support load_table_changes_as_pandas in python connector.
Also updated the order, to ensure `_commit_version` is always before `_commit_timestamp`

### DIFF
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -19,6 +19,8 @@ from pathlib import Path
 
 import pandas as pd
 
+from delta_sharing.protocol import CdfOptions
+
 try:
     from pyspark.sql import DataFrame as PySparkDataFrame
 except ImportError:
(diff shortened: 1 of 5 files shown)


python/delta_sharing/delta_sharing.py
```diff
@@ -96,6 +98,36 @@ def load_as_spark(url: str) -> "PySparkDataFrame":  # noqa: F821
...
+) -> pd.DataFrame:
+    """
+    Load the table changes of shared table as a pandas DataFrame using the give url.
+
+    :param url: a url under the format "<profile>#<share>.<schema>.<table>"
+    :param starting_version: The starting version of table changes
+    :param ending_version: The ending version of table changes
+    :param starting_timestamp: The starting timestamp of table changes
+    :param ending_timestamp: The ending timestamp of table changes
+    :return: A pandas DataFrame representing the shared table.
```
chakankardb, line 116: Similar comment as the other, since this is public facing, we should perhaps add some comments on valid combinations.

python/delta_sharing/delta_sharing.py
```diff
@@ -96,6 +98,36 @@ def load_as_spark(url: str) -> "PySparkDataFrame":  # noqa: F821
...
 
+def load_table_changes_as_pandas(
+    url: str,
+    starting_version: Optional[int] = None,
+    ending_version: Optional[int] = None,
+    starting_timestamp: Optional[str] = None,
+    ending_timestamp: Optional[str] = None
+) -> pd.DataFrame:
+    """
+    Load the table changes of shared table as a pandas DataFrame using the give url.
```
chakankardb, line 109: give url -> given url?

linzhou-db: updated.

python/delta_sharing/delta_sharing.py
linzhou-db: Added the comment.

chakankardb (approved):

PR #162, opened 2022-06-15
Support for per-share bearer token configs

atgabe:
At the moment authorization is everything or nothing, with a single bearer token providing access to the entire server. This PR expands authorization to support per-share bearer token configs.

- changes the `authorization.bearerToken` property to `authorization.universalBearerToken`, and adds an `authorization.shares` array to configure additional tokens unique to a share
- adds an AuthManager trait and two implementations (BearerTokenAuth and NoAuth) to handle authorization
- expands unit tests to cover per-share authorization cases

There are certainly other approaches to this feature, so let me know if an alternative approach is deemed preferable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,18 +302,43 @@ Replace `KEY_PATH` with path of the JSON file that contains your service account
 
 ## Authorization
 
-The server supports a basic authorization with pre-configed bearer token. You can add the following config to your server yaml file:
-
+The server supports authorization with bearer tokens. For basic authorization, you can add the following config to your server yaml file:
 ```yaml
 authorization:
-  bearerToken: <token>
+  universalBearerToken: <token>
 ```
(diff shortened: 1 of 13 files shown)


zhuansunxt:
An alternative approach on this is to introduce `Recipient` as a first class concept, each with its own token and support fine-grained `Share <-> Recipient` access control. @atgabe Do you think that approach satisfy your needs?

atgabe:
@zhuansunxt yup, that approach would work for our use case and sounds like it more closely mirrors the Databricks strategy. Is that approach in development elsewhere or a change you'd like to see on this PR?

zhuansunxt:
@atgabe It would be great to submit an issue and use it to describe the general approach (like an RFC) so we can agree on design decisions.

PR #168, opened 2022-07-05
Add ability to configure threshold for delta sharing JAR network timeouts

patrickjmccauley:
This allows users to configure an environmental variable `SPARK_DELTA_SHARING_NETWORK_TIMEOUT_SECONDS` to override the default 120s timeout threshold for the delta sharing JAR

Signed-off-by: Patrick McCauley [patrickjmccauley@gmail.com](mailto:patrickjmccauley@gmail.com)


zsxwing:
Thanks for the contribution. In general, we prefer to use Spark configs (this is the standard way in Spark) rather than system envs. Could you explain what's the motivation?

patrickjmccauley:
> Thanks for the contribution. In general, we prefer to use Spark configs (this is the standard way in Spark) rather than system envs. Could you explain what's the motivation?

My mistake, I'm pretty new to Spark and didn't realize that was the convention. I'll close this and resubmit if it still seems useful.

PR #170, opened 2022-07-12
Upgrade AWS SDK to 1.12.189

zsxwing:
The current AWS SDK used by the server doesn't support [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html). This PR just upgrades it to a newer version 1.12.189 to support IMDSv2.


zhuansunxt:
Currently do we have a dependency on IMDS on the OSS server?

zsxwing:
https://github.com/delta-io/delta-sharing#ec2-iam-metadata-authentication-recommended requires Instance Metadata Service.

zhuansunxt (approved):
LGTM

zsxwing:
For the context, an EC2 instance can be forced to IMDSv2 today (be more secure) and such instance cannot config https://github.com/delta-io/delta-sharing#ec2-iam-metadata-authentication-recommended without the SDK upgrade.

PR #176, opened 2022-07-20
Update DeltaSharedTableLoader.scala

airizarryDB:
For S3 regions that only support Signature Version 4, a protocol for authenticating inbound API requests to AWS services, the the default S3 service endpoint will not work, the S3A client needs to be given the endpoint to use via the fs.s3a.endpoint property


zsxwing:
Closing this. As I mentioned in #177, this can be set by a hadoop config file.

PR #179, opened 2022-08-11
Document that tables should use s3a paths rather than s3 paths

zsxwing:
`s3://` paths usually are configured to use EMR's proprietary file system which we don't support. This PR asks users to use `s3a://` which should be configured to use Hadoop S3AFileSystem.


zhuansunxt (approved):
LGTM

PR #181, opened 2022-08-25
Fix GCS path parsing

zsxwing:
For a path such as `gs://delta_sharing_test/foo`, `getHost` will return null and trigger NPE. This PR changes the code to call GCS APIs to parse a path instead to avoid such issue (It calls `getAuthority` underlying instead).

This is not an issue for S3 or Azure because they do require the bucket name to be a valid host name (cannot contain `_`).

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
+++ b/server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
@@ -22,6 +22,7 @@ import java.util.concurrent.TimeUnit.SECONDS
 
 import com.amazonaws.HttpMethod
 import com.amazonaws.services.s3.model.GeneratePresignedUrlRequest
+import com.google.cloud.hadoop.gcsio.StorageResourceId
 import com.google.cloud.storage.BlobId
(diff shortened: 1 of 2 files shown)


server/src/main/scala/io/delta/sharing/server/CloudFileSigner.scala
```diff
@@ -202,13 +203,18 @@ class GCSFileSigner(
...
     assert(objectName.nonEmpty, s"cannot get object key from $path")
     val blobInfo = BlobInfo.newBuilder(BlobId.of(bucketName, objectName)).build
     storage.signUrl(
       blobInfo, preSignedUrlTimeoutSeconds, SECONDS, Storage.SignUrlOption.withV4Signature())
       .toString
   }
 }
+
+object GCSFileSigner {
+  def getBucketAndObjectNames(path: Path): (String, String) = {
```
zsxwing, line 216: Create a new method for this so that I can write tests

zhuansunxt (approved):
LGTM. Thanks!

PR #186, opened 2022-09-22
Support timestamp in table query request in oss delta sharing server

linzhou-db:
Support timestamp in table query request in oss delta sharing server, only one of timestamp and version is supported, and they are only supported when cdfEnabled is true.

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -251,6 +251,7 @@ class DeltaSharingService(serverConfig: ServerConfig) {
       includeFiles = false,
       Nil,
       None,
+      None,
       None)
     streamingOutput(Some(version), actions)
   }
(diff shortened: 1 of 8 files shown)


server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -268,12 +269,13 @@ class DeltaSharingService(serverConfig: ServerConfig) {
 
     val start = System.currentTimeMillis
     val tableConfig = sharedTableManager.getTable(share, schema, table)
-    if (queryTableRequest.version.isDefined) {
+    if (queryTableRequest.version.isDefined || queryTableRequest.timestamp.isDefined) {
       if (!tableConfig.cdfEnabled) {
-        throw new DeltaSharingIllegalArgumentException("Reading table by version is not supported" +
-          s" because change data feed is not enabled on table: $share.$schema.$table")
+        throw new DeltaSharingIllegalArgumentException("Reading table by version or timestamp is" +
+          " not supported because change data feed is not enabled on table: " +
```
chakankardb, line 275: Question: With streaming, we will allow time travel queries even if cdf is not enabled... right?

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -133,10 +133,19 @@ class DeltaSharedTable(
...
       predicateHints: Seq[String],
       limitHint: Option[Long],
-      version: Option[Long]): (Long, Seq[model.SingleAction]) = withClassLoader {
+      version: Option[Long],
+      timestamp: Option[String]): (Long, Seq[model.SingleAction]) = withClassLoader {
     // TODO Support `limitHint`
-    val snapshot = if (version.isEmpty) deltaLog.snapshot else {
+    if (version.isDefined && timestamp.isDefined) {
+      throw new IllegalArgumentException("Please either provide '<version>' or '<timestamp>'")
+    }
```
chakankardb, line 141: Should we add a unit test for this?

chakankardb (approved):

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
linzhou-db: Yes. We'll have another field to guard this, pending on the decision about default behavior change.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
linzhou-db: hmmm, missed again..

linzhou-db: I'll add it in #187

PR #187, opened 2022-09-23
Support timestampAsOf in DeltaSharingDataSource

linzhou-db:
1. Support timestampAsOf in DeltaSharingDataSource
2. As part of 1., support timestampAsOf in DeltaSharingClient and RemoteDeltaLog
3. Upgrade dependency on [delta-standalone to 0.5.0](https://github.com/delta-io/connectors/releases), this is particularly leverage a fix in `deltaLog.getSnapshotForTimestampAsOf`: the error type on timestamp out of range is consistently `IllegalArgumentException` so we can translate it to a 4xx error from the OSS Delta Sharing Server, instead of 5xx.

### DIFF
--- a/server/src/main/scala/io/delta/standalone/internal/DeltaSharingHistoryManager.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/DeltaSharingHistoryManager.scala
@@ -21,7 +21,7 @@ import java.sql.Timestamp
 
 import io.delta.standalone.internal.actions.CommitMarker
 import io.delta.standalone.internal.util.FileNames
-import io.delta.standalone.storage.LogStore
+import io.delta.storage.LogStore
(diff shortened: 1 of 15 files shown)


server/src/main/scala/io/delta/standalone/internal/DeltaSharingHistoryManager.scala
```diff
@@ -21,7 +21,7 @@ import java.sql.Timestamp
 
 import io.delta.standalone.internal.actions.CommitMarker
 import io.delta.standalone.internal.util.FileNames
-import io.delta.standalone.storage.LogStore
+import io.delta.storage.LogStore
```
linzhou-db, line 24: This is changed because of the upgrade of dependency on delta-standalone to 0.5.0

chakankardb (approved):

PR #188, opened 2022-09-26
Support timestamp in queryTable in python connector

linzhou-db:
- Support timestamp in load_as_pandas
- Support timestamp in load_as_spark

### DIFF
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -54,7 +54,8 @@ def _parse_url(url: str) -> Tuple[str, str, str, str]:
 def load_as_pandas(
     url: str,
     limit: Optional[int] = None,
-    version: Optional[int] = None
+    version: Optional[int] = None,
+    timestamp: Optional[str] = None
 ) -> pd.DataFrame:
     """
     Load the shared table using the given url as a pandas DataFrame.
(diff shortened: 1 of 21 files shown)


python/delta_sharing/delta_sharing.py
```diff
@@ -73,17 +74,24 @@ def load_as_pandas(
...
+) -> "PySparkDataFrame":  # noqa: F821
     """
     Load the shared table using the given url as a Spark DataFrame. `PySpark` must be installed,
     and the application must be a PySpark application with the Apache Spark Connector for Delta
     Sharing installed.
 
     :param url: a url under the format "<profile>#<share>.<schema>.<table>".
     :param version: an optional non-negative int. Load the snapshot of table at version.
+    :param timestamp: an optional string. Load the snapshot of table at version corresponding
+      to the timestamp.
```
chakankardb, line 94: Should we add comment that at most one of version/timestamp should be specified?

python/delta_sharing/delta_sharing.py
```diff
@@ -99,6 +107,8 @@ def load_as_spark(url: str, version: Optional[int] = None) -> "PySparkDataFrame"
     df = spark.read.format("deltaSharing")
     if version is not None:
         df.option("versionAsOf", version)
+    if timestamp is not None:
+        df.option("timestampAsOf", timestamp)
```
chakankardb, line 111: Question: If the caller specifies both, are we relying on the server to return error?

chakankardb (approved):

python/delta_sharing/delta_sharing.py
linzhou-db: added.

python/delta_sharing/delta_sharing.py
linzhou-db: there's a test case in test_delta_sharing.py L623, it's checked in the deltaSharingDataSource, still at the client side.

PR #189, opened 2022-09-28
Support startingVersion in queryTable rpc in OSS Delta Sharing Server

linzhou-db:
Support startingVersion in queryTable rpc in OSS Delta Sharing Server, to query data change since startingVersion until the current version of the table.

Also, converted InvalidProtocolVersionException to be a client error instead of internal error in oss delta sharing server.

### DIFF
--- a/server/src/main/protobuf/protocol.proto
+++ b/server/src/main/protobuf/protocol.proto
@@ -46,12 +46,18 @@ message QueryTableRequest {
     repeated string predicateHints = 1;
     optional int64 limitHint = 2;
 
-    // If neither version nor timestamp is specified, the query is for the latest version.
-    // Only one of the two parameters can be supported in a single query.
+    // Only one of the three parameters can be supported in a single query.
+    // If none of them is specified, the query is for the latest version.
+    //
+    // - If either version or timestamp is specified, the query is for the snapshot at the version.
+    // - If startingVersion is specified, the query is for all dataChange files until the
+    //   current version.
     // The table version being queried.
     optional int64 version = 3;
     // The table version corresponding to the timestamp being queried.
     optional string timestamp = 4;
+    // Query all data change files since startingVersion
+    optional int64 startingVersion = 5;
 }
 
 message ListSharesResponse {
--- a/server/src/main/protobuf/protocol.proto
+++ b/server/src/main/protobuf/protocol.proto
@@ -50,8 +50,8 @@ message QueryTableRequest {
     // If none of them is specified, the query is for the latest version.
     //
     // - If either version or timestamp is specified, the query is for the snapshot at the version.
-    // - If startingVersion is specified, the query is for all dataChange files until the
-    //   current version.
+    // - If startingVersion is specified, the query is for all dataChange files from startingVersion
+    //   until the current version.
     // The table version being queried.
     optional int64 version = 3;
     // The table version corresponding to the timestamp being queried.
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -46,6 +46,7 @@ import io.delta.sharing.server.protocol._
 import io.delta.sharing.server.util.JsonUtils
 
 object ErrorCode {
+  val UNSUPPORTED_OPERATION = "UNSUPPORTED_OPERATION"
   val INTERNAL_ERROR = "INTERNAL_ERROR"
   val RESOURCE_DOES_NOT_EXIST = "RESOURCE_DOES_NOT_EXIST"
   val INVALID_PARAMETER_VALUE = "INVALID_PARAMETER_VALUE"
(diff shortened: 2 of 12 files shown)


server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -262,31 +273,43 @@ class DeltaSharingService(serverConfig: ServerConfig) {
       @Param("share") share: String,
       @Param("schema") schema: String,
       @Param("table") table: String,
-      queryTableRequest: QueryTableRequest): HttpResponse = processRequest {
-    if (queryTableRequest.version.isDefined && queryTableRequest.version.get < 0) {
+      request: QueryTableRequest): HttpResponse = processRequest {
+    val numVersionParams = Seq(request.version, request.timestamp, request.startingVersion)
+      .filter(_.isDefined).size
+    if (numVersionParams >= 2) {
```
chakankardb, line 279: Nit: consider making this > 1, as that matches the comments better.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -262,31 +273,43 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+    }
 
     val start = System.currentTimeMillis
     val tableConfig = sharedTableManager.getTable(share, schema, table)
-    if (queryTableRequest.version.isDefined || queryTableRequest.timestamp.isDefined) {
+    if (numVersionParams > 0) {
       if (!tableConfig.cdfEnabled) {
         throw new DeltaSharingIllegalArgumentException("Reading table by version or timestamp is" +
           " not supported because change data feed is not enabled on table: " +
           s"$share.$schema.$table")
```
chakankardb, line 297: Do we need to change this? Will streaming will be permitted on non-cdf-enabled tables?

server/src/main/protobuf/protocol.proto
```diff
@@ -46,12 +46,18 @@ message QueryTableRequest {
     repeated string predicateHints = 1;
     optional int64 limitHint = 2;
 
-    // If neither version nor timestamp is specified, the query is for the latest version.
-    // Only one of the two parameters can be supported in a single query.
+    // Only one of the three parameters can be supported in a single query.
+    // If none of them is specified, the query is for the latest version.
+    //
+    // - If either version or timestamp is specified, the query is for the snapshot at the version.
+    // - If startingVersion is specified, the query is for all dataChange files until the
```
chakankardb, line 53: all dataChange files -> all dataChange files from startingVersion until ....

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -131,3 +131,8 @@ case class RemoveFile(
 
   override def wrap: SingleAction = SingleAction(remove = this)
 }
+
+object Action {
+  /** The maximum version of the protocol that this version of Delta Standalone understands. */
+  val readerVersion = 1
```
chakankardb, line 137: call this maxReaderVersion?

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -288,6 +367,13 @@ class DeltaSharedTable(
     deltaLog.update()
   }
 
+  private def protocolRead(protocol: Protocol): Unit = {
```
chakankardb, line 370: Should this be called: validateProtocol ?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
linzhou-db: Streaming will be permitted on non-cdf-enabled tables, still need some guard.
And we'll changes this once we finalize on the default behavior change, since with that decision, the guard field  will change. It's sort of orthogonal to the streaming functionality, so will change it in a separate PR.

server/src/main/scala/io/delta/sharing/server/model.scala
linzhou-db: updated.

server/src/main/protobuf/protocol.proto
linzhou-db: updated.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
linzhou-db: updated as `assertProtocolRead` to be consistent with this: https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala#L212

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
chakankardb: Consider adding a todo for now?

chakankardb (approved):

PR #190, opened 2022-09-29
Support startVersion in DeltaSharingClient

linzhou-db:
Support startVersion in DeltaSharingClient: adding a new function `getFiles` just for this purpose, which is cleaner than sharing the original getFiles for version/timestamp. This function should later be directly used in DeltaSharingSource.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -55,6 +55,8 @@ private[sharing] trait DeltaSharingClient {
     versionAsOf: Option[Long],
     timestampAsOf: Option[String]): DeltaTableFiles
 
+  def getFiles(table: Table, startingVersion: Long): DeltaTableFiles
+
   def getCDFFiles(table: Table, cdfOptions: Map[String, String]): DeltaTableFiles
 }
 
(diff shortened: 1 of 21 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -215,6 +218,33 @@ private[spark] class DeltaSharingRestClient(
...
+    val encodedSchemaName = URLEncoder.encode(table.schema, "UTF-8")
+    val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
+    val target = getTargetUrl(
+      s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/query")
+    val (version, lines) = getNDJson(
+      target, QueryTableRequest(Nil, None, None, None, Some(startingVersion)))
+    val protocol = JsonUtils.fromJson[SingleAction](lines(0)).protocol
+    checkProtocol(protocol)
+    val metadata = JsonUtils.fromJson[SingleAction](lines(1)).metaData
+    val addFiles = ArrayBuffer[AddFileForCDF]()
```
chakankardb, line 232: Question: Should we eventually merge AddFile and AddFileForCDF into AddFile?

chakankardb (approved):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: The only difference is that AddFileForCDF requires `version` and `timestamp` field.
To merge into one protocol:
Pro:
- One less protocol, to represent duplicated file type.
Con:
- A breaking change to the protocol, since AddFileForCDF requires `version` and `timestamp` field.
- If we make the two fields optional, which should work. But they are required for cdf query and queryTable with startingVersion.

Worth considering but I feel it's ok to leave them as is.

PR #192, opened 2022-10-07
Add a Custom Http Header Provider

zhuansunxt:
So that the spark connector can be forked and customized to send arbitrary HTTP headers to any specific implementation of Delta Sharing Server.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -35,7 +35,6 @@ import org.apache.http.conn.ssl.{SSLConnectionSocketFactory, SSLContextBuilder, 
 import org.apache.http.entity.StringEntity
 import org.apache.http.impl.client.{HttpClientBuilder, HttpClients}
 import org.apache.spark.internal.Logging
-import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
 import io.delta.sharing.spark.model._
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -340,10 +340,18 @@ private[spark] class DeltaSharingRestClient(
   private def getResponse(httpRequest: HttpRequestBase): (Option[Long], String) =
     RetryUtils.runWithExponentialBackoff(numRetries) {
       val profile = profileProvider.getProfile
+      val customeHeaders = profileProvider.getCustomHeaders
+      if (customeHeaders.contains(HttpHeaders.AUTHORIZATION)
+          || customeHeaders.contains(HttpHeaders.USER_AGENT)) {
+        throw new IllegalArgumentException(
+          s"HTTP header ${HttpHeaders.AUTHORIZATION} and ${HttpHeaders.USER_AGENT} cannot be"
+            + "overriden."
+        )
+      }
       val headers = Map(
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -64,6 +65,20 @@ private[sharing] trait PaginationResponse {
...
+ * A provider that provides an custom set of HTTP headers to be sent as part of the HTTP request.
+ *
+ * Extend this interface to specify extra HTTP headers for the Delta Sharing server.
+ */
+trait CustomHttpHeadersProvider {
+  def getHeaders: Map[String, String]
+}
+
+private [sharing] class BaseCustomHttpHeadersProvider extends CustomHttpHeadersProvider {
+  // By default there are not custom headers to be set.
```
wchau, line 78: "there are no"

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -333,15 +348,31 @@ private[spark] class DeltaSharingRestClient(
     }
   }
 
+  private[sharing] def getHttpHeaders(profile: DeltaSharingProfile): Map[String, String] = {
+    // Load the custom http header provider and get custom headers if there are any.
+    val customHttpHeadersProviderClass =
+      SparkSession.active.sessionState.conf
+        .getConfString("spark.delta.sharing.customhttpheaders.provider.class",
+        "io.delta.sharing.spark.BaseCustomHttpHeadersProvider")
+    val customeHttpHeadersProvider =
```
wchau, line 357: typo (custome -> custom)

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -333,15 +348,31 @@ private[spark] class DeltaSharingRestClient(
...
+      Class.forName(customHttpHeadersProviderClass)
+        .getConstructor()
+        .newInstance()
+        .asInstanceOf[CustomHttpHeadersProvider]
+    Map(
+      HttpHeaders.AUTHORIZATION -> s"Bearer ${profile.bearerToken}",
+      HttpHeaders.USER_AGENT -> DeltaSharingRestClient.USER_AGENT
+    ) ++ customeHttpHeadersProvider.getHeaders
```
wchau, line 365: Should we check that the custom headers do not conflict with some important headers?

wchau (changes requested):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
zhuansunxt: reverted this code

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
zhuansunxt: [HTTP RFC](https://stackoverflow.com/questions/4371328/are-duplicate-http-response-headers-acceptable) says multiple headers with the same key is allowed. Let's conform with the standard and allow the user to do whatever is legal.

zsxwing (approved):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
wchau: When you combine maps like that in Scala, the values get overridden by the custom headers.

How does the Delta Sharing server handle multiple AUTHORIZATION headers?

zhuansunxt: Ok -- added check that authz and user-agent can not be overridden.

PR #201, opened 2022-10-18
Adds DeltasharingSource

linzhou-db:
Adds DeltasharingSource, similar to DeltaSource, supports streaming addfiles, 

- local cache: sortedFetchedFiles
- 30 seconds interval is added for getTableVersion
- ignoreChanges/ignoreDeletes/maxFilesPerTrigger/maxBytesPerTrigger are supported

Not in this PR, will be supported in a separate PR
- readChangeFeed/readChangeData
- read compatibility check on schema
- SupportsTriggerAvailableNow
- startingTimestamp

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
@@ -26,67 +26,82 @@ import org.apache.spark.delta.sharing.PreSignedUrlCache
 import org.apache.spark.sql.{SparkSession, SQLContext}
 import org.apache.spark.sql.connector.catalog.{Table, TableCapability, TableProvider}
 import org.apache.spark.sql.connector.expressions.Transform
-import org.apache.spark.sql.sources.{BaseRelation, DataSourceRegister, RelationProvider}
+import org.apache.spark.sql.execution.streaming.Source
+import org.apache.spark.sql.sources.{
+  BaseRelation,
+  DataSourceRegister,
+  RelationProvider,
+  StreamSourceProvider
+}
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
 /** A DataSource V1 for integrating Delta into Spark SQL batch APIs. */
-private[sharing] class DeltaSharingDataSource extends RelationProvider with DataSourceRegister {
+private[sharing] class DeltaSharingDataSource
+  extends RelationProvider
+    with StreamSourceProvider
+    with DataSourceRegister {
 
   override def createRelation(
-      sqlContext: SQLContext,
-      parameters: Map[String, String]): BaseRelation = {
+    sqlContext: SQLContext,
+    parameters: Map[String, String]): BaseRelation = {
     DeltaSharingDataSource.setupFileSystem(sqlContext)
-    val path = parameters.getOrElse("path", throw new IllegalArgumentException(
-      "'path' is not specified. If you use SQL to create a Delta Sharing table, " +
-        "LOCATION must be specified"))
-
-    var cdfOptions: mutable.Map[String, String] = mutable.Map.empty
-    val caseInsensitiveParams = new CaseInsensitiveStringMap(parameters.asJava)
-    if (DeltaSharingDataSource.isCDFRead(caseInsensitiveParams)) {
-      cdfOptions = mutable.Map[String, String](DeltaSharingDataSource.CDF_ENABLED_KEY -> "true")
-      if (caseInsensitiveParams.containsKey(DeltaSharingDataSource.CDF_START_VERSION_KEY)) {
-        cdfOptions(DeltaSharingDataSource.CDF_START_VERSION_KEY) = caseInsensitiveParams.get(
-          DeltaSharingDataSource.CDF_START_VERSION_KEY)
-      }
-      if (caseInsensitiveParams.containsKey(DeltaSharingDataSource.CDF_START_TIMESTAMP_KEY)) {
-        cdfOptions(DeltaSharingDataSource.CDF_START_TIMESTAMP_KEY) = caseInsensitiveParams.get(
-          DeltaSharingDataSource.CDF_START_TIMESTAMP_KEY)
-      }
-      if (caseInsensitiveParams.containsKey(DeltaSharingDataSource.CDF_END_VERSION_KEY)) {
-        cdfOptions(DeltaSharingDataSource.CDF_END_VERSION_KEY) = caseInsensitiveParams.get(
-          DeltaSharingDataSource.CDF_END_VERSION_KEY)
-      }
-      if (caseInsensitiveParams.containsKey(DeltaSharingDataSource.CDF_END_TIMESTAMP_KEY)) {
-        cdfOptions(DeltaSharingDataSource.CDF_END_TIMESTAMP_KEY) = caseInsensitiveParams.get(
-          DeltaSharingDataSource.CDF_END_TIMESTAMP_KEY)
-      }
-    }
-
-    if (parameters.get("versionAsOf").isDefined && parameters.get("timestampAsOf").isDefined) {
-      throw new IllegalArgumentException("Please either provide 'versionAsOf' or 'timestampAsOf'.")
-    }
-
-    var versionAsOf: Option[Long] = None
-    if (parameters.get("versionAsOf").isDefined) {
-      try {
-        versionAsOf = Some(parameters.get("versionAsOf").get.toLong)
-      } catch {
-        case _: NumberFormatException =>
-          throw new IllegalArgumentException("versionAsOf is not a valid number.")
-      }
-    }
+    val options = new DeltaSharingOptions(parameters)
+    val path = options.options.getOrElse("path", throw DeltaSharingErrors.pathNotSpecifiedException)
 
     val deltaLog = RemoteDeltaLog(path)
-    deltaLog.createRelation(versionAsOf, parameters.get("timestampAsOf"), cdfOptions.toMap)
+    deltaLog.createRelation(options.versionAsOf, options.timestampAsOf, options.cdfOptions)
+  }
+
+  override def sourceSchema(
+    sqlContext: SQLContext,
+    schema: Option[StructType],
+    providerName: String,
+    parameters: Map[String, String]): (String, StructType) = {
+    if (schema.nonEmpty && schema.get.nonEmpty) {
+      throw DeltaSharingErrors.specifySchemaAtReadTimeException
+    }
+    val options = new DeltaSharingOptions(parameters)
+    val path = options.options.getOrElse("path", throw DeltaSharingErrors.pathNotSpecifiedException)
+
+    val deltaLog = RemoteDeltaLog(path)
+
+    if (options.isTimeTravel) {
+      throw DeltaSharingErrors.timeTravelNotSupportedException
+    }
+
+    val schemaToUse = deltaLog.snapshot().schema
+    if (schemaToUse.isEmpty) {
+      throw DeltaSharingErrors.schemaNotSetException
+    }
+    if (options.readChangeFeed) {
+      throw DeltaSharingErrors.CDFNotSupportedInStreaming
+    } else {
+      (shortName(), schemaToUse)
+    }
+  }
+
+  override def createSource(
+    sqlContext: SQLContext,
+    metadataPath: String,
+    schema: Option[StructType],
+    providerName: String,
+    parameters: Map[String, String]): Source = {
+    DeltaSharingDataSource.setupFileSystem(sqlContext)
+    if (schema.nonEmpty && schema.get.nonEmpty) {
+      throw DeltaSharingErrors.specifySchemaAtReadTimeException
+    }
+    val options = new DeltaSharingOptions(parameters)
+    val path = options.options.getOrElse("path", throw DeltaSharingErrors.pathNotSpecifiedException)
+    val deltaLog = RemoteDeltaLog(path)
+
+    DeltaSharingSource(SparkSession.active, deltaLog, options)
   }
 
   override def shortName: String = "deltaSharing"
 }
 
-
 private[sharing] object DeltaSharingDataSource {
-
   def setupFileSystem(sqlContext: SQLContext): Unit = {
     // We have put our class name in the `org.apache.hadoop.fs.FileSystem` resource file. However,
     // this file will be loaded only if the class `FileSystem` is loaded. Hence, it won't work when
@@ -96,21 +111,4 @@ private[sharing] object DeltaSharingDataSource {
       .setIfUnset("fs.delta-sharing.impl", "io.delta.sharing.spark.DeltaSharingFileSystem")
     PreSignedUrlCache.registerIfNeeded(SparkEnv.get)
   }
-
-  // Based on the read options passed it indicates whether the read was a cdf read or not.
-  def isCDFRead(options: CaseInsensitiveStringMap): Boolean = {
-    options.containsKey(DeltaSharingDataSource.CDF_ENABLED_KEY) &&
-      options.get(DeltaSharingDataSource.CDF_ENABLED_KEY) == "true"
-  }
-
-  // Constants for cdf parameters
-  final val CDF_ENABLED_KEY = "readChangeFeed"
-
-  final val CDF_START_VERSION_KEY = "startingVersion"
-
-  final val CDF_START_TIMESTAMP_KEY = "startingTimestamp"
-
-  final val CDF_END_VERSION_KEY = "endingVersion"
-
-  final val CDF_END_TIMESTAMP_KEY = "endingTimestamp"
 }
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
@@ -55,6 +55,7 @@ private[sharing] class DeltaSharingDataSource
     deltaLog.createRelation(options.versionAsOf, options.timestampAsOf, options.cdfOptions)
   }
 
+  // Returns the schema of the latest table snapshot.
   override def sourceSchema(
     sqlContext: SQLContext,
     schema: Option[StructType],
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
@@ -65,23 +65,21 @@ private[sharing] class DeltaSharingDataSource
       throw DeltaSharingErrors.specifySchemaAtReadTimeException
     }
     val options = new DeltaSharingOptions(parameters)
-    val path = options.options.getOrElse("path", throw DeltaSharingErrors.pathNotSpecifiedException)
-
-    val deltaLog = RemoteDeltaLog(path)
-
     if (options.isTimeTravel) {
       throw DeltaSharingErrors.timeTravelNotSupportedException
     }
+    if (options.readChangeFeed) {
+      throw DeltaSharingErrors.CDFNotSupportedInStreaming
+    }
 
+    val path = options.options.getOrElse("path", throw DeltaSharingErrors.pathNotSpecifiedException)
+    val deltaLog = RemoteDeltaLog(path)
     val schemaToUse = deltaLog.snapshot().schema
     if (schemaToUse.isEmpty) {
       throw DeltaSharingErrors.schemaNotSetException
     }
-    if (options.readChangeFeed) {
-      throw DeltaSharingErrors.CDFNotSupportedInStreaming
-    } else {
-      (shortName(), schemaToUse)
-    }
+
+    (shortName(), schemaToUse)
   }
 
   override def createSource(
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -1,0 +1,518 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+  }
+
+  /**
+   * Fetch the changes from delta sharing server starting from (fromVersion, fromIndex).
+   * The start point should not be included in the result, it should be consumed in the last batch.
+   *
+   * If sortedFetchedFiles is not empty, return directly.
+   * Else, fetch file changes from the delta sharing server and store them in sortedFetchedFiles.
+   */
+  protected def getFileChanges(
```
chakankardb, line 105: call this maybeGetFileChanges?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+        case (file, index) if (index > fromIndex) =>
+          appendToSortedFetchedFiles(
+            IndexedFile(fromVersion, index, file, isLast = (index + 1 == numFiles)))
+      }
+    } else {
+      // TODO(lin.zhou) return metadata for each version, and check schema read compatibility
+      val tableFiles = deltaLog.client.getFiles(deltaLog.table, fromVersion)
+      val addFiles = verifyStreamHygieneAndFilterAddFiles(tableFiles)
+      addFiles.groupBy(a => a.version).toSeq.sortWith(_._1 < _._1).foreach{
+        case (v, adds) =>
```
chakankardb, line 138: Nit: adds -> vAddFiles?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+          val numFiles = adds.size
+          adds.sortWith(sortAddFileForCDF).zipWithIndex.foreach{
+            case (add, index) if (v > fromVersion || index > fromIndex) =>
```
chakankardb, line 141: v > fromVersion || (v == fromVersion && index > fromIndex) ?

Ideally v >= fromVersion right? so we can add that check as well.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+                index,
+                AddFile(add.url, add.id, add.partitionValues, add.size, add.stats),
+                isLast = (index + 1 == numFiles))
+              )
+          }
+      }
+    }
+  }
+
+  protected def getLastFileChangeWithRateLimit(
```
chakankardb, line 153: Consider documenting this method.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+    }
+  }
+
+  /**
+   * Get the changes from startVersion, startIndex to the endOffset, and create DataFrame
+   * @param startVersion - calculated starting version
+   * @param startIndex - calculated starting index
+   * @param isStartingVersion - whether the stream has to return the initial snapshot or not
+   * @param endOffset - Offset that signifies the end of the stream.
+   * @return
```
chakankardb, line 187: incomplete comment?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+   */
+  protected def getFileChangesAndCreateDataFrame(
+    startVersion: Long,
+    startIndex: Long,
+    isStartingVersion: Boolean,
+    endOffset: DeltaSharingSourceOffset): DataFrame = {
+    if (options.readChangeFeed) {
+      throw DeltaSharingErrors.CDFNotSupportedInStreaming
+    } else {
```
chakankardb, line 196: Nit: Do we need to put the following code in else block?

if (..) {
  throw ...
}
...

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+          return lastFileChange
+        }
+      }
+
+      lastFileChange
+    }
+  }
+
+  /**
+   * Get the changes from startVersion, startIndex to the endOffset, and create DataFrame
```
chakankardb, line 182: Consider adding comments about side effects (how sortedFetchedFiles is manipulated).

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+   */
+  protected def getFileChanges(
+    fromVersion: Long,
+    fromIndex: Long,
+    isStartingVersion: Boolean): Unit = {
+    if (!sortedFetchedFiles.isEmpty) { return }
+
+    if (fromVersion > intervalGetTableVersion) {
+      return
+    }
```
chakankardb, line 113: Question: The getFiles call below may return files from a version which may be greater than intervalGetTableVersion... right? Is that ok? Consider adding some comments.

chakankardb (commented):
Some initial comments.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+          } else if (!groupedAddFiles.contains(version) && !ignoreDeletes) {
+            throw DeltaSharingErrors.deltaSourceIgnoreDeleteError(version)
+          }
+      }
+    }
+
+    tableFiles.addFiles
+  }
+
+  private def getStartingOffset(
```
chakankardb, line 352: consider documenting this function.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+  }
+
+  override def getOffset: Option[Offset] = {
+    throw new UnsupportedOperationException(
+      "latestOffset(Offset, ReadLimit) should be called instead of this method")
+  }
+
+  override def getBatch(startOffsetOption: Option[Offset], end: Offset): DataFrame = {
+    val endOffset = DeltaSharingSourceOffset(tableId, end)
+    previousOffset = endOffset // Proceed the offset, and for recovery,
```
chakankardb, line 389: The comment seems incomplete. Consider expanding on it.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+
+  protected val tableId = snapshot.metadata.id
+
+  private var previousOffset: DeltaSharingSourceOffset = null
+
+  protected def verifyStreamHygieneAndFilterAddFiles(
+    tableFiles: DeltaTableFiles): Seq[AddFileForCDF] = {
+    if (!tableFiles.removeFiles.isEmpty) {
+      val groupedRemoveFiles = tableFiles.removeFiles.groupBy(r => r.version)
+      val groupedAddFiles = tableFiles.addFiles.groupBy(a => a.version)
```
chakankardb, line 338: Question: It seems groupBy will do a lot more work than is needed... right? 
technically we don't really need groups for verification, we just need info such as whether a version has add and/or remove files... right?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+    startSourceVersion) = if (startOffsetOption.isEmpty) {
+      getStartingVersion match {
+        case Some(v) =>
+          // startingVersion is provided by the user
+          (v, -1L, false, None)
+
+        case _ =>
+          // startingVersion is NOT provided by the user
+          if (endOffset.isStartingVersion) {
+            // get all files in this version if endOffset is startingVersion
```
chakankardb, line 403: Nit: get -> Get?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+    )
+
+    createdDf
+  }
+
+  override def stop(): Unit = {}
+
+  override def toString(): String = s"DeltaSharingSource[${deltaLog.table.toString}]"
+
+  trait DeltaSharingSourceAdmissionBase { self: AdmissionLimits =>
```
chakankardb, line 436: Question: Why do we use this design pattern with a trait and a self directive?
Why not simply move this admit function into AdmissionsLimits itself?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+   * Build the latest offset based on the last indexedFile. The function also checks if latest
+   * version is valid by comparing with previous version.
+   * @param indexedFile The last indexed file used to build offset from.
+   * @param version Previous offset table version.
+   * @param isStartingVersion Whether previous offset is starting version or not.
+   */
+  private def buildOffsetFromIndexedFile(
+    indexedFile: IndexedFile,
+    version: Long,
+    isStartingVersion: Boolean): Option[DeltaSharingSourceOffset] = {
```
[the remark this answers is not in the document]
linzhou-db: updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: updated to Set.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
```diff
@@ -46,10 +55,54 @@ private[sharing] class DeltaSharingDataSource extends RelationProvider with Data
     deltaLog.createRelation(options.versionAsOf, options.timestampAsOf, options.cdfOptions)
   }
 
+  override def sourceSchema(
```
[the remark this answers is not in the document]
linzhou-db: added

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,566 @@
...
+}
+
+/**
+ * Base trait for the Delta Sharing Source, that contains methods that deal with
+ * getting changes from the delta sharing server.
+ * TODO(lin.zhou) Support SupportsTriggerAvailableNow
+ */
+trait DeltaSharingSourceBase extends Source
+  with SupportsAdmissionControl
+  with Logging { self: DeltaSharingSource =>
```
chakankardb, line 69: Same question: What is the use case for this design pattern where we use a trait with the self directive? What benefit does it provide?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,566 @@
...
+    lastIndexedFile: IndexedFile,
+    previousOffsetVersion: Long,
+    ispreviousOffsetStartingVersion: Boolean): Option[DeltaSharingSourceOffset] = {
+    val IndexedFile(v, i, _, isLastFileInVersion) = lastIndexedFile
+    assert(v >= previousOffsetVersion,
+      s"buildOffsetFromIndexedFile receives an invalid previousOffsetVersion: $v " +
+        s"(expected: >= $previousOffsetVersion), tableId: $tableId")
+
+    // If the last file in previous batch is the last file of that version, automatically bump
+    // to next version to skip accessing that version file altogether.
```
chakankardb, line 336: It is not clear from the comment where do we skip accessing the version (it is when we call maybeGetFiles?)
Consider expanding on it.

Is it true that we are using the fact that logically <previousVersion, lastIndex> == <nextVersion, -1> ?
If so, perhaps mention it to make it clearer.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,516 @@
...
+ * the same answer as a batch query that had processed the entire dataset at any given point.
+ */
+case class DeltaSharingSource(
+  spark: SparkSession,
+  deltaLog: RemoteDeltaLog,
+  options: DeltaSharingOptions)
+  extends DeltaSharingSourceBase {
+
+  /** A check on the source table that disallows deletes on the source data. */
+  private val ignoreChanges = options.ignoreChanges
```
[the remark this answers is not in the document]
chakankardb: sg

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,566 @@
...
+        Some(new AdmissionLimits(Some(composite.files.maxFiles()), composite.bytes.maxBytes))
+      case other => throw DeltaSharingErrors.unknownReadLimit(other.toString())
+    }
+  }
+
+  /**
+   * Extracts whether users provided the option to time travel a relation. If a query restarts from
+   * a checkpoint and the checkpoint has recorded the offset, this method should never been called.
+   */
+  protected lazy val getStartingVersion: Option[Long] = {
```
chakankardb, line 544: can this be private?
Same question for other protected methods.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,566 @@
...
+          // startingVersion is NOT provided by the user
+          if (endOffset.isStartingVersion) {
+            // Get all files in this version if endOffset is startingVersion
+            (endOffset.tableVersion, -1L, true, None)
+          } else {
+            assert(
+              endOffset.tableVersion > 0, s"invalid tableVersion in endOffset: $endOffset")
+            // Load from snapshot `endOffset.tableVersion - 1L` if endOffset is not
+            // startingVersion
+            (endOffset.tableVersion - 1L, -1L, true, None)
```
chakankardb, line 459: Question: 
Why do we load the snapshot from prev version? 
Do we want to use its metadata ?

linzhou-db: I don't think it's anything related to the metadata. The schema of the latest table version at the time the stream query starts is used for constructing the dataframe.

I copied this logic from DeltaSource, I guess:
It's more of a product behavior decision, when `getBatch` is called and startOffsetOption is not set, and endOffset is not a startingVersion, it means: 1) no startingVersion is provided, 2) the stream query is not starting from a checkpoint 3) it's the first `getBatch` after the stream query started, 4) endOffset is from a previous `latestOffset` call. 5) and `endOffset` is NOT startingVersion. 6) and given 1) in L410, some version is the startingVersion.

Now, it needs to decide where the starting version is, then I guess it's a prod decision to use (endOffset.tableVersion - 1), or maybe we should use the version we used in L410?

@zsxwing Do you have an idea on this? Thanks!

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: changed all of them to private.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,566 @@
...
+
+  // Serves as local cache to store all the files fetched from the delta sharing server.
+  // If not empty, will advance the offset and fetch data from this list based on the read limit.
+  // If empty, will try to load all possible new data files through delta sharing rpc to this list,
+  //   sorted by version and id.
+  protected var sortedFetchedFiles: Seq[IndexedFile] = Seq.empty
+
+  protected var lastGetVersionTimestamp: Long = -1
+  protected var lastQueriedTableVersion: Long = -1
+  private val QUERY_TABLE_VERSION_INTERVAL_MILLIS = 30000 // 30 seconds
```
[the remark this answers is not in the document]
linzhou-db: yes, if we need to increase it we need a release.

The thing is the two ways of configs that I know are both available to the end user who runs the stream query, which I think is not something we want. 

And yeah, calling this MIN, and add a param that can increase it, still it relies on the recipient who start the stream query to config the param, is that desirable?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -0,0 +1,471 @@
...
+      query.processAllAvailable()
+      val progress = query.recentProgress.filter(_.numInputRows != 0)
+      assert(progress.length === 4)
+      progress.foreach { p =>
+        assert(p.numInputRows === 1)
+      }
+    } finally {
+      query.stop()
+    }
+  }
```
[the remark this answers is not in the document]
linzhou-db: added

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -197,20 +184,23 @@ class DeltaSharingSourceSuite extends QueryTest
   /**
    * Test maxFilesPerTrigger and maxBytesPerTrigger
    */
-  integrationTest("maxFilesPerTrigger - success") {
-    val query = withStreamReaderAtVersion()
-      .option("maxFilesPerTrigger", "1")
-      .load().writeStream.format("console").start()
+  integrationTest("maxFilesPerTrigger - success with different values") {
+    Map(1 -> Seq(1, 1, 1, 1), 2 -> Seq(2, 2), 3 -> Seq(3, 1), 4 -> Seq(4), 5 -> Seq(4)).foreach{
```
chakankardb, line 188: Consider adding a comment on what's in the map.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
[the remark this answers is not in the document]
chakankardb: Let's start with this; we can add in future if needed.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+    def appendToSortedFetchedFiles(indexedFile: IndexedFile): Unit = {
+      sortedFetchedFiles = sortedFetchedFiles :+ indexedFile
+    }
+
+    if (isStartingVersion) {
+      // If isStartingVersion is true, it means to fetch the snapshot at the fromVersion, which may
+      // include table changes from previous versions.
+      val tableFiles = deltaLog.client.getFiles(deltaLog.table, Nil, None, Some(fromVersion), None)
```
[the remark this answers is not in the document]
chakankardb: As discussed, let's do a one-decision on this.
I do think that this remote stream should be a lot more resilient to provider side outages compared to a local stream.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
chakankardb: Consider adding to the comment that we do this to maintain same behavior as delta.

chakankardb (approved):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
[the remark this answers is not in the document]
linzhou-db: sg

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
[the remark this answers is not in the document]
linzhou-db: will do.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: added the comment, will merge this first and still confirm with Ryan later.

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
linzhou-db: added

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
zsxwing: This hack for the following case:

When we start a streaming query, we load the latest version X - 1. However, when `latestOffset` is called, we get an offset F in version X (new data got written in between)

Now `getBatch` is called with (startOffsetOption=None, end=the above offset). Since this is the first batch, we need to return the table data in version `X - 1` + the data in version X from offset 0 to offset F (Note: we should not return data after offset F). So we use (offset=-1, version = X-1, isStartingVersion=true) to load the table data in version `X - 1`.

chakankardb: @linzhou-db Consider adding Ryan's explanation as a comment.

linzhou-db: Thanks! I just want to clarify and confirm:
This only happens to the first latestOffset and getBatch call of a stream query without `startingVersion`, and the Offset returned by latestOffset and used as `end` in getBatch can only be 1 version away from the startingVersion returned [here](https://github.com/delta-io/delta-sharing/blob/main/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala#L407), because it is loading snapshot at startingVersion. 

I'll add the comment in another PR.

PR #204, opened 2022-10-26
Return metadata for each table version on queryTable with startingVersion 

linzhou-db:
Return metadata for each table version, if seen in delta log, on queryTable with startingVersion.

Also updated the share config for integration test, added share8, so that when adding new tables later, we don't need to update the test for share1.


chakankardb (commented):
Do we need a one-decision before moving forward with this?

linzhou-db:
> Do we need a one-decision before moving forward with this?

Added in the doc: https://docs.google.com/document/d/1BFAaZ4h5llE_7YzVO6cbROvOdtIA0KM4KVJbUY9Gs4s/edit#heading=h.nytfc746dvw8

chakankardb (approved):

PR #205, opened 2022-10-26
Adds SchemaUtils to delta sharing spark client to check readCompatibility of schemas for streaming requests

linzhou-db:
Adds SchemaUtils to delta sharing spark client to check readCompatibility of schemas for streaming requests.


linzhou-db:
@chakankardb Discussed with Ryan, we'll skip the schema check on server and let client do all the schema check.

spark/src/main/scala/io/delta/sharing/spark/util/SchemaUtils.scala
```diff
@@ -0,0 +1,101 @@
...
+    def toFieldMap(
+      fields: Seq[StructField],
+      caseSensitive: Boolean = false): Map[String, StructField] = {
+      val fieldMap = fields.map(field => field.name -> field).toMap
+      if (caseSensitive) {
+        fieldMap
+      } else {
+        CaseInsensitiveMap(fieldMap)
+      }
+    }
```
chakankardb, line 47: Do we use caseSensitive = true?
Should we avoid constructing a map first and then copying it into case insensitive form?

spark/src/main/scala/io/delta/sharing/spark/util/SchemaUtils.scala
```diff
@@ -0,0 +1,101 @@
...
+            isDatatypeReadCompatible(e.keyType, n.keyType) &&
+            isDatatypeReadCompatible(e.valueType, n.valueType)
+        case (a, b) => a == b
+      }
+    }
+
+    def isStructReadCompatible: Boolean = {
+      val existingFieldMap = toFieldMap(existingSchema)
+      // scalastyle:off caselocale
+      val existingFieldNames = existingSchema.fieldNames.map(_.toLowerCase).toSet
```
chakankardb, line 70: caseInsensitiveMap uses toLowerCase(Locale.ROOT) which makes it locale insensitive. Should we follow that?

spark/src/test/scala/io/delta/sharing/spark/util/SchemaUtilsSuite.scala
```diff
@@ -0,0 +1,174 @@
...
+      ("int", make(IntegerType)),
+      ("string", make(StringType)),
+      ("struct", make(new StructType().add("a", StringType))),
+      ("array", make(ArrayType(IntegerType))),
+      ("map", make(MapType(StringType, FloatType)))
+    )
+    test(s"change of datatype should fail read compatibility - $scenario") {
+      for (a <- schemas.keys; b <- schemas.keys if a != b) {
+        assert(!isReadCompatible(schemas(a), schemas(b)),
+          s"isReadCompatible should have failed for: ${schemas(a)}, ${schemas(b)}")
```
chakankardb, line 48: Should we also test that if keys are equal then they are compatible?

spark/src/main/scala/io/delta/sharing/spark/util/SchemaUtils.scala
linzhou-db: We don't use caseSensitive = true, I removed it.

I think constructing from a map is the only available constructor of CaseInsensitiveMap.

spark/src/test/scala/io/delta/sharing/spark/util/SchemaUtilsSuite.scala
linzhou-db: added

chakankardb (approved):

PR #207, opened 2022-11-04
Return version and timestamp for each table for queryTable with version/timestamp

linzhou-db:
This is to support delta sharing streaming on cdf, where timestamp is required as a column.

### DIFF
--- a/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
@@ -230,6 +230,15 @@ class DeltaSharedTable(
         // queryDataChangeSinceStartVersion.
         queryDataChangeSinceStartVersion(startingVersion.get)
       } else if (includeFiles) {
+        val timestampsByVersion = DeltaSharingHistoryManager.getTimestampsByVersion(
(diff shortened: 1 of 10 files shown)


server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -245,11 +254,22 @@ class DeltaSharedTable(
         filteredFilters.map { addFile =>
           val cloudPath = absolutePath(deltaLog.dataPath, addFile.path)
           val signedUrl = fileSigner.sign(cloudPath)
-          val modelAddFile = model.AddFile(url = signedUrl,
-            id = Hashing.md5().hashString(addFile.path, UTF_8).toString,
-            partitionValues = addFile.partitionValues,
-            size = addFile.size,
-            stats = addFile.stats)
+          val modelAddFile = if (Seq(version, timestamp).filter(_.isDefined).isEmpty) {
```
chakankardb, line 257: same comment as before (pull this out of the loop) ?

linzhou-db: updated.

chakankardb (approved):

PR #209, opened 2022-11-09
Add StreamingQuery in the UserAgent http header for streaming queries

linzhou-db:
Add StreamingQuery in the UserAgent http header for streaming queries

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -403,7 +403,7 @@ private[spark] class DeltaSharingRestClient(
 
   private def getUserAgent(): String = {
     DeltaSharingRestClient.USER_AGENT + (if (forStreaming) {
-      " StreamingQuery/1.0.0"
+      s" SparkStructuredStreaming/$STREAMING_VERSION"
     } else {
       ""
     })
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -58,6 +58,8 @@ private[sharing] trait DeltaSharingClient {
   def getFiles(table: Table, startingVersion: Long): DeltaTableFiles
 
   def getCDFFiles(table: Table, cdfOptions: Map[String, String]): DeltaTableFiles
+
+  def getForStreaming(): Boolean = false
 }
 
 private[sharing] trait PaginationResponse {
@@ -161,6 +163,8 @@ private[spark] class DeltaSharingRestClient(
     }
     tables
   }
+
+  override def getForStreaming(): Boolean = forStreaming
 
   override def getTableVersion(table: Table, startingTimestamp: Option[String] = None): Long = {
     val encodedShareName = URLEncoder.encode(table.share, "UTF-8")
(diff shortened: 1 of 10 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -396,6 +401,14 @@ private[spark] class DeltaSharingRestClient(
       }
     }
 
+  private def getUserAgent(): String = {
+    DeltaSharingRestClient.USER_AGENT + (if (forStreaming) {
+      " StreamingQuery/1.0.0"
```
wchau, line 406: SparkStructuredStreaming ?

wchau, line 406: Can the version number be a constant somewhere?

I wonder in what scenario would this version change... not sure if possible to pass this in as a variable somehow

linzhou-db: I actually don't know what's the best string to put here. So trying to put a feature+version.
This version could change, as we support more parameters(such as Trigger.AvailableNow, which is available in spark 3.3.0, but we are currently at 3.1.1, we are not bumping up the version for other reasons) in streaming.


@chakankardb Thinking about this more, this could be in the same place as what version of dv/cm is supported in the client(once we have them), I hope it to be in a different header, but I don't see a good label for this purpose in the list of headers: https://hc.apache.org/httpcomponents-core-5.2.x/current/httpcore5/apidocs/org/apache/hc/core5/http/HttpHeaders.html.

So maybe we can leverage the user agent header for this purpose, what do you think?

wchau (approved):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
chakankardb: Using UA for encoding supported features seems a bit odd. 
Let's discuss it more when we work on it.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -396,6 +401,14 @@ private[spark] class DeltaSharingRestClient(
       }
     }
 
+  private def getUserAgent(): String = {
```
chakankardb, line 404: Consider adding a comment on why we append streaming version.

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -51,7 +51,7 @@ class DeltaSharingSourceSuite extends QueryTest
   lazy val toNotNullTable = testProfileFile.getCanonicalPath +
     "#share8.default.streaming_null_to_notnull"
 
-  lazy val deltaLog = RemoteDeltaLog(tablePath)
+  lazy val deltaLog = RemoteDeltaLog(tablePath, forStreaming = true)
```
chakankardb, line 54: Is there a way to test that the request header has the streaming version?

chakankardb (approved):

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
linzhou-db: Added a test.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: added

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: I feel odd too, sg, let's revisit later.

PR #211, opened 2022-11-12
Non-breaking change: return version and timestamp for queryTable with version

linzhou-db:
Non-breaking change: return version and timestamp for queryTable with version: We are extending `AddFile` instead of switching to `AddFileForCDF`, so the line in the response still has key 'file' instead of 'add'. 

Verified that this is non-breaking to reserve the client change in another PR, so the integration test verified that old client works with new server response.

### DIFF
--- a/spark/src/test/scala/io/delta/sharing/spark/TestDeltaSharingClient.scala
+++ b/spark/src/test/scala/io/delta/sharing/spark/TestDeltaSharingClient.scala
@@ -62,25 +62,14 @@ class TestDeltaSharingClient(
     timestampAsOf: Option[String]): DeltaTableFiles = {
     limit.foreach(lim => TestDeltaSharingClient.limits = TestDeltaSharingClient.limits :+ lim)
 
-    if (versionAsOf.isDefined) {
-      val addFiles: Seq[AddFileForCDF] = Seq(
-        AddFileForCDF("f1.parquet", "f1", Map.empty, 0, 1, 1600000000L),
-        AddFileForCDF("f2.parquet", "f2", Map.empty, 0, 1, 1600000000L),
-        AddFileForCDF("f3.parquet", "f3", Map.empty, 0, 1, 1600000000L),
-        AddFileForCDF("f4.parquet", "f4", Map.empty, 0, 1, 1600000000L)
-      ).take(limit.getOrElse(4L).toInt)
+    val addFiles: Seq[AddFile] = Seq(
+      AddFile("f1.parquet", "f1", Map.empty, 0),
+      AddFile("f2.parquet", "f2", Map.empty, 0),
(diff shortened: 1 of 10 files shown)


linzhou-db:
cc @zhuansunxt

spark/src/test/scala/io/delta/sharing/spark/TestDeltaSharingClient.scala
```diff
@@ -62,25 +62,14 @@ class TestDeltaSharingClient(
...
-        AddFileForCDF("f1.parquet", "f1", Map.empty, 0, 1, 1600000000L),
-        AddFileForCDF("f2.parquet", "f2", Map.empty, 0, 1, 1600000000L),
-        AddFileForCDF("f3.parquet", "f3", Map.empty, 0, 1, 1600000000L),
-        AddFileForCDF("f4.parquet", "f4", Map.empty, 0, 1, 1600000000L)
-      ).take(limit.getOrElse(4L).toInt)
+    val addFiles: Seq[AddFile] = Seq(
+      AddFile("f1.parquet", "f1", Map.empty, 0),
+      AddFile("f2.parquet", "f2", Map.empty, 0),
+      AddFile("f3.parquet", "f3", Map.empty, 0),
+      AddFile("f4.parquet", "f4", Map.empty, 0)
```
chakankardb, line 69: Question: If versionAsOf is defined, should we not return some version/timestamp?

linzhou-db: You are actually right! it should

linzhou-db: o wait, it should, but in another PR. 

This AddFile is from spark/model.scala. This PR would like to demonstrate that with server side changes, and without client side changes, the test/existing clients won't break.

chakankardb: Question: Do we have bwd compatibility tests where we test new server changes with older clients?

linzhou-db: So this PR reverted the client changes in https://github.com/delta-io/delta-sharing/pull/207, so the integration test in test_rest_client.py and DeltaSharingRestClientSuite.scala are testing old clients with new servers.

chakankardb (approved):

spark/src/test/scala/io/delta/sharing/spark/TestDeltaSharingClient.scala
linzhou-db: I created an entry in the backlog spreadsheet.

PR #212, opened 2022-11-13
Return metadata for queryTableChanges, and use it for streaming

linzhou-db:
- Return metadata if seen in deltaLog, for queryTableChanges from delta sharing server
- Use additional metadata for schema compatibility check in spark structured streaming
- Ignore additional  metadata for other requests.
- Before we figure out how to get headers in http request in the oss server, oss connector on old versions will break for new oss server with this change.


python/delta_sharing/rest_client.py
```diff
@@ -330,7 +330,9 @@ def list_table_changes(self, table: Table, cdfOptions: CdfOptions) -> ListTableC
             metadata_json = json.loads(next(lines))
             actions: List[FileAction] = []
             for line in lines:
-                actions.append(FileAction.from_json(json.loads(line)))
+                file_action = FileAction.from_json(json.loads(line))
+                if (file_action is not None):
```
chakankardb, line 334: Consider adding a comment on when can this be None .

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -367,13 +367,31 @@ class DeltaSharedTable(
...
+          id = metadata.id,
+          name = metadata.name,
+          description = metadata.description,
+          format = model.Format(),
+          schemaString = cleanUpTableSchema(metadata.schemaString),
+          configuration = getMetadataConfiguration(metadata.configuration),
+          partitionColumns = metadata.partitionColumns,
+          version = cdcDataSpec.version
+        )
+        actions.append(modelMetadata.wrap)
```
chakankardb, line 392: Question: If we always return metadata actions for CDF queries, would it break older clients?

linzhou-db: Yes it will. 

There are 3 options:
Option 1: The current PR, client and server are both changed, and will be released together. New OSS server will break old client. But this could be resolved by letting the client upgrade to the latest release.

Option 2: Check http header user_agent to return metadata only for streaming query. This is the Databricks delta sharing server behavior. I'm not doing it here because I don't know how to get the httpRequest here: https://sourcegraph.com/github.com/delta-io/delta-sharing/-/blob/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala?L349

Option 3: Add a parameter in the http request which serve as the same purpose as http user_agent. So the server can be compatible with the older client.

I prefer option 1 or option 2. If I got enough time I would go for option 2. For now I feel option 1 is ok because it only break old clients with new oss server and there's a way to work around it. WDYT?

linzhou-db: Updated the PR with option 2. Now old client is compatible with new server.

python/delta_sharing/rest_client.py
linzhou-db: this change is reverted.

chakankardb (approved):

PR #214, opened 2022-11-14
Support new parameter includeHistoricalMetadata for queryTableChange RPC

linzhou-db:
A couple changes:

- Support new parameter includeHistoricalMetadata for queryTableChanges.
- Update the way SparkStructuredStreaming is put in user agent header.
- Added two more tests on service side to verify additional metadata is only returned for queryTableChanges from spark streaming.
- Update to real id in DeltaSharingRestClientSuite.scala


zsxwing:
Any reason you need to change these ids? They should be deterministic?

linzhou-db:
> Any reason you need to change these ids? They should be deterministic?

Not often, it's just during streaming development, I wanna change the table properties of one table that's already in test, instead of creating another one. And then broke the test, which makes me wonder the id doesn't "have to" be fixed, we are more testing the e2e workflow and functionality.
I'm ok either way. If you prefer keeping the ids, I'll just fix that single test, and revert other changes in the PR.

chakankardb:
> 

I don't have a strong opinion... will you and Ryan decide.

zhuansunxt:
I agree with Ryan that ID should be deterministic. If we need a table with different config we should just create a new one for testing.

linzhou-db:
I reverted the id changes, just fixing the failed tests.

linzhou-db:
Hi all,
This change now is less about "id" in test, but more about streaming logic change. Will look forward more for approval from @chakankardb

spark/src/test/scala/io/delta/sharing/spark/TestDeltaSharingClient.scala
```diff
@@ -62,12 +62,21 @@ class TestDeltaSharingClient(
     timestampAsOf: Option[String]): DeltaTableFiles = {
     limit.foreach(lim => TestDeltaSharingClient.limits = TestDeltaSharingClient.limits :+ lim)
 
-    val addFiles: Seq[AddFile] = Seq(
-      AddFile("f1.parquet", "f1", Map.empty, 0),
-      AddFile("f2.parquet", "f2", Map.empty, 0),
-      AddFile("f3.parquet", "f3", Map.empty, 0),
-      AddFile("f4.parquet", "f4", Map.empty, 0)
-    ).take(limit.getOrElse(4L).toInt)
+    val addFiles: Seq[AddFile] = if (versionAsOf.isDefined) {
```
chakankardb, line 65: versionAsOf or timestampAsOf?

linzhou-db: updated and added a test.

spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
```diff
@@ -212,6 +212,75 @@ class RemoteDeltaLogSuite extends SparkFunSuite with SharedSparkSession {
...
+    val fileIndexLimit2 = {
+      val params = RemoteDeltaFileIndexParams(spark, snapshot)
+      RemoteDeltaSnapshotFileIndex(params, Some(2L))
+    }
+    val listFilesResult2 = fileIndexLimit2.listFiles(Seq.empty, Seq.empty)
+    assert(listFilesResult2.size == 1)
+    assert(listFilesResult2(0).files.size == 2)
+    assert(listFilesResult2(0).files(0).getPath.toString == "delta-sharing:/test/f1/0")
+    assert(listFilesResult2(0).files(1).getPath.toString == "delta-sharing:/test/f2/0")
+    client.clear()
```
chakankardb, line 281: Do we have validations that version and timestamp info is returned?

chakankardb (approved):

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -106,13 +106,35 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
...
+    url: String,
+    method: Option[String],
+    data: Option[String] = None,
+    expectedTableVersion: Option[Long] = None,
+    expectedContentType: String,
+    isStreamingQuery: Boolean = false): String = {
     val connection = new URL(url).openConnection().asInstanceOf[HttpsURLConnection]
     connection.setRequestProperty("Authorization", s"Bearer ${TestResource.testAuthorizationToken}")
+    if (isStreamingQuery) {
+      connection.setRequestProperty("User-Agent", "SparkStructuredStreaming")
```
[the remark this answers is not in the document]
linzhou-db: For the server to have different behavior based on the client(i.e., return metadata for queryTableChanges), in order to not break old client.

spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
linzhou-db: Reading through the [code here](https://sourcegraph.com/github.com/delta-io/delta-sharing/-/blob/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala?L76&subtree=true), only path is available in FileStatus. 

The purpose of this test is to verify that the RemoteDeltaSnapshotFileIndex works well with files with additional field returned from server.

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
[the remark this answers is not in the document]
linzhou-db: @chakankardb Had a discussion with Ryan, it's not a good idea to leverage user-agent for different server behavior on RPC request, I'll add a parameter such as 'returnMetadata' for queryTableChanges, and update both Databricks and OSS server/client.

user-agent can still be used for usage tracking, and gate DFF (as it will be removed eventually).

I'll update this PR and send it later.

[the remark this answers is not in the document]
chakankardb: @zsxwing Can you provide the reasoning here (for documentation)?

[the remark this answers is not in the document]
zsxwing: `User-Agent` is usually informational and it's weird to use it as a parameter to request different results. In addition, `SparkStructuredStreaming` is a special system, but our network protocol should be designed for all systems.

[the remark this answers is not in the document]
linzhou-db: Another point Ryan mentioned is that for column mapping, we may also need to pass metadata to client to render dataframe correctly, which is not a streaming only use case.

linzhou-db:
The PR is updated with what we discussed, please review again, thanks!

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -359,28 +360,29 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
-      @Header("User-Agent") @Nullable userAgent: String,
       @Param("share") share: String,
       @Param("schema") schema: String,
       @Param("table") table: String,
       @Param("startingVersion") @Nullable startingVersion: String,
       @Param("endingVersion") @Nullable endingVersion: String,
       @Param("startingTimestamp") @Nullable startingTimestamp: String,
-      @Param("endingTimestamp") @Nullable endingTimestamp: String): HttpResponse = processRequest {
+      @Param("endingTimestamp") @Nullable endingTimestamp: String,
+      @Param("returnMetadata") @Nullable returnMetadata: String): HttpResponse = processRequest {
```
chakankardb, line 370: should we call this returnMetadataChanges since queryTableChanges already returns the latest metadata ?

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -426,7 +442,7 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
   integrationTest("getTableVersion - get exceptions") {
     // timestamp can be any string here, it's resolved in DeltaSharedTableLoader
     assertHttpError(
-      url = requestPath("/shares/share2/schemas/default/tables/table2?startingTimestamp=abc"),
+      url = requestPath("/shares/share2/schemas/default/tables/table2/version?startingTimestamp=abc"),
```
chakankardb, line 445: Question: why is version needed?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -412,14 +424,15 @@ private[spark] class DeltaSharingRestClient(
       }
     }
 
-  // Append SparkStructuredStreaming in the USER_AGENT header, in order for the delta sharing server
+  // Add SparkStructuredStreaming in the USER_AGENT header, in order for the delta sharing server
   // to recognize the request for streaming, and take corresponding actions.
   private def getUserAgent(): String = {
-    DeltaSharingRestClient.USER_AGENT + (if (forStreaming) {
-      s" ${DeltaSharingRestClient.SPARK_STRUCTURED_STREAMING}/$STREAMING_VERSION"
+    val sparkAgent = if (forStreaming) {
```
chakankardb, line 430: Why is this still needed? Is it to track the streaming requests on the provider side?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
linzhou-db: I don't know. 
I'm not fully happy with returnMetadata either. But returnMetadataChanges sounds like the changes of metadata, which we are returning is just the metadata it self. They are the additional metadata seen in the deltaLog. 
returnAdditionalMetadata? not as good as returnMetadata even..

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
linzhou-db: This is to leave the endpoint without `/version` for getTable.  And this is one of the missing tests not updated in that PR.
Now getTable is not needed, I think it's a good idea to have `/version` suffix for getTableVersion.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: yes. 1. tracking in the usage log 2. gate traffic during private preview.

chakankardb (approved):

PR #215, opened 2022-11-18
Revert "Add get table api (#210)"

andyl-db:
Reverts get table api as it is unnecessary. Keeps the fix for integration tests and get table version changes.


linzhou-db (approved):
Thanks!

PR #223, opened 2022-12-09
Extends DeltaSharingProfileProvider to customize tablePath and refresher 

linzhou-db:
Extends DeltaSharingProfileProvider to customize tablePath and refresher

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
@@ -38,8 +38,11 @@ import io.delta.sharing.spark.model.{
 
 private[sharing] case class RemoteDeltaFileIndexParams(
     val spark: SparkSession,
-    val snapshotAtAnalysis: RemoteSnapshot) {
-  def path: Path = snapshotAtAnalysis.getTablePath
+    val snapshotAtAnalysis: RemoteSnapshot,
+    val profileProvider: DeltaSharingProfileProvider) {
+  def path: Path = new Path(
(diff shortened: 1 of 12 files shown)


spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -179,21 +181,8 @@ private[sharing] case class RemoteDeltaBatchFileIndex(
...
     partitionFilters: Seq[Expression],
     dataFilters: Seq[Expression]): Seq[PartitionDirectory] = {
-    // TODO(lin.zhou): Actually refresh the presigned url in the cache instead of just using
-    // getIdToUrlMap
-    CachedTableManager.INSTANCE
-      .register(params.path.toString, getIdToUrlMap, Seq(new WeakReference(this)), () => {
-        getIdToUrlMap
-      })
-
     // We ignore partition filters for list files, since the delta sharing server already
```
zsxwing, line 184: We cannot ignore partition filters here. For streaming cases, this is okay as Spark doesn't support partition filter push down for streaming sources yet. But CDF batch queries, we need to apply the partition filters.

zsxwing (approved):
Overall this looks good to me. I left a comment unrelated to this PR

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
linzhou-db: But `RemoteDeltaBatchFileIndex` is used for the delta sharing streaming source? 
Do you mean I need to apply the partition filters in `RemoteDeltaCDFFileIndexBase.listFiles`?

PR #228, opened 2022-12-14
Spark connector changes to consume size from metadata.

chakankardb:
In this change, delta sharing will try to get the table size from table metadata if it is available.
If not available, it will fall back to computing the size from full parquet scan.
The existing code always computes the size from parquet scan, making it a performance bottleneck.

Tested the following combinations:

New server, new connector -> avoids extra queryTable call
New server, old connector -> works (backwards compatibility)
Old server, new connector -> works (backwards compatibility)

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/model.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/model.scala
@@ -96,7 +96,9 @@ private[sharing] case class Metadata(
     schemaString: String = null,
     configuration: Map[String, String] = Map.empty,
     partitionColumns: Seq[String] = Nil,
-    version: java.lang.Long = null) extends Action {
+    version: java.lang.Long = null,
+    size: java.lang.Long = null,
+    numFiles: java.lang.Long = null) extends Action {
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/io/delta/sharing/spark/model.scala
```diff
@@ -96,7 +96,9 @@ private[sharing] case class Metadata(
     schemaString: String = null,
     configuration: Map[String, String] = Map.empty,
     partitionColumns: Seq[String] = Nil,
-    version: java.lang.Long = null) extends Action {
+    version: java.lang.Long = null,
+    size: java.lang.Long = null,
```
zsxwing, line 100: could you also add these two new fields to PROTOCOL.md?

chakankardb: Good catch, done.

zsxwing (approved):

zhuansunxt (approved):
🚢

linzhou-db (approved):

PR #229, opened 2022-12-16
Support filtering on additional cdf columns

linzhou-db:
Support filtering on additional cdf columns: _commit_version/_commit_timestamp/_change_type

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
@@ -130,6 +130,7 @@ private[sharing] abstract class RemoteDeltaCDFFileIndexBase(
     auxPartitionSchema: Map[String, DataType] = Map.empty)
     extends RemoteDeltaFileIndexBase(params) {
 
+  // TODO: linzhou, update this
   override def partitionSchema: StructType = {
     DeltaTableUtils.updateSchema(params.snapshotAtAnalysis.partitionSchema, auxPartitionSchema)
   }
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -130,6 +130,7 @@ private[sharing] abstract class RemoteDeltaCDFFileIndexBase(
     auxPartitionSchema: Map[String, DataType] = Map.empty)
     extends RemoteDeltaFileIndexBase(params) {
 
+  // TODO: linzhou, update this
```
chakankardb, line 133: Consider adding more details on what needs to be updated.

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -151,10 +152,13 @@ private[sharing] case class RemoteDeltaCDFAddFileIndex(
   override def listFiles(
     partitionFilters: Seq[Expression],
     dataFilters: Seq[Expression]): Seq[PartitionDirectory] = {
+    val updatedFiles = addFiles.map { a =>
+      AddFileForCDF(a.url, a.id, a.getPartitionValuesInDF, a.size, a.version, a.timestamp, a.stats)
```
chakankardb, line 156: Question: makePartitionDirectories will call getPartitionValuesInDF again? Would that lead to double inclusion of the cdf columns?
Consider adding some comments.

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
linzhou-db: This need to be removed.

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
linzhou-db: It's not double inclusion, and actually to make them consistent. Here, it's applied to each file, makePartitionDirectories is applying the partitions to the directory path.

chakankardb (approved):

PR #232, opened 2022-12-17
Support more flexible format for timestamp options in spark in v0.5"

linzhou-db:
Client side:
Support more flexible format for startingTimestamp/endingTimestamp options in spark, to be similar with delta.

Server side:
Support ISO_OFFSET_DATE_TIME format for all timestamp related parameters: cdf.startingTimestamp, cdf.endingTimestamp.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
@@ -24,10 +24,13 @@ import scala.collection.mutable
 import org.apache.spark.SparkEnv
 import org.apache.spark.delta.sharing.PreSignedUrlCache
 import org.apache.spark.sql.{SparkSession, SQLContext}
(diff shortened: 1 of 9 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala
```diff
@@ -76,6 +79,16 @@ private[sharing] class DeltaSharingDataSource extends RelationProvider with Data
     deltaLog.createRelation(versionAsOf, cdfOptions = cdfOptions.toMap)
   }
 
+  private def getFormattedTimestamp(str: String): String = {
```
chakankardb, line 82: consider adding comments.

chakankardb (approved):

linzhou-db:
This is not a bug fix but rather an UX improvement, so we decided to withdraw this change in v.0.5.

Issue #238, opened 2022-12-21
Support for load_as_pyarrow_dataset or load_as_pyarrow_table


chitralverma:
This is a new feature request or rather a little refactoring in the code for reader to allow users to read datasets directly as `pyarrow` datasets and tables.

As you can see https://github.com/delta-io/delta-sharing/blob/main/python/delta_sharing/reader.py#L163-L164, we are anyways creating the pyarrow dataset and table, which is then used to convert to a pandas DF.

I would like to refactor this part and expose this as separate functions - `to_pyarrow_dataset` and `to_pyarrow_table`.

Advantage of this refactoring is that users will then be able to efficiently get the pyarrow things directly with out an additional full conversion to pandas dataset if required. This will allow the extension of delta-sharing on other processing systems like Datafusion, Polars etc, since they all extensively rely on pyarrow datasets.

Please let me know if this issue makes sense to you, I can raise a PR quick for this in a day or so.

PR #240, opened 2022-12-23
Add load_as methods for pyarrow dataset and table

chitralverma:
Adds separate implementations for `load_as_pyarrow_table` and `load_as_pyarrow_dataset` that allows users to read delta sharing tables as pyarrow table and dataset respectively.

- [x] Add basic implementation
- [x] Fix lint
- [x] Refactor common code
- [x] Verify performance with and without limit
- [x] Add tests - converter
- [x] Add tests - reader
- [ ] Add tests - delta_sharing
- [x] Add examples
- [ ] Fix review comments

closes https://github.com/delta-io/delta-sharing/issues/238

### DIFF
--- a/python/delta_sharing/converter.py
+++ b/python/delta_sharing/converter.py
@@ -15,11 +15,14 @@
 #
 from decimal import Decimal
 from typing import Any, Callable, Dict
-from pyarrow.dataset import Dataset as PyArrowDataset
-from pyarrow.dataset import dataset
-from pyarrow import Table as PyArrowTable
+
 import numpy as np
 import pandas as pd
+import pyarrow as pa
+from pyarrow import Schema as PyArrowSchema
+from pyarrow import Table as PyArrowTable
+from pyarrow.dataset import Dataset as PyArrowDataset
+from pyarrow.dataset import dataset
 
 
 def _get_dummy_column(schema_type):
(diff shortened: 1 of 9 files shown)


chitralverma:
@goodwillpunning @linzhou-db From the build logs I can see that the `PYARROW_VERSION` has been pinned to 4.x somewhere in the environment variables. This version of pyarrow came out in May, 2021 and since then there have been 6 major version releases.

Seems like there are some API inconsistencies the pinned version 4.x which is causing build failure on GitHub but locally test cases are passing. I also verified with versions 5.x to 10.x and was not able to reproduce the issue. Can you please unpin or upgrade this `PYARROW_VERSION`.

linzhou-db:
Thanks @chitralverma , will take a look once back in Jan. 
cc @zsxwing

python/delta_sharing/converter.py
```diff
@@ -57,7 +62,157 @@ def _get_dummy_column(schema_type):
...
+                raise TypeError(
+                    f"Could not parse map field types: array<timestamp> to PyArrow type."
+                )
+            elif isinstance(element_type, dict) and element_type["type"] == "struct":
+                if any(
+                    isinstance(struct_field["type"], dict)
+                    and struct_field["type"]["type"] == "struct"
+                    for struct_field in element_type["fields"]
+                ):
+                    raise TypeError("Nested StructType cannot be converted to PyArrow type.")
```
linzhou-db, line 147: "Double Nested cannot ..."?

python/delta_sharing/converter.py
```diff
@@ -57,7 +62,157 @@ def _get_dummy_column(schema_type):
...
+                    f"Could not parse map field with struct key or value types to PyArrow type."
+                )
+            return pa.map_(_to_arrow_type(key_type), _to_arrow_type(value_type))
+
+        elif is_nested and f_type["type"] == "struct":
+            if any(
+                isinstance(struct_field["type"], dict) and struct_field["type"]["type"] == "struct"
+                for struct_field in f_type["fields"]
+            ):
+                raise TypeError("Nested StructType cannot be converted to PyArrow type.")
```
linzhou-db, line 172: double nested?

python/delta_sharing/tests/test_converter.py
```diff
@@ -81,8 +82,269 @@ def test_get_empty_table():
...
+        ]
+        _to_pyarrow_schema(base_schema_dict_tmp)
+
+    assert str(exp_info3.value) == "Could not parse type: xyz to PyArrow type."
+
+
+def test_pyarrow_schema_base():
+    base_schema_dict = {
+        "type": "struct",
+        "fields": [
```
linzhou-db, line 318: cover all types in this test?

python/delta_sharing/reader.py
```diff
@@ -71,19 +79,112 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
             timestamp=self._timestamp
         )
 
-    def to_pandas(self) -> pd.DataFrame:
+    def _get_response(self) -> ListFilesInTableResponse:
         response = self._rest_client.list_files_in_table(
```
linzhou-db, line 83: you can directly `return self._rest_client.list...`?

python/delta_sharing/reader.py
```diff
@@ -71,19 +79,112 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
...
+        else:
+            left = self._limit
+            pa_tables = []
+
+            for file_url in file_urls:
+                ds = DeltaSharingReader._to_pyarrow_dataset(file_url, pyarrow_ds_options)
+                tbl: PyArrowTable = ds.head(left, **pyarrow_tbl_options)
+                pa_tables.append(tbl)
+                left -= tbl.num_rows
+                assert (
```
linzhou-db, line 160: is this a hard limit? and does it require exact `limit` number of rows returned?

python/delta_sharing/reader.py
```diff
@@ -71,19 +79,112 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
...
+        # the other URLs have same scheme as well.
+        # Otherwise, fsspec filesystem cannot be instantiated.
+        if file_urls and isinstance(file_urls, list):
+            same_scheme = True
+            for url in file_urls:
+                if not urlparse(url).scheme == first_scheme:
+                    same_scheme = False
+                    break
+
+            assert same_scheme, "All files did not follow the same URL scheme."
```
linzhou-db, line 125: nit: "All files should follow the same URL scheme" ?

python/delta_sharing/tests/test_reader.py
```diff
@@ -448,3 +469,186 @@ def list_table_changes(
...
+
+    reader = DeltaSharingReader(Table("table_name", "share_name", "schema_name"), RestClientMock())
+    ds = reader.to_pyarrow_dataset()
+
+    assert expected_schema.equals(ds.schema, check_metadata=True)
+    assert len([f.path for f in ds.get_fragments()]) == 0
+    assert ds.count_rows() == 0
+
+
+def test_to_pyarrow_table_non_partitioned(tmp_path):
```
linzhou-db, line 578: What's the difference between this test and `test_to_pyarrow_dataset`?

python/requirements-dev.txt
```diff
@@ -1,7 +1,7 @@
 # Dependencies. When you update don't forget to update setup.py.
 pandas
-pyarrow>=4.0.0
```
linzhou-db, line 3: question: why are these removed?

python/delta_sharing/reader.py
linzhou-db: Question: what's an example of this failure? And is it possible to add a test case to cover it?

linzhou-db:
Also what's your thought on loading cdf in pyarrow? is it something not needed for now?

python/delta_sharing/reader.py
chitralverma: yes it results exactly the number of rows asked for. but does it file by file instead as I saw that in practice this is faster than just calling `.head()` on the pyarrow table.

python/delta_sharing/reader.py
chitralverma: I dont think the delta server can return files from different places for for the same table but I added this case just in case if in the future delta sharing turns in to a cross cloud service (some data in s3 and some data in GCS).

Another reason for adding this was that we dont have to initialize FSSPEC FS for each path if they all follow the same scheme.

python/delta_sharing/tests/test_reader.py
chitralverma: thats for pyarrow dataset (lazy, faster) and this is for pyarrow table (eager)

python/requirements-dev.txt
chitralverma: remove temporarily to see if this is causing the build to fail. I will add it again before the PR is completely ready.

please see my original comment regarding this.

chitralverma:
> Also what's your thought on loading cdf in pyarrow? is it something not needed for now?

I would prefer to raise a separate PR for the CDF to keep things simple and concise, this is just for the data.

python/delta_sharing/reader.py
linzhou-db: right. I wonder we don't have to fail if we got a few more rows?

python/delta_sharing/reader.py
linzhou-db: I don't foresee delta sharing support a single table from multiple clouds any time soon. 
Plus no test coverage on the code, could we rather turn this into a TODO.

ion-elgreco:
@chitralverma @linzhou-db can we revive this PR?

PR #244, opened 2023-01-05
Spark connector changes to consume size from metadata in v0.5

linzhou-db:
In this change, delta sharing will try to get the table size from table metadata if it is available.
If not available, it will fall back to computing the size from full parquet scan.
The existing code always computes the size from parquet scan, making it a performance bottleneck.

Tested the following combinations:

New server, new connector -> avoids extra queryTable call


chakankardb (approved):
Thanks for taking care of this!

PR #253, opened 2023-02-02
Basic case classes for json filtering ops, and their evaluation.

chakankardb:
This is the first PR for Json based partition predicate filtering.
It implements basic operations and their evaluation.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
@@ -79,6 +79,11 @@ trait BaseOp {
   // If the method throws an exception, the caller should skip filtering for this context.
   @throws[IllegalArgumentException]
   def eval(ctx: EvalContext): Any
+
+  // Evaluation of a non-leaf operation is expected to return a boolean.
+  def evalExpectBoolean(ctx: EvalContext): Boolean = {
+    eval(ctx).asInstanceOf[Boolean]
+  }
 }
 
 // Represents a leaf operation.
@@ -95,12 +100,7 @@ trait LeafOp extends BaseOp {
 }
 
 // Represents a non-leaf operation.
-trait NonLeafOp extends BaseOp {
-  // Evaluation of a non-leaf operation is expected to return a boolean.
-  def evalExpectBoolean(ctx: EvalContext): Boolean = {
-    eval(ctx).asInstanceOf[Boolean]
-  }
-}
+trait NonLeafOp extends BaseOp;
 
 // Represents a unary operation.
 trait UnaryOp {
@@ -258,7 +258,7 @@ case class GreaterThanOrEqualOp(children: Seq[LeafOp]) extends NonLeafOp with Bi
  *   "&&", "||", "!"
  */
 
-case class AndOp(val children: Seq[NonLeafOp]) extends NonLeafOp with NaryOp {
+case class AndOp(val children: Seq[BaseOp]) extends NonLeafOp with NaryOp {
   override def validate(): Unit = validateChildren(children)
 
   override def eval(ctx: EvalContext): Any = {
@@ -266,7 +266,7 @@ case class AndOp(val children: Seq[NonLeafOp]) extends NonLeafOp with NaryOp {
   }
 }
 
-case class OrOp(val children: Seq[NonLeafOp]) extends NonLeafOp with NaryOp {
+case class OrOp(val children: Seq[BaseOp]) extends NonLeafOp with NaryOp {
   override def validate(): Unit = validateChildren(children)
 
   override def eval(ctx: EvalContext): Any = {
@@ -274,7 +274,7 @@ case class OrOp(val children: Seq[NonLeafOp]) extends NonLeafOp with NaryOp {
   }
 }
 
-case class NotOp(val children: Seq[NonLeafOp]) extends NonLeafOp with UnaryOp {
+case class NotOp(val children: Seq[BaseOp]) extends NonLeafOp with UnaryOp {
   override def validate(): Unit = validateChildren(children)
 
   override def eval(ctx: EvalContext): Any = !children(0).evalExpectBoolean(ctx)
--- a/spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
@@ -164,6 +164,16 @@ case class ColumnOp(val name: String, val valueType: String) extends LeafOp {
   // Returns the column value using the specified context, and its type.
   override def eval(ctx: EvalContext): Any = (resolve(ctx), valueType)
 
+  // We support column evaluation to return boolean if it is of that type.
+  override def evalExpectBoolean(ctx: EvalContext): Boolean = {
+    if (valueType != OpDataTypes.BoolType) {
+      throw new IllegalArgumentException(
+        "Unsupported type for boolean evaluation: " + valueType
+      )
+    }
+    resolve(ctx).toBoolean
+  }
+
   // Determine if the column value is null.
   override def isNull(ctx: EvalContext): Boolean = (resolve(ctx) == null)
 
@@ -295,12 +305,15 @@ object EvalHelper {
          children(0) + " and " + children(1)
       )
     }
-    // Handle nulls.
-    if (leftVal == null && rightVal == null) {
-      return true
-    } else if (leftVal == null || rightVal == null) {
+
+    // We throw an exception for nulls, which will skip filtering.
+    if (leftVal == null || rightVal == null) {
+      throw new IllegalArgumentException(
+        "Comparison with null is not supported: " + children(0) + " and " + children(1)
+      )
       return false
     }
+
     // Perform data type conversion and match.
     // TODO(abhijit): For literal operations, we can optimize evaluation by caching
     //                this conversion.
@@ -329,7 +342,7 @@ object EvalHelper {
       )
     }
 
-    // We throw and exception for nulls, which will skip filtering.
+    // We throw an exception for nulls, which will skip filtering.
     if (leftVal == null || rightVal == null) {
       throw new IllegalArgumentException(
         "Comparison with null is not supported: " + children(0) + " and " + children(1)
--- a/spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
@@ -111,6 +111,7 @@ trait UnaryOp {
           this + " : expected 1 but found " + children.size + " children"
       )
     }
+    children(0).validate()
   }
 }
 
@@ -123,6 +124,7 @@ trait BinaryOp {
           this + " : expected 2 but found " + children.size + " children"
       )
     }
+    children.map(c => c.validate())
   }
 }
 
@@ -135,6 +137,7 @@ trait NaryOp {
           this + " : expected at least 2 but found " + children.size + " children"
       )
     }
+    children.map(c => c.validate())
   }
 }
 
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,346 @@
...
+      case OpDataTypes.BoolType => leftVal.toBoolean == rightVal.toBoolean
+      case OpDataTypes.IntType => leftVal.toInt == rightVal.toInt
+      case OpDataTypes.LongType => leftVal.toLong == rightVal.toLong
+      case OpDataTypes.StringType => leftVal == rightVal
+      case OpDataTypes.DateType =>
+        java.sql.Date.valueOf(leftVal).equals(java.sql.Date.valueOf(rightVal))
+      case _ =>
+        throw new IllegalArgumentException("Unsupported type: " + leftType)
+    }
+  }
```
chakankardb, line 314: I don't like this type handling.
Any thoughts on simplifying this structure are welcome?

spark/src/test/scala/io/delta/sharing/spark/filters/JsonPredicateSuite.scala
```diff
@@ -0,0 +1,291 @@
...
+    test_op(op_from_json)
+
+    // Test validation failures.
+    intercept[IllegalArgumentException] {
+      LiteralOp(value = "2021-04-29", valueType = "junk").validate()
+    }
+    intercept[IllegalArgumentException] {
+      LiteralOp(value = null, valueType = "int").validate()
+    }
+  }
```
chakankardb, line 49: I'll write a lot more unit tests as the review makes progress. Just wanted to get this one out.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,346 @@
...
+ * The data types supported by the filtering operations.
+ */
+object OpDataTypes {
+  val BoolType = "bool"
+  val IntType = "int"
+  val LongType = "long"
+  val StringType = "string"
+  val DateType = "date"
+
+  val supportedTypes = Set(BoolType, IntType, LongType, StringType, DateType)
```
linzhou-db, line 40: How is this list decided?

spark/src/test/scala/io/delta/sharing/spark/filters/JsonPredicateSuite.scala
```diff
@@ -0,0 +1,291 @@
...
+    }
+    val op = LiteralOp(value = "2021-04-29", valueType = "date")
+    test_op(op)
+
+    // Check that we can convert to json and back.
+    val op_from_json = JsonUtils.fromJson[BaseOp](JsonUtils.toJson[BaseOp](op))
+    test_op(op_from_json)
+
+    // Test validation failures.
+    intercept[IllegalArgumentException] {
```
linzhou-db, line 43: verify error message?

linzhou-db (commented):
During which step are these predicates used?

My initial understanding is that they are constructed on the client side and converted to json, and are evaluated on the server side? I must be missing something.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,346 @@
...
+  override def validate(): Unit = validateChildren(children)
+
+  override def eval(ctx: EvalContext): Any = !children(0).evalExpectBoolean(ctx)
+}
+
+// A helper for evaluating opertions.
+object EvalHelper {
+
+  // Implements "equal" between two leaf operations.
+  def equal(children: Seq[LeafOp], ctx: EvalContext): Boolean = {
```
zhuansunxt, line 284: This only accepts leaf node, it means we don't support expressions like `(a EQ b) EQ True`?

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,346 @@
...
+ * The following ops represent boolean logic related operations.
+ * The (logically) supported operations are:
+ *   "&&", "||", "!"
+ */
+
+case class AndOp(val children: Seq[NonLeafOp]) extends NonLeafOp with NaryOp {
+  override def validate(): Unit = validateChildren(children)
+
+  override def eval(ctx: EvalContext): Any = {
+    children.forall(c => c.evalExpectBoolean(ctx))
```
zhuansunxt, line 262: This assumes the children are all non-leaf node. What about expression like `True AND False`? I guess it's not supported.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,346 @@
...
+    // We simply throw an exception, which will cause filtering to be skipped.
+    if (leftType != rightType) {
+      throw new IllegalArgumentException(
+        "Type mismatch: " + leftType + " vs " + rightType + " for " +
+         children(0) + " and " + children(1)
+      )
+    }
+
+    // Handle nulls.
+    if (leftVal == null || rightVal == null) {
```
zhuansunxt, line 330: Should we throw exception here instead? Is `1 < null` supposed to be evaluated to `false`? Maybe null should be forbidden in comparison operator because the behavior looks more like `undefined`.

chakankardb:
> 

@linzhou-db Yes, these classes will be created by spark client, sent over the wire as json in query table, and  will get recreated and evaluated by delta sharing server.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
chakankardb: We are starting out with a minimal list (kubit mostly uses date, and I also added basic types).
We will expand this set as needed.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
chakankardb: Your example can be expressed as: (a eq b) && True.

We only need equality for leaf nodes (columns and literals) for both partition and stats filtering as of now.
I'm intentionally starting out with a restricted version, and we can relax this as needed.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
chakankardb: This implementation is intentionally restricted towards file filtering based on partition values and stats.
 So we don't expect to compare two leaf level literals. 
The "True" node will typically be an output of either a comparator node, or a logical operation.

These constraints are very easy to relax if we see a use case.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
chakankardb: Yes, good catch.
Nulls can happen if the partition column is nullable, and some rows have null (not common I guess).
Let us return the corresponding parquet files in such cases and let spark deal with it.

spark/src/test/scala/io/delta/sharing/spark/filters/JsonPredicateSuite.scala
chakankardb: good catch, done.

linzhou-db:
> > 
> 
> @linzhou-db Yes, these classes will be created by spark client, sent over the wire as json in query table, and will get recreated and evaluated by delta sharing server.

Then is `eval` needed on the client side?

chakankardb:
> 

@linzhou-db Eval is strictly not needed on the client side. It still needs to be part of op hierarchy due to the recursive nature. This is similar to how SQL expressions are evaluated.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,351 @@
...
+  //
+  // The method throws an exception if any of the validations fail.
+  @throws[IllegalArgumentException]
+  def validate(): Unit
+
+  // Evaluates this operation in the given context.
+  //
+  // If the method throws an exception, the caller should skip filtering for this context.
+  @throws[IllegalArgumentException]
+  def eval(ctx: EvalContext): Any
```
wchau, line 81: I am uncomfortable that `Any` can be drastically different types:
- (String, String)
- Boolean

Can the predicate evaluation return something better than just `Boolean`?

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
wchau: If you're enforcing the same type already, since `leftVal` and `rightVal` are both string, you can just compare them directly. It would have been nicer to pass the actual types from the bottom though instead of converting from String to their native types.

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
chakankardb: Made the children baseOps

spark/src/main/scala/io/delta/sharing/spark/filters/JsonPredicates.scala
```diff
@@ -0,0 +1,351 @@
...
+    // If the types don't match, it implies a malformed predicate tree.
+    // We simply throw an exception, which will cause filtering to be skipped.
+    if (leftType != rightType) {
+      throw new IllegalArgumentException(
+        "Type mismatch: " + leftType + " vs " + rightType + " for " +
+         children(0) + " and " + children(1)
+      )
+    }
+    // Handle nulls.
+    if (leftVal == null && rightVal == null) {
```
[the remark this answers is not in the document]
wchau: In the SQL standard, NULL = NULL returns false.

https://stackoverflow.com/questions/1843451/why-does-null-null-evaluate-to-false-in-sql-server

Unless you support IS (NOT) NULL, this is the expected behavior.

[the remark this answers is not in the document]
chakankardb: https://src.dev.databricks.com/databricks/runtime@master/-/blob/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/predicates.scala?L1267:12&subtree=true&popover=pinned#tab=references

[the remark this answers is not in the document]
chakankardb: Also, we are not SQL, but json:

If someone create the following tree:
  EqualOp(ColumnOp(name="hireDate", valueType="date"), LiteralOp(value=null, valueType="date"))

Intuitively, I will expect the evaluation to return true if the column hireDate is nullable, and it has a null.

So let's start with a safe null check, and evolve the code as needed.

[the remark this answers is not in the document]
wchau: That is the null-safe equal operator (<=>), which is not the same as the equal operator (=).

https://docs.databricks.com/sql/language-manual/sql-ref-null-semantics.html#comparison-operators

I would want this to be accurate here, even if we are supporting a limited subset of operations.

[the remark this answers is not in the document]
wchau: More accurately, it should evaluate to NULL, and not true / false, which your current code does not support.

To simplify things without accidentally implementing the wrong behavior, maybe we should just raise an exception if either operand is NULL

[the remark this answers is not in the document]
chakankardb: Mode changes as discussed offline, PTAL.

chakankardb:
jenkins merge

PR #257, opened 2023-02-10
Document json predicates.

chakankardb:
This PR documents the json predicates.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -2852,7 +2852,7 @@ ValueType | Description
 "int" | Represents an Integer type.
 "long" | Represents a Long type.
 "string" | Represents a String type.
-"date" | Represents a Date type.
+"date" | Represents a Date type in "yyyy-mm-dd" format.
 
 
 Examples
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1700,7 +1700,7 @@ This is the API for clients to read data from a table.
   ],
   "limitHint": 0,
   "version": 0,
-  "jsonPredicates": "base64_encoded_string"
+  "jsonPredicateHints": "base64_encoded_string"
 }
 
 ```
@@ -1918,9 +1918,9 @@ The request body should be a JSON string containing the following optional field
     - If the server fails to parse one of the SQL predicates, or fails to evaluate it, the server may skip it.
     - Predicate expressions are conjunctive (AND-ed together).
   - When it’s absent, the server will return all of files in the table.
-  - **These will be deprecated once all the client and server implementation move to using jsonPredicates below**.
+  - **These will be deprecated once all the client and server implementation move to using jsonPredicateHints below**.
 
-- **jsonPredicates** (type: String, optional): query predicates on partition columns specified using [a structured JSON format](#json-predicates-for-filtering), which is encoded using Base64 encoding.
+- **jsonPredicateHints** (type: String, optional): query predicates on partition columns specified using [a structured JSON format](#json-predicates-for-filtering), which is encoded using Base64 encoding.
   - When it’s present, the server will try to use the predicates to filter table's files, which could boost query performance.
     - As with **predicateHints**, this filtering is **BEST EFFORT**. The server may return files that don’t satisfy the predicates.
     - If the server encounters any errors during predicate processing (for example, invalid syntax or non existing columns), it will skip filtering and return all the files. 
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1700,7 +1700,7 @@ This is the API for clients to read data from a table.
   ],
   "limitHint": 0,
   "version": 0,
-  "jsonPredicateHints": "base64_encoded_string"
+  "jsonPredicateHints": "string"
 }
 
 ```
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -1908,14 +1910,21 @@ The response contains multiple lines:
...
  
 - **predicateHints** (type: String, optional): a list of SQL boolean expressions using [a restricted subset of SQL](#sql-expressions-for-filtering), in a JSON array.
   - When it’s present, the server will use the provided predicates as a hint to apply the SQL predicates on the returned files.
     - Filtering files based on the SQL predicates is **BEST EFFORT**. The server may return files that don’t satisfy the predicates.
     - If the server fails to parse one of the SQL predicates, or fails to evaluate it, the server may skip it.
     - Predicate expressions are conjunctive (AND-ed together).
   - When it’s absent, the server will return all of files in the table.
+  - **These will be deprecated once all the client and server implementation move to using jsonPredicateHints below**.
+
+- **jsonPredicateHints** (type: String, optional): query predicates on partition columns specified using [a structured JSON format](#json-predicates-for-filtering), which is encoded using Base64 encoding.
```
zsxwing, line 1923: Can we avoid using Base64? We can make `jsonPredicateHints` be a string, and just convert the content to a json string, then set it. See the `stats` field in `file`. In addition, can we make it be a json array similar to `predicateHints`?

Another way to embed the json content is:
```
{
  "jsonPredicateHints": [
    {"op":"isNull","children":[{"op":"column","name":"hireDate","valueType":"date"}]},
    {"op":"isNull","children":[{"op":"column","name":"state","valueType":"string"}]}
  ],
  "limitHint": 1000,
  "version": 123
}
```

chakankardb: Base64 encoding is a common practice for not trivial data types.
We url-encode a lot of our params as well.

I had trouble packing/unpacking json predicate tree through various hoops during my testing (for example, http query body did not like it).
Once I switch to base64 encoding, things worked like charm.

We will need to change our implementation in spark connector and UC's delta sharing server.
Let me know if you all feel strongly against base64 encoding.

zsxwing: > (for example, http query body did not like it).

This is likely you are not using the right API? Which approach were you using? `stats` approach or embedded json?

Base64 encoding is not readable. Would be great to avoid that.

PROTOCOL.md
```diff
@@ -1908,14 +1910,21 @@ The response contains multiple lines:
...
-The request body should be a JSON string containing the following two optional fields:
+The request body should be a JSON string containing the following optional fields:
  
 - **predicateHints** (type: String, optional): a list of SQL boolean expressions using [a restricted subset of SQL](#sql-expressions-for-filtering), in a JSON array.
   - When it’s present, the server will use the provided predicates as a hint to apply the SQL predicates on the returned files.
     - Filtering files based on the SQL predicates is **BEST EFFORT**. The server may return files that don’t satisfy the predicates.
     - If the server fails to parse one of the SQL predicates, or fails to evaluate it, the server may skip it.
     - Predicate expressions are conjunctive (AND-ed together).
   - When it’s absent, the server will return all of files in the table.
+  - **These will be deprecated once all the client and server implementation move to using jsonPredicateHints below**.
```
linzhou-db, line 1921: nit: `This will be ...` ?

PROTOCOL.md
```diff
@@ -1908,14 +1910,21 @@ The response contains multiple lines:
...
+
+- **jsonPredicateHints** (type: String, optional): query predicates on partition columns specified using [a structured JSON format](#json-predicates-for-filtering), which is encoded using Base64 encoding.
+  - When it’s present, the server will try to use the predicates to filter table's files, which could boost query performance.
+    - As with **predicateHints**, this filtering is **BEST EFFORT**. The server may return files that don’t satisfy the predicates.
+    - If the server encounters any errors during predicate processing (for example, invalid syntax or non existing columns), it will skip filtering and return all the files.
```
linzhou-db, line 1926: Do we need to adjust the content here, though skip filtering, the server may still respect limitHints, which doesn't return all files?

PROTOCOL.md
```diff
@@ -2801,6 +2810,107 @@ Expression | Example
...
+-|-
+"bool" | Represents an Boolean type.
+"int" | Represents an Integer type.
+"long" | Represents a Long type.
+"string" | Represents a String type.
+"date" | Represents a Date type.
+
+
+Examples
+  - Note that the json should be encoded using Base64 encoding.
```
linzhou-db, line 2859: Could you point me to the code where this encoding is applied?

chakankardb: https://src.dev.databricks.com/delta-io/delta-sharing/-/blob/spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala?L119:9&subtree=true#tab=references

PROTOCOL.md
chakankardb: Done.

PROTOCOL.md
chakankardb: ok, removed the base64 encoding as per offline discussion.

PROTOCOL.md
chakankardb: I think limitHint today is suppressed if there is predicate filtering.

linzhou-db (approved):

PROTOCOL.md
```diff
@@ -2801,6 +2810,106 @@ Expression | Example
...
+
+The supported value types:
+
+ValueType | Description
+-|-
+"bool" | Represents an Boolean type.
+"int" | Represents an Integer type.
+"long" | Represents a Long type.
+"string" | Represents a String type.
+"date" | Represents a Date type.
```
zsxwing, line 2855: Should we document what format we support? Currently, we only support formats like `2021-04-29`. Right?

zsxwing (approved):

PROTOCOL.md
chakankardb: Updated.

PR #262, opened 2023-02-13
Add UUIDs as Table IDs on the reference server

zhuansunxt:
The protocol currently says the table ID is optional however in lots of client implementations it requires table IDs to be in place. This change returns table IDs on all the tables hosted by the reference server.

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/SharedTableManager.scala
+++ b/server/src/main/scala/io/delta/sharing/server/SharedTableManager.scala
@@ -138,8 +138,14 @@ class SharedTableManager(serverConfig: ServerConfig) {
     val schemaConfig = getSchema(getShareInternal(share), schema)
     getPage(nextPageToken, Some(share), Some(schema), maxResults, schemaConfig.getTables.size) {
       (start, end) =>
-        schemaConfig.getTables.asScala.map { tableConfig =>
-          Table().withName(tableConfig.getName).withSchema(schema).withShare(share)
+        schemaConfig.getTables.asScala.map {
+          tableConfig =>
+            Table(
+              name = Some(tableConfig.getName),
+              schema = Some(schema),
+              share = Some(share),
+              id = Some(tableConfig.getId)
+            )
         }.slice(start, end)
     }
   }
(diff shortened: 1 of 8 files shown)


linzhou-db (approved):
So is there a required format on the id?

zhuansunxt:
@linzhou-db The protocol does not demand a strict format as of now.

zhuansunxt:
There are some test flakiness in the `DeltaSharingServerSuite` and the failures are not directly related to this change.

wchau:
The Databricks implementation requires UUID. For the sake of consistency, maybe we should require UUID

server/src/main/scala/io/delta/sharing/server/SharedTableManager.scala
```diff
@@ -138,8 +138,14 @@ class SharedTableManager(serverConfig: ServerConfig) {
...
       (start, end) =>
-        schemaConfig.getTables.asScala.map { tableConfig =>
-          Table().withName(tableConfig.getName).withSchema(schema).withShare(share)
+        schemaConfig.getTables.asScala.map {
+          tableConfig =>
+            Table(
+              name = Some(tableConfig.getName),
+              schema = Some(schema),
+              share = Some(share),
+              id = Some(tableConfig.getId)
```
wchau, line 147: Should this be optional?

zhuansunxt: What do you mean?

zhuansunxt: The ID in tableConfig has a default value so it means if a table in the config is not configured with id, it'll have an empty value. When a tableConfig is translated here to a proto, a table without ID will have `Some("")` and will produce an empty string in the API response. This behavior will not break any existing usage right?

wchau: It may not break current usage, unless a client wants to track shares as they are renamed. The field is optional, but it doesn't mean that it should be defaulted to "". The share ID is no longer unique across shares with this.

zhuansunxt: Makes sense. I'll send out a fix. Btw this PR does not add Share ID but only Table ID. I plan to remove the requirement on Share ID on Databricks client.

PR #263, opened 2023-02-13
Update PROTOCOL.md documentation to recommend UUID as the format for table and share id

zhuansunxt:


linzhou-db (approved):

wchau (approved):

zhuansunxt:
@rustyconover  the letter in the UUID is essentially a hexadecimal value so it should be case insensitive.

PR #268, opened 2023-02-15
delta sharing server changes for json predicates.

chakankardb:
This PR implements the json predicate based filtering in the delta sharing server.
The changes are disabled by default.

### DIFF
--- a/server/src/main/scala/io/delta/standalone/internal/JsonPredicates.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/JsonPredicates.scala
@@ -1,0 +1,417 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone.internal
+
+// NOTE: This file should be kept in sync with the corresponding file
+// in the spark connector.
+
+import scala.collection.mutable.ListBuffer
+
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonSubTypes, JsonTypeInfo}
+
+/**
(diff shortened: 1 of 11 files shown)


server/src/main/scala/io/delta/standalone/internal/JsonPredicates.scala
```diff
@@ -0,0 +1,417 @@
...
+      case OpDataTypes.IntType => leftVal.toInt < rightVal.toInt
+      case OpDataTypes.LongType => leftVal.toLong < rightVal.toLong
+      case OpDataTypes.StringType => leftVal < rightVal
+      case OpDataTypes.DateType =>
+        java.sql.Date.valueOf(leftVal).before(java.sql.Date.valueOf(rightVal))
+      case _ =>
+        throw new IllegalArgumentException("Unsupported type: " + leftType)
+    }
+  }
+}
```
chakankardb, line 417: This file is mostly copied from spark connector.

server/src/test/scala/io/delta/standalone/internal/JsonPredicateSuite.scala
```diff
@@ -0,0 +1,470 @@
...
+    val numEval = 1000000
+
+    val start = System.currentTimeMillis()
+    for (i <- 1 to numEval) {
+      evalExpectBoolean(op, EvalContext(Map("date" -> dates(i % dates.size))))
+    }
+    val t = System.currentTimeMillis() - start
+    println("Stress test took " + t + " millis")
+  }
+}
```
chakankardb, line 471: copied from spark connector.

wchau (approved):

chakankardb:
@linzhou-db You want to take a quick look?

server/src/main/scala/io/delta/standalone/internal/JsonPredicates.scala
```diff
@@ -0,0 +1,417 @@
...
+ * The evaluation context in which operations will be evaluated.
+ *
+ * @param partitionValues The partition values associated with parquet files.
+ */
+case class EvalContext(partitionValues: Map[String, String])
+
+/**
+ * The data types supported by the filtering operations.
+ */
+object OpDataTypes {
```
linzhou-db, line 36: Seems like there's no way to use the same Jsonpredicates in both server/ and spark/?

linzhou-db (approved):

server/src/main/scala/io/delta/standalone/internal/JsonPredicates.scala
chakankardb: Ya, I had checked with Ryan Zu as well.
The way would be to make a common libraries, but that will trigger additional dependencies.
We can do that as a follow up change.

PR #270, opened 2023-02-15
Add support for Sql IN predicate.

chakankardb:
This PR adds support for Sql IN predicate.
We limit the size of the list to guard against pathological inputs.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/filters/OpConverter.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/filters/OpConverter.scala
@@ -53,6 +53,8 @@ import org.apache.spark.sql.types.{
 // should skip filtering in such cases.
 object OpConverter {
 
+  val kMaxSqlInOpSizeLimit = 20
+
   // Converts a set of SQL expression trees into a json predicate op.
   // We perform an implicit And across the input set at the top level.
   //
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/io/delta/sharing/spark/filters/OpConverter.scala
```diff
@@ -53,6 +53,8 @@ import org.apache.spark.sql.types.{
 // should skip filtering in such cases.
 object OpConverter {
 
+  val kMaxSqlInOpSizeLimit = 20
```
zhuansunxt, line 56: Can you add some documentation to the definition of this limit?

linzhou-db (approved):

spark/src/test/scala/io/delta/sharing/spark/filters/OpConverterSuite.scala
```diff
@@ -275,4 +275,32 @@ class OpConverterSuite extends SparkFunSuite {
...
+    val op = OpConverter.convert(Seq(sqlIn)).get
+    op.validate()
+    assert(op.evalExpectBoolean(EvalContext(Map("userId" -> "22"))) == false)
+    assert(op.evalExpectBoolean(EvalContext(Map("userId" -> "23"))) == true)
+    assert(op.evalExpectBoolean(EvalContext(Map("userId" -> "24"))) == true)
+    assert(op.evalExpectBoolean(EvalContext(Map("userId" -> "25"))) == true)
+    assert(op.evalExpectBoolean(EvalContext(Map("userId" -> "26"))) == true)
+    assert(op.evalExpectBoolean(EvalContext(Map("userId" -> "27"))) == false)
+
+    var sqlLiterals2: Seq[SqlExpression] = Seq.empty
```
zhuansunxt, line 297: Can you add a test case where the list is empty and make sure the op is evaluated to `false`?

zhuansunxt (approved):
LGTM. 2 quick comments.

spark/src/main/scala/io/delta/sharing/spark/filters/OpConverter.scala
chakankardb: Done.

spark/src/test/scala/io/delta/sharing/spark/filters/OpConverterSuite.scala
chakankardb: Spark will never send us an empty IN.
In fact, if you specify one entry in the IN operator in sql, spark converts that into an equality check in the expression tree it sends us.

However, I decided to handle both cases.
We throw an error if there are no predicates, and convert it to an equality if there is 1.

zhuansunxt (approved):

PR #273, opened 2023-02-16
test json predicates with flag set to false.

chakankardb:
This PR tests that the json predicate flag is honored by the client.


linzhou-db (approved):
Thanks!

Issue #277, opened 2023-02-28
Add support for expressing the sharing profile using JSON


goodwillpunning:
Currently, data recipients cannot specify the location of a sharing profile using Azure Synapse FS (`synfs:` schema). Rather than expanding support for `synfs` and future, alternate schemas, it may be more flexible to allow data recipients to specify the sharing profile as a JSON object, passing it as an argument to the Python sharing client constructor. For example:

```
import delta_sharing

sharing_profile = {
  "shareCredentialsVersion": 1,
  "endpoint": "https://sharing.delta.io/delta-sharing/",
  "bearerToken": "dapi1234567890"
}
client = delta_sharing.SharingClient(sharing_profile)
```

While it could be argued that the sharing access token is now clear text, a secrets manager client, like Azure key vault client for Python, could be used to securely access the sharing token.

PR #278, opened 2023-03-06
Add support for expressing sharing profile using a JSON

goodwillpunning:
Fixes [#277].

This commit adds support for specifying a sharing profile as a JSON string, bytes, or byte array to the Python connector. In particular, this commit adds the following:

- Adds a static function `_is_valid_json()` to the `delta_sharing` module that determines if the profile is a JSON string
- Adds a conditional check to the DeltaSharingClient constructor to determine if the profile should be `read_from_file()` or `from_json`
- Exposes the `DeltaSharingProfile` class, so that the sharing profile can be expressed using a `DeltaSharingProfile` object

### DIFF
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -210,8 +210,8 @@ class SharingClient:
     A Delta Sharing client to query shares/schemas/tables from a Delta Sharing Server.
     """
 
-    def __init__(self,
-                 profile: Union[str, bytes, bytearray, BinaryIO, TextIO, Path, DeltaSharingProfile]):
+    def __init__(self, profile: Union[str, bytes, bytearray,
+                                      BinaryIO, TextIO, Path, DeltaSharingProfile]):
         if _is_valid_json(profile):
             profile = DeltaSharingProfile.from_json(profile)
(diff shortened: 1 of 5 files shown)


python/delta_sharing/delta_sharing.py
```diff
@@ -51,6 +52,18 @@ def _parse_url(url: str) -> Tuple[str, str, str, str]:
     return (profile, share, schema, table)
 
 
+def _is_valid_json(json_str):
+    """Helper function that tests if a sharing profile is a valid JSON string
```
linzhou-db, line 56: not necessarily sharing profile, any input string?

python/delta_sharing/protocol.py
```diff
@@ -40,7 +40,7 @@ def __post_init__(self):
             )
 
     @staticmethod
-    def read_from_file(profile: Union[str, IO, Path]) -> "DeltaSharingProfile":
+    def read_from_file(profile: Union[str, bytes, bytearray, IO, Path]) -> "DeltaSharingProfile":
```
linzhou-db, line 43: What's this change for?

linzhou-db (commented):
Is there a way to add some tests?

PR #280, opened 2023-03-10
Governance APIs for the Delta Sharing Server

wchau:
This is still a draft in progress. The idea is to have APIs based off of the Databricks Unity Catalog APIs.


wchau:
> What's the context? and what would it be used for?

The context is that there have been multiple requests to not use a static config file, so instead, we want to give a simple option to allow SQL-backed management APIs

samgriek-arc:
Is anyone pushing this forward?  This would be super handy for my use case.

chakankardb:
@wchau Is this still a draft?

linzhou-db:
closing as it's stale.

PR #294, opened 2023-04-27
Optimize delta sharing spark client handling of presigned url response.

chakankardb:
In this PR, we optimize the way delta sharing client assembles the server response for pre-signed urls.
Instead of first concatenating and then splitting the response, we now avoid concatenation, making things more efficient.
It also avoid running into 1GB limit on string concatenation.


linzhou-db (approved):
So the new limitation would be the memory size?

chakankardb:
> So the new limitation would be the memory size?

Yes. We should discuss allowing a spill over to disk to scale this further.

PR #296, opened 2023-04-27
[Prototype] Build delta-sharing-client library

linzhou-db:
Build delta-sharing-client library


linzhou-db:
Moved to https://github.com/delta-io/delta-sharing/pull/334

PR #302, opened 2023-05-10
Support get_table_version/get_table_protocol/get_table_metadata in python

linzhou-db:
Support get_table_version/get_table_protocol/get_table_metadata as top api in python

### DIFF
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -63,7 +63,10 @@ def get_table_version(
     profile_json, share, schema, table = _parse_url(url)
     profile = DeltaSharingProfile.read_from_file(profile_json)
     rest_client=DataSharingRestClient(profile)
-    response = rest_client.query_table_version(Table(name=table, share=share, schema=schema))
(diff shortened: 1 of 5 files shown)


python/delta_sharing/delta_sharing.py
```diff
@@ -51,6 +51,53 @@ def _parse_url(url: str) -> Tuple[str, str, str, str]:
...
+def get_table_version(
+    url: str,
+    starting_timestamp: Optional[str] = None
+) -> int:
+    """
+    Get the shared table version using the given url.
+
+    :param url: a url under the format "<profile>#<share>.<schema>.<table>"
+    :param starting_timestamp: a string in the format of YYYY-MM-DDThh:mm:ssZ. Get the version at or
+      after the given timestamp.
```
chakankardb, line 63: Perhaps add that if this is not specified we get the latest version?

chakankardb (approved):

python/delta_sharing/delta_sharing.py
linzhou-db: added

PR #306, opened 2023-05-12
Set spark.delta.sharing.source.refreshPresignedUrls.enabled default to true

linzhou-db:
Set spark.delta.sharing.source.refreshPresignedUrls.enabled default to true


chakankardb (approved):

zsxwing (approved):
LGTM

PR #312, opened 2023-05-16
Add an optional `expirationTimestamp` field for pre signed urls

linzhou-db:
- In delta sharing sever, add expirationTimestamp for each action in the response
- In delta sharing client, update refresh function to return the min expiration timestamp across all urls.
- In delta sharing client, refresh based on the expiration timestamp
- In DeltaSharingSource, updated the refresh function to refresh the `sortedFetchedFiles`, so it doesn't have to do an extra local refresh -- delegate the refresh to CachedTableManager entirely.

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/model.scala
+++ b/server/src/main/scala/io/delta/sharing/server/model.scala
@@ -78,6 +78,7 @@ case class AddFile(
     size: Long,
     @JsonRawValue
     stats: String = null,
+    expirationTimestamp: Long,
     timestamp: java.lang.Long = null,
     version: java.lang.Long = null) extends Action {
 
@@ -90,6 +91,7 @@ case class AddFileForCDF(
     @JsonInclude(JsonInclude.Include.ALWAYS)
     partitionValues: Map[String, String],
     size: Long,
+    expirationTimestamp: Long,
     version: Long,
     timestamp: Long,
     @JsonRawValue
@@ -104,6 +106,7 @@ case class AddCDCFile(
     @JsonInclude(JsonInclude.Include.ALWAYS)
     partitionValues: Map[String, String],
     size: Long,
+    expirationTimestamp: Long,
     timestamp: Long,
     version: Long)
     extends Action {
@@ -117,6 +120,7 @@ case class RemoveFile(
     @JsonInclude(JsonInclude.Include.ALWAYS)
     partitionValues: Map[String, String],
     size: Long,
+    expirationTimestamp: Long,
     timestamp: Long,
     version: Long)
     extends Action {
--- a/server/src/main/scala/io/delta/sharing/server/model.scala
+++ b/server/src/main/scala/io/delta/sharing/server/model.scala
@@ -78,7 +78,7 @@ case class AddFile(
     size: Long,
     @JsonRawValue
     stats: String = null,
-    expirationTimestamp: Long,
+    expirationTimestamp: java.lang.Long = null,
     timestamp: java.lang.Long = null,
     version: java.lang.Long = null) extends Action {
 
@@ -91,7 +91,7 @@ case class AddFileForCDF(
     @JsonInclude(JsonInclude.Include.ALWAYS)
     partitionValues: Map[String, String],
     size: Long,
-    expirationTimestamp: Long,
+    expirationTimestamp: java.lang.Long = null,
     version: Long,
     timestamp: Long,
     @JsonRawValue
@@ -106,7 +106,7 @@ case class AddCDCFile(
     @JsonInclude(JsonInclude.Include.ALWAYS)
     partitionValues: Map[String, String],
     size: Long,
-    expirationTimestamp: Long,
+    expirationTimestamp: java.lang.Long = null,
     timestamp: Long,
     version: Long)
     extends Action {
@@ -120,7 +120,7 @@ case class RemoveFile(
     @JsonInclude(JsonInclude.Include.ALWAYS)
     partitionValues: Map[String, String],
     size: Long,
-    expirationTimestamp: Long,
+    expirationTimestamp: java.lang.Long = null,
     timestamp: Long,
     version: Long)
     extends Action {
--- a/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
@@ -266,12 +266,18 @@ class DeltaSharedTable(
           } else {
             filteredFiles
           }
+        var setUrlExpirationTimestamp: Boolean = false
         filteredFiles.map { addFile =>
           val cloudPath = absolutePath(deltaLog.dataPath, addFile.path)
           val signedUrl = fileSigner.sign(cloudPath)
           val modelAddFile = model.AddFile(
-            url = signedUrl,
-            expirationTimestamp = fileSigner.getExpirationTimestamp,
+            url = signedUrl.url,
+            expirationTimestamp = if (!setUrlExpirationTimestamp) {
+              setUrlExpirationTimestamp = true
+              signedUrl.expirationTimestamp
+            } else {
+              null
+            },
             id = Hashing.md5().hashString(addFile.path, UTF_8).toString,
             partitionValues = addFile.partitionValues,
             size = addFile.size,
@@ -303,15 +309,22 @@ class DeltaSharedTable(
     )
 
     val actions = ListBuffer[model.SingleAction]()
+    var setUrlExpirationTimestamp: Boolean = false
     deltaLog.getChanges(startingVersion, true).asScala.toSeq.foreach{versionLog =>
       val v = versionLog.getVersion
       val versionActions = versionLog.getActions.asScala.map(x => ConversionUtils.convertActionJ(x))
       val ts = timestampsByVersion.get(v).orNull
       versionActions.foreach {
         case a: AddFile if a.dataChange =>
+          val signedUrl = fileSigner.sign(absolutePath(deltaLog.dataPath, a.path))
           val modelAddFile = model.AddFileForCDF(
-            url = fileSigner.sign(absolutePath(deltaLog.dataPath, a.path)),
-            expirationTimestamp = fileSigner.getExpirationTimestamp,
+            url = signedUrl.url,
+            expirationTimestamp = if (!setUrlExpirationTimestamp) {
+              setUrlExpirationTimestamp = true
+              signedUrl.expirationTimestamp
+            } else {
+              null
+            },
             id = Hashing.md5().hashString(a.path, UTF_8).toString,
             partitionValues = a.partitionValues,
             size = a.size,
@@ -321,9 +334,15 @@ class DeltaSharedTable(
           )
           actions.append(modelAddFile.wrap)
         case r: RemoveFile if r.dataChange =>
+          val signedUrl = fileSigner.sign(absolutePath(deltaLog.dataPath, r.path))
           val modelRemoveFile = model.RemoveFile(
-            url = fileSigner.sign(absolutePath(deltaLog.dataPath, r.path)),
-            expirationTimestamp = fileSigner.getExpirationTimestamp,
+            url = signedUrl.url,
+            expirationTimestamp = if (!setUrlExpirationTimestamp) {
+              setUrlExpirationTimestamp = true
+              signedUrl.expirationTimestamp
+            } else {
+              null
+            },
             id = Hashing.md5().hashString(r.path, UTF_8).toString,
             partitionValues = r.partitionValues,
             size = r.size.get,
@@ -405,14 +424,20 @@ class DeltaSharedTable(
         actions.append(modelMetadata.wrap)
       }
     }
+    var setUrlExpirationTimestamp: Boolean = false
     changeFiles.foreach { cdcDataSpec =>
       cdcDataSpec.actions.foreach { action =>
         val addCDCFile = action.asInstanceOf[AddCDCFile]
         val cloudPath = absolutePath(deltaLog.dataPath, addCDCFile.path)
         val signedUrl = fileSigner.sign(cloudPath)
         val modelCDCFile = model.AddCDCFile(
-          url = signedUrl,
-          expirationTimestamp = fileSigner.getExpirationTimestamp,
+          url = signedUrl.url,
+          expirationTimestamp = if (!setUrlExpirationTimestamp) {
+            setUrlExpirationTimestamp = true
+            signedUrl.expirationTimestamp
+          } else {
+            null
+          },
           id = Hashing.md5().hashString(addCDCFile.path, UTF_8).toString,
           partitionValues = addCDCFile.partitionValues,
           size = addCDCFile.size,
@@ -428,8 +453,13 @@ class DeltaSharedTable(
         val cloudPath = absolutePath(deltaLog.dataPath, addFile.path)
         val signedUrl = fileSigner.sign(cloudPath)
         val modelAddFile = model.AddFileForCDF(
-          url = signedUrl,
-          expirationTimestamp = fileSigner.getExpirationTimestamp,
+          url = signedUrl.url,
+          expirationTimestamp = if (!setUrlExpirationTimestamp) {
+            setUrlExpirationTimestamp = true
+            signedUrl.expirationTimestamp
+          } else {
+            null
+          },
           id = Hashing.md5().hashString(addFile.path, UTF_8).toString,
           partitionValues = addFile.partitionValues,
           size = addFile.size,
@@ -446,8 +476,13 @@ class DeltaSharedTable(
         val cloudPath = absolutePath(deltaLog.dataPath, removeFile.path)
         val signedUrl = fileSigner.sign(cloudPath)
         val modelRemoveFile = model.RemoveFile(
-          url = signedUrl,
-          expirationTimestamp = fileSigner.getExpirationTimestamp,
+          url = signedUrl.url,
+          expirationTimestamp = if (!setUrlExpirationTimestamp) {
+            setUrlExpirationTimestamp = true
+            signedUrl.expirationTimestamp
+          } else {
+            null
+          },
           id = Hashing.md5().hashString(removeFile.path, UTF_8).toString,
           partitionValues = removeFile.partitionValues,
           size = removeFile.size.get,
--- a/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
@@ -269,7 +269,9 @@ class DeltaSharedTable(
         filteredFiles.map { addFile =>
           val cloudPath = absolutePath(deltaLog.dataPath, addFile.path)
           val signedUrl = fileSigner.sign(cloudPath)
-          val modelAddFile = model.AddFile(url = signedUrl,
+          val modelAddFile = model.AddFile(
+            url = signedUrl,
+            expirationTimestamp = fileSigner.getExpirationTimestamp,
             id = Hashing.md5().hashString(addFile.path, UTF_8).toString,
             partitionValues = addFile.partitionValues,
             size = addFile.size,
@@ -309,6 +311,7 @@ class DeltaSharedTable(
         case a: AddFile if a.dataChange =>
           val modelAddFile = model.AddFileForCDF(
             url = fileSigner.sign(absolutePath(deltaLog.dataPath, a.path)),
+            expirationTimestamp = fileSigner.getExpirationTimestamp,
             id = Hashing.md5().hashString(a.path, UTF_8).toString,
             partitionValues = a.partitionValues,
             size = a.size,
@@ -320,6 +323,7 @@ class DeltaSharedTable(
         case r: RemoveFile if r.dataChange =>
           val modelRemoveFile = model.RemoveFile(
             url = fileSigner.sign(absolutePath(deltaLog.dataPath, r.path)),
+            expirationTimestamp = fileSigner.getExpirationTimestamp,
             id = Hashing.md5().hashString(r.path, UTF_8).toString,
             partitionValues = r.partitionValues,
             size = r.size.get,
@@ -408,6 +412,7 @@ class DeltaSharedTable(
         val signedUrl = fileSigner.sign(cloudPath)
         val modelCDCFile = model.AddCDCFile(
           url = signedUrl,
+          expirationTimestamp = fileSigner.getExpirationTimestamp,
           id = Hashing.md5().hashString(addCDCFile.path, UTF_8).toString,
           partitionValues = addCDCFile.partitionValues,
           size = addCDCFile.size,
@@ -424,6 +429,7 @@ class DeltaSharedTable(
         val signedUrl = fileSigner.sign(cloudPath)
         val modelAddFile = model.AddFileForCDF(
           url = signedUrl,
+          expirationTimestamp = fileSigner.getExpirationTimestamp,
           id = Hashing.md5().hashString(addFile.path, UTF_8).toString,
           partitionValues = addFile.partitionValues,
           size = addFile.size,
@@ -441,6 +447,7 @@ class DeltaSharedTable(
         val signedUrl = fileSigner.sign(cloudPath)
         val modelRemoveFile = model.RemoveFile(
           url = signedUrl,
+          expirationTimestamp = fileSigner.getExpirationTimestamp,
           id = Hashing.md5().hashString(removeFile.path, UTF_8).toString,
           partitionValues = removeFile.partitionValues,
           size = removeFile.size.get,
--- a/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
@@ -266,18 +266,12 @@ class DeltaSharedTable(
           } else {
             filteredFiles
           }
-        var setUrlExpirationTimestamp: Boolean = false
         filteredFiles.map { addFile =>
           val cloudPath = absolutePath(deltaLog.dataPath, addFile.path)
           val signedUrl = fileSigner.sign(cloudPath)
           val modelAddFile = model.AddFile(
             url = signedUrl.url,
-            expirationTimestamp = if (!setUrlExpirationTimestamp) {
-              setUrlExpirationTimestamp = true
-              signedUrl.expirationTimestamp
-            } else {
-              null
-            },
+            expirationTimestamp = signedUrl.expirationTimestamp,
             id = Hashing.md5().hashString(addFile.path, UTF_8).toString,
             partitionValues = addFile.partitionValues,
             size = addFile.size,
@@ -309,7 +303,6 @@ class DeltaSharedTable(
     )
 
     val actions = ListBuffer[model.SingleAction]()
-    var setUrlExpirationTimestamp: Boolean = false
     deltaLog.getChanges(startingVersion, true).asScala.toSeq.foreach{versionLog =>
       val v = versionLog.getVersion
       val versionActions = versionLog.getActions.asScala.map(x => ConversionUtils.convertActionJ(x))
@@ -319,12 +312,7 @@ class DeltaSharedTable(
           val signedUrl = fileSigner.sign(absolutePath(deltaLog.dataPath, a.path))
           val modelAddFile = model.AddFileForCDF(
             url = signedUrl.url,
-            expirationTimestamp = if (!setUrlExpirationTimestamp) {
-              setUrlExpirationTimestamp = true
-              signedUrl.expirationTimestamp
-            } else {
-              null
-            },
+            expirationTimestamp = signedUrl.expirationTimestamp,
             id = Hashing.md5().hashString(a.path, UTF_8).toString,
             partitionValues = a.partitionValues,
             size = a.size,
@@ -337,12 +325,7 @@ class DeltaSharedTable(
           val signedUrl = fileSigner.sign(absolutePath(deltaLog.dataPath, r.path))
           val modelRemoveFile = model.RemoveFile(
             url = signedUrl.url,
-            expirationTimestamp = if (!setUrlExpirationTimestamp) {
-              setUrlExpirationTimestamp = true
-              signedUrl.expirationTimestamp
-            } else {
-              null
-            },
+            expirationTimestamp = signedUrl.expirationTimestamp,
             id = Hashing.md5().hashString(r.path, UTF_8).toString,
             partitionValues = r.partitionValues,
             size = r.size.get,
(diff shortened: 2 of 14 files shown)


spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -81,9 +95,21 @@ class CachedTableManager(
...
+          // scalastyle:off println
+          Console.println(s"----[linzhou]------refreshed, timeToExpier:" +
+            s"${expiration - System.currentTimeMillis()}")
           val newTable = new CachedTable(
-            preSignedUrlExpirationMs + System.currentTimeMillis(),
-            cachedTable.refresher(),
+            if (isValidUrlExpirationTime(expiration)) {
+              Console.println(s"----[linzhou]------valid:" +
+                s"${expiration - System.currentTimeMillis()}")
+              expiration
```
linzhou-db, line 106: A question is, if the customer by anychance set `preSignedUrlExpirationMs` to a smaller number, do we respect `preSignedUrlExpirationMs` or still respect `expiration` of the urls

@chakankardb @zsxwing

chakankardb: Can we pick the more conservative value between the two?

linzhou-db: no, a conservative value (if you mean larger value) may result in an expired url.

chakankardb: @linzhou-db In this context, conservative would be min... :)

linzhou-db: sg, updated.

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -78,6 +78,7 @@ case class AddFile(
     size: Long,
     @JsonRawValue
     stats: String = null,
+    expirationTimestamp: java.lang.Long = null,
```
chakankardb, line 81: Do we need protocol change for this?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -219,6 +217,45 @@ case class DeltaSharingSource(
     }
   }
 
+  private def refreshSortedFetchedFiles(newIdToUrl: Map[String, String]): Unit = {
```
chakankardb, line 220: Consider adding a function header.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -240,48 +277,90 @@ case class DeltaSharingSource(
...
       val tableFiles = deltaLog.client.getFiles(
         deltaLog.table, Nil, None, Some(fromVersion), None, None
       )
       latestRefreshFunc = () => {
-        deltaLog.client.getFiles(
+        val files = deltaLog.client.getFiles(
           deltaLog.table, Nil, None, Some(fromVersion), None, None
-        ).files.map { f =>
+        ).files
+        var minUrlExpiration: Long = -1L
```
chakankardb, line 291: Consider initializing this to Long.MaxValue, so that the min will work.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -240,48 +277,90 @@ case class DeltaSharingSource(
...
+            if (minUrlExpiration == -1L) {
+              minUrlExpiration = f.expirationTimestamp
+            } else {
+              minUrlExpiration.min(f.expirationTimestamp)
+            }
+          }
           f.id -> f.url
         }.toMap
+
+        refreshSortedFetchedFiles(idToUrl)
```
chakankardb, line 303: Does this fix the double fetch issue?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -240,48 +277,90 @@ case class DeltaSharingSource(
       isStartingVersion: Boolean,
       currentLatestVersion: Long): Unit = {
     lastQueryTableTimestamp = System.currentTimeMillis()
+    minUrlExpirationTimestamp = -1
```
chakankardb, line 280: Initialize this to Long.MaxValue?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -560,7 +612,12 @@ case class DeltaSharingSource(
       idToUrl,
       Seq(new WeakReference(fileIndex)),
       params.profileProvider,
-      latestRefreshFunc
+      latestRefreshFunc,
+      if (minUrlExpirationTimestamp != -1) {
+        minUrlExpirationTimestamp
+      } else {
+        lastQueryTableTimestamp + CachedTableManager.INSTANCE.preSignedUrlExpirationMs
```
chakankardb, line 619: Should we use refreshThresholdMs in this case as before?

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
```diff
@@ -71,8 +81,9 @@ object DeltaSharingCDFReader {
       removeFiles: Seq[RemoveFile],
       schema: StructType,
       isStreaming: Boolean,
-      refresher: () => Map[String, String],
-      lastQueryTableTimestamp: Long = System.currentTimeMillis()
+      refresher: () => (Map[String, String], Long),
+      lastQueryTableTimestamp: Long,
+      minUrlExpirationTimestamp: Long
```
chakankardb, line 86: Should we just call this expirationTimestamp?

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -42,7 +42,7 @@ class CachedTable(
     val idToUrl: Map[String, String],
     val refs: Seq[WeakReference[AnyRef]],
     @volatile var lastAccess: Long,
-    val refresher: () => Map[String, String])
+    val refresher: () => (Map[String, String], Long))
```
chakankardb, line 45: Consider updating the comment about refresh in function header.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -63,6 +63,14 @@ class CachedTableManager(
     thread
   }
 
+  def isValidUrlExpirationTime(expiration: Long): Boolean = {
+    expiration > (System.currentTimeMillis() + refreshThresholdMs)
```
chakankardb, line 67: Can you document why we need the extra refreshThresholdMs ?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -240,48 +277,90 @@ case class DeltaSharingSource(
...
+            } else {
+              minUrlExpiration.min(f.expirationTimestamp)
+            }
+          }
           f.id -> f.url
         }.toMap
+
+        refreshSortedFetchedFiles(idToUrl)
+
+        (idToUrl, minUrlExpiration)
```
chakankardb, line 305: If we also set the minUrlExpirationTimestamp as part of  refreshSortedFetchedFiles, can we avoid the complexity of always checking for -1?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -240,48 +277,90 @@ case class DeltaSharingSource(
...
+                  file.expirationTimestamp
+                ),
+                isLast = (index + 1 == numFiles))
+            )
+          }
         // For files with index <= fromIndex, skip them, otherwise an exception will be thrown.
         case _ => ()
       }
+      if (!CachedTableManager.INSTANCE.isValidUrlExpirationTime(minUrlExpirationTimestamp)) {
+        // reset to -1 to indicate that it's not a valid url expiration timestamp.
```
chakankardb, line 340: Is this correct?
Since this is returned by the server, we should simply trust it and do refresh as needed (perhaps log an error).
Otherwise the urls will expire anyways.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -266,10 +266,18 @@ class DeltaSharedTable(
...
+        var setUrlExpirationTimestamp: Boolean = false
         filteredFiles.map { addFile =>
           val cloudPath = absolutePath(deltaLog.dataPath, addFile.path)
           val signedUrl = fileSigner.sign(cloudPath)
-          val modelAddFile = model.AddFile(url = signedUrl,
+          val modelAddFile = model.AddFile(
+            url = signedUrl.url,
+            expirationTimestamp = if (!setUrlExpirationTimestamp) {
+              setUrlExpirationTimestamp = true
+              signedUrl.expirationTimestamp
```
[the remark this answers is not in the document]
linzhou-db: updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -132,9 +132,6 @@ case class DeltaSharingSource(
 
   private val tableId = initSnapshot.metadata.id
 
-  private val refreshPresignedUrls = spark.sessionState.conf.getConfString(
-    "spark.delta.sharing.source.refreshPresignedUrls.enabled", "true").toBoolean
```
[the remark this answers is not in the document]
linzhou-db: yes, or in another word, we are not refreshing in DeltaShairngSource, but let the CachedTableManager to refresh, and update sortedFetchedFiles in DeltaSharingSource(as you suggested I remember?), so there's no place to use this flag anymore.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: added

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: yes, it will fix it. This is what you/Ryan suggested.
The refresh only happens CachedTableManager, and will update sortedFetchedFiles

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: hmmm, if this is set to a timestamp expiring very soon and we always respect it, then the client will be refreshing endlessly, refresh after refresh...
Is that desired?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: Could you elaborate a bit?

`refreshThresholdMs` is the buffer time before expiration for the refresh, how is it related to the value here?

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
linzhou-db: updated.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
linzhou-db: updated.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
linzhou-db: added.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: we avoid checking for -1 by setting it as Long.MaxValue.

What do you mean by set the minUrlExpieraetionTimestamp as part of refrshSortedFetchedFiles?

server/src/main/scala/io/delta/sharing/server/model.scala
linzhou-db: updated

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingProfileProvider.scala
```diff
@@ -47,7 +47,8 @@ trait DeltaSharingProfileProvider {
 
   def getCustomTablePath(tablePath: String): String = tablePath
 
-  def getCustomRefresher(refresher: () => Map[String, String]): () => Map[String, String] = {
+  def getCustomRefresher(refresher: () => (Map[String, String], Long)): () =>
```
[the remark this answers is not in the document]
linzhou-db: added

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -63,6 +64,16 @@ class CachedTableManager(
     thread
   }
 
+  def isValidUrlExpirationTime(expiration: Long): Boolean = {
+    // refreshThresholdMs is the buffer time for the refresh RPC.
+    // It could also help the client from keeping refreshing endlessly.
+    expiration > (System.currentTimeMillis() + refreshThresholdMs)
+  }
+
+  def getFomattedTS(ts: Long): String = {
```
zsxwing, line 73: unused method?

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -81,9 +92,14 @@ class CachedTableManager(
         logInfo(s"Updating pre signed urls for $tablePath (expiration time: " +
           s"${new java.util.Date(cachedTable.expiration)})")
         try {
+          val (idToUrl, expiration) = cachedTable.refresher()
           val newTable = new CachedTable(
-            preSignedUrlExpirationMs + System.currentTimeMillis(),
-            cachedTable.refresher(),
+            if (isValidUrlExpirationTime(expiration)) {
+              expiration.min(preSignedUrlExpirationMs + System.currentTimeMillis())
```
zsxwing, line 98: The config `preSignedUrlExpirationMs` is because we don't know the expiration time. If the server tells us the expiration time, why do we still need to use `preSignedUrlExpirationMs`?

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -81,9 +92,14 @@ class CachedTableManager(
...
+            } else {
+              preSignedUrlExpirationMs + System.currentTimeMillis()
```
zsxwing, line 100: If `isValidUrlExpirationTime` returns `false`,  it means `expiration <= (System.currentTimeMillis() + refreshThresholdMs)`. There are two possible cases:
- -1: the server doesn't tell us the expiration time. In this case, we should use `preSignedUrlExpirationMs + System.currentTimeMillis()`, which is the existing behavior.
- The server sends us a pre signed URL whose expiration time that is about to expire. This could be a few possible cases:
  -  clock differences between the server and the client is too large.
  - the client `refreshThresholdMs` is too large.
  - other unknown issues

Is `preSignedUrlExpirationMs + System.currentTimeMillis()` the right value for the second case? Do you want to avoid refreshing URLs every minute for the second case? Anyway, a warning log here would be pretty helpful to debug the second case.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -225,6 +223,55 @@ case class DeltaSharingSource(
     }
   }
 
+  // Function to be called in latestRefreshFunc, to refresh the pre-signed urls in
+  // sortedFetchedFiles. newIdToUrl contains the refreshed urls.
+  private def refreshSortedFetchedFiles(newIdToUrl: Map[String, String]): Unit = {
```
zsxwing, line 228: We need a lock on this method and all places that modify `sortedFetchedFiles` to avoid race conditions. For example,

- The streaming thread changes it to another `sortedFetchedFiles`.
- The refresh thread changes it back to the old `sortedFetchedFiles`.

Could you also add `@volatile` to `sortedFetchedFiles` so that the streaming thread can see the latest value set by the refresh thread?

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
```diff
@@ -111,6 +126,33 @@ object DeltaSharingCDFReader {
       removeFiles.map(r => r.id -> r.url).toMap
   }
 
+  def getMinUrlExpiration(
+      addFiles: Seq[AddFileForCDF],
+      cdfFiles: Seq[AddCDCFile],
+      removeFiles: Seq[RemoveFile]
+  ): Long = {
+    var minUrlExpiration: Long = Long.MaxValue
```
zsxwing, line 134: Can we use Option[Long] and use `Some(v)` to indicate that we know the expiration timestamp, and None to indicate that we don't know the expiration time rather than -1 or Long.MaxValue?

Otherwise, for each place using the expiration time, we have to think about three cases: a valid value, -1 and Long.MaxValue

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: Will add.
Though it's not necessary as getBatch is a blocking method, so there won't be multiple threads trying to access `sortedFetchedFiles` at the same time.

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
linzhou-db: sg, will make the change.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
linzhou-db: was used for debugging, will remove.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
linzhou-db: makes sense, I'll just use expiration then.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
linzhou-db: yeah, as you described, the reason is not clear, so I'm not sure what's the best behavior, thus I just use what we are currently using. I'm hesitate to let it refresh every minute because the refresh itself anyway takes time (likely longer than 1 min), and may not catch up with `expiration`, so I'd rather let it fail/expire and then debug.

Good idea to add a warning log here, I'll add it with all necessary info.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
chakankardb: The refresh will be called from the CacheTable as well, but the read happens in the main thread. 
So we should synchronize access to sortedFetchFiles.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingProfileProvider.scala
```diff
@@ -47,7 +47,10 @@ trait DeltaSharingProfileProvider {
 
   def getCustomTablePath(tablePath: String): String = tablePath
 
-  def getCustomRefresher(refresher: () => Map[String, String]): () => Map[String, String] = {
+  // Map[String, String] is the id to url map.
+  // Long is the minimum url expiration time for all the urls.
```
chakankardb, line 51: Is this min across all urls, or just the expiry for the one?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -247,37 +295,66 @@ case class DeltaSharingSource(
...
+              minUrlExpiration
+            } else {
+              Some(f.expirationTimestamp)
+            }
+          }
           f.id -> f.url
         }.toMap
+
+        refreshSortedFetchedFiles(idToUrl)
+        minUrlExpirationTimestamp = minUrlExpiration
```
chakankardb, line 324: Either make this atomic, or protect it with synchronized.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
chakankardb: It will be desirable if we update the sortedFetchedFiles and minUrlExpirationTime in one function together.
It would be easy to apply some config (for example, to enable/disable this) to that one place if needed.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: added.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingProfileProvider.scala
linzhou-db: across all urls.

linzhou-db: it's in the comment.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: added synchronized in all places.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -493,57 +635,15 @@ case class DeltaSharingSource(
...
-        )
+    var fileActions = Seq[IndexedFile]()
+    synchronized {
+      fileActions = sortedFetchedFiles.takeWhile {
+        case IndexedFile(version, index, _, _, _, _) =>
+          version < endOffset.tableVersion ||
+            (version == endOffset.tableVersion && index <= endOffset.index)
       }
+      sortedFetchedFiles = sortedFetchedFiles.drop(fileActions.size)
     }
```
chakankardb, line 646: Consider defining a method:
popSortedFetchFiles(endOffset)

The synchronization should be localized to setters/getters/modifiers

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -579,13 +679,20 @@ case class DeltaSharingSource(
...
+        Seq(new WeakReference(fileIndex)),
+        params.profileProvider,
+        latestRefreshFunc,
+        if (minUrlExpirationTimestamp.isDefined) {
+          minUrlExpirationTimestamp.get
+        } else {
+          lastQueryTableTimestamp + CachedTableManager.INSTANCE.preSignedUrlExpirationMs
+        }
+      )
+    }
```
chakankardb, line 695: This synchronize is not needed if we define getMinUrlExpirationTimestamp (and sync within that).

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -616,17 +723,20 @@ case class DeltaSharingSource(
...
+        addFiles,
+        cdfFiles,
+        removeFiles,
+        schema,
+        isStreaming = true,
+        latestRefreshFunc,
+        lastQueryTableTimestamp,
+        minUrlExpirationTimestamp
+      )
+    }
```
chakankardb, line 739: as above.

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaCDFRelation.scala
```diff
@@ -111,6 +126,48 @@ object DeltaSharingCDFReader {
       removeFiles.map(r => r.id -> r.url).toMap
   }
 
+  def getMinUrlExpiration(
```
chakankardb, line 129: Add a function header.

linzhou-db: added.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -355,6 +486,17 @@ case class DeltaSharingSource(
...
+    synchronized {
+      minUrlExpirationTimestamp = DeltaSharingCDFReader.getMinUrlExpiration(
+        tableFiles.addFiles,
+        tableFiles.cdfFiles,
+        tableFiles.removeFiles
+      )
+      if (!CachedTableManager.INSTANCE.isValidUrlExpirationTime(minUrlExpirationTimestamp)) {
+        // reset to None to indicate that it's not a valid url expiration timestamp.
+        minUrlExpirationTimestamp = None
+      }
```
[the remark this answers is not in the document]
linzhou-db: good point, updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: added.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: makes sense, updated.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: updated.

chakankardb (approved):

chakankardb:
@zsxwing Can you also please take another look?

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -225,6 +240,109 @@ case class DeltaSharingSource(
...
   }
 
+  private def resetGlobalTimestamp(): Unit = {
+    synchronized {
+      lastQueryTableTimestamp = System.currentTimeMillis()
+      minUrlExpirationTimestamp = None
+    }
+  }
+
+  private def validateMinUrlExpirationTimestamp(inputTimestamp: Option[Long] = None): Unit = {
```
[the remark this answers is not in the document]
linzhou-db: added.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -579,14 +709,21 @@ case class DeltaSharingSource(
     val params = new RemoteDeltaFileIndexParams(
       spark, initSnapshot, deltaLog.client.getProfileProvider)
     val fileIndex = new RemoteDeltaBatchFileIndex(params, addFilesList)
+    val urlExpirationTimestamp = getMinUrlExpirationTimestamp()
```
zsxwing, line 712: `sortedFetchedFiles`, `lastQueryTableTimestamp` and `minUrlExpirationTimestamp` should probably be accessed together in the same `synchronized`. Right now there is a race condition here: if a refresh is done before line 712, then `urlExpirationTimestamp` will be for the latest URLs, but `idToUrl` is still using the old URLs. This would cause us to use a wrong bigger `urlExpirationTimestamp`. Since `CachedTableManager` sees a larger `urlExpirationTimestamp`, it will not refresh before these old URLs expire and cause token expiration issue.

spark/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -295,10 +304,25 @@ class RemoteSnapshot(
           Seq(new WeakReference(fileIndex)),
           fileIndex.params.profileProvider,
           () => {
-            client.getFiles(table, Nil, None, versionAsOf, timestampAsOf, jsonPredicateHints)
-            .files.map { add =>
+            val files = client.getFiles(
+              table, Nil, None, versionAsOf, timestampAsOf, jsonPredicateHints).files
+            var minUrlExpiration: Option[Long] = None
+            val idToUrl = files.map { add =>
+              if (add.expirationTimestamp != null) {
```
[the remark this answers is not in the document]
linzhou-db: updeted.

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: Good catch. Returned the three parameters together from `popSortedFetchedFiles`.

spark/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -63,6 +64,18 @@ class CachedTableManager(
     thread
   }
 
+  def isValidUrlExpirationTime(expiration: Option[Long]): Boolean = {
+    // refreshThresholdMs is the buffer time for the refresh RPC.
+    // It could also help the client from keeping refreshing endlessly.
+    val isValid = expiration.isDefined && (
+      expiration.get > (System.currentTimeMillis() + refreshThresholdMs))
+    if (!isValid) {
```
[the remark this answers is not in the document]
linzhou-db: updated.

linzhou-db:
Manually tested in https://github.com/delta-io/delta-sharing/pull/313

PR #314, opened 2023-05-28
Retry on java.net.SocketTimeoutException

linzhou-db:
Retry on java.net.SocketTimeoutException, and add logging for why not retry.


chakankardb (approved):
Thanks for fixing this!

chakankardb:
should we check this into master as well?

spark/src/main/scala/io/delta/sharing/spark/util/RetryUtils.scala
```diff
@@ -39,6 +39,8 @@ private[sharing] object RetryUtils extends Logging {
           logWarning(s"Sleeping $sleepMs ms to retry because of error: ${e.getMessage}", e)
           sleeper(sleepMs)
           sleepMs *= 2
+        case e: Exception =>
+          logError(s"Not retrying delta sharing rpc on error: ${e.getMessage}", e)
```
chakankardb, line 43: I think this is problematic.... we need to throw the exception.

PR #315, opened 2023-05-29
Fix the error order of client retry

linzhou-db:
Fix the error order of client retry, and throw exception when not retrying.


chakankardb (approved):

chakankardb:
Thanks for fixing this!

Issue #320, opened 2023-06-05
Getting Caused by: UnexpectedHttpStatus: HTTP request failed with status: HTTP/1.1 403 while reading with Proxy Configuration


pavan-kumar-chalamcharla:
When using the proxy configuration such as http_proxy and https_proxy environment variables when using the open delt-sharing for reading the data the proxy environment variables are not picked up causing the below error when using the bucket policy which allows only proxy IPs:
FileReadException: Error while reading file delta-sharing:/dbfs%253A%252FFileStore%252F059501f2aeb8fad0607470b70008727a/62477. Caused by: UnexpectedHttpStatus: HTTP request failed with status: HTTP/1.1 403 Forbidden <?xml version="1.0" encoding="UTF-8"?>
**DEBUG:fsspec.http**:Cannot connect to host [<bucket name>.s3.us-east-1.amazonaws.com](https://<bucketname>.s3.us-east-1.amazonaws.com/):443 ssl:default [Connect call failed ('', 443)]

when debugging it further we see that the fsspec.http/aiohttp is used when reading the pre-signed URLs and those libraries are not using the HTTP_PROXY env variables that are set and causing the failure while reading the data.

We are looking for support of proxy with open delta sharing while reading data via delta-sharing python libraries.

linzhou-db:
Do they already have an idea of how to fix the issue? 
If so, feel free to send out a PR, as this is oss code.

quertenmont:
facing the same issue....

quertenmont:
After your fix #326 , I can go one step further with my proxy configuration, but I am still having troubles

1) if I use a https_proxy, I get the following error emitted from aiohttp
HTTPS proxies https://mitmproxy:8080/ are not supported, ignoring

2) If I replace my proxy configuration to use http instead of https,
then it's my proxy server that complain, because the TLS handshake is failling

```
[09:08:48.665][10.244.13.19:51180] client connect
[09:08:48.860][10.244.13.19:51180] server connect open-delta-sharing.s3.us-west-2.amazonaws.com:443 (52.218.196.225:443)
[09:08:49.207][10.244.13.19:51180] Client TLS handshake failed. The client does not trust the proxy's certificate for open-delta-sharing.s3.us-west-2.amazonaws.com (tlsv1 alert unknown ca)
[09:08:49.208][10.244.13.19:51180] client disconnect
```

Any idea how I can sort this out ?
Thanks in advance
Loic

[...]

PR #326, opened 2023-06-13
Fix for issue #320

pavan-kumar-chalamcharla:
The issue with the proxy env variables not being used when reading the data via the pre-signed urls when using the "load_as_pandas" which uses the fsspec and aiohttp libraries to read the data via the pre-signed urls.

this fix will make sure that the fsspec and aiohttp clients created will use the proxy settings if the proxy env variables are set in the environment.

### DIFF
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -161,7 +161,7 @@ class DeltaSharingReader:
         protocol = url.scheme
         proxy = getproxies()
         if len(proxy) != 0:
-            filesystem = fsspec.filesystem(protocol,trust_env=True)
+            filesystem = fsspec.filesystem(protocol, client_kwargs={"trust_env":True})
         else:
             filesystem = fsspec.filesystem(protocol)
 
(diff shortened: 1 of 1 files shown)


python/delta_sharing/reader.py
```diff
@@ -158,7 +159,11 @@ def _to_pandas(
             import delta_sharing._yarl_patch  # noqa: F401
 
         protocol = url.scheme
-        filesystem = fsspec.filesystem(protocol)
+        proxy = getproxies()
```
linzhou-db, line 162: are we able to test these code? it's fine to test in a real environment if it's hard to do unit test.

or could you comment on what does this do and an example of what `proxy` could look like? and why not always set `trust_env` to true?

pavan-kumar-chalamcharla: yes, I have tested it inhouse and also in customer env and it works as expected.

but never tested what would happen if trust_env=True is set everytime will try to test it once on what would happen.

pavan-kumar-chalamcharla: I have tested it again I see that even setting "trust_env" to true always also works. it would pick the HTTP proxy env variables if available else it wouldn't.

if you can also test it further

linzhou-db: thanks!

PR #332, opened 2023-06-22
[Prototype] E2E testing with queryDeltaLog

linzhou-db:
1. client able to query delta log [done]
2. Able to make rpc with RemoteDeltaFileIndex [done]
3. spark able to parse response with queryDeltaLog[done]
4. delta sharing file system able to store and serve the response as deltalog with the help of SparkEnv.get.blockManager[done]
5. delta log able to produce BaseRelation or DataFrame with the response.[done]

TODOs:
1. check the data size block manager could handle.
2. verify blockmanager could be used for Driver/Executor communication.
3. check the DeltaSharingFileSystem.decode() logic.
4. We need another DeltaSharingLogFileSystem. 
5. convert partitionFilters to jsonPredicateHints
6. We'll NOT prototype cdf and streaming queries.


linzhou-db:
this PR is done with its job

PR #334, opened 2023-06-24
Code Refactoring: Move delta-sharing-client out as a separate directory

linzhou-db:
Code Refactoring: Move delta-sharing-client out as a separate directory


charlenelyu-db (approved):

client/src/main/scala/io/delta/sharing/client/filters/JsonPredicates.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.delta.sharing.spark.filters
+package io.delta.sharing.client.filters
```
chakankardb, line 17: Why do we need to move the predicates?
The client treats them as a json string right?

client/src/main/scala/io/delta/sharing/client/filters/OpConverter.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.delta.sharing.spark.filters
+package io.delta.sharing.client.filters
```
chakankardb, line 17: as above.

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.delta.sharing.spark.util
+package io.delta.sharing.client.util
```
chakankardb, line 17: This seems odd. These are all spark confs. Why should they belong int the client?

linzhou-db: ConfUtils  is actually used by DeltaSharingFileSystem, and used to initialize DeltaSharingRestClient in RemoteDeltaLog, I'll refactor that later as well.

client/src/main/scala/io/delta/sharing/client/filters/JsonPredicates.scala
linzhou-db: good point. moved back to spark.

client/src/main/scala/io/delta/sharing/client/filters/OpConverter.scala
linzhou-db: good point. moved back to spark.

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
chakankardb: Having DeltaSharingFileSystem in the client seems a bit odd.
Why do we need it in the client?

linzhou-db: It was discussed in the design doc actually. 
delta-sharing-client serves two main purposes: 1) DeltaSharingClient to handle http requests/responses 2) DeltaSharingFileSystem to handle reading data out of pre-signed urls.

chakankardb (approved):

PR #335, opened 2023-06-27
Delta Sharing Server to support return response in delta log format

linzhou-db:
Delta Sharing Server to support return response in delta log format, 2 main changes:
1. Added a set of `Delta*` actions in model.scala, used when the requested format is `delta`.
2. Delta sharing server able to parse `delta-sharing-capabilities` header in request and handle the requested format.

This is part 1 for issue https://github.com/delta-io/delta-sharing/issues/341

original prototype in https://github.com/delta-io/delta-sharing/pull/298.


server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -135,3 +135,126 @@ object Action {
...
+    expirationTimestamp: Long) extends DeltaAction {
+  require(path.nonEmpty)
+
+  override def wrap: DeltaSingleAction = DeltaSingleAction(add = this)
+}
+
+/**
+ * DeltaRemoveFile used in delta sharing protocol, copied from RemoveFile in delta.
+ *   Adding 4 fields: id/version/timestamp/expirationTimestamp.
+ *   Ignoring 1 field: tags.
```
prakharjain09, line 188: same as above.

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -135,3 +135,126 @@ object Action {
...
+ * breaking change to the protocol, clients should be tolerant of messages and
+ * fields that they do not understand.
+ */
+case class DeltaProtocol(minReaderVersion: Int) extends DeltaAction {
+  override def wrap: DeltaSingleAction = DeltaSingleAction(protocol = this)
+}
+
+/**
+ * DeltaAddFile used in delta sharing protocol, copied from AddFile in delta.
+ *   Adding 4 fields: id/version/timestamp/expirationTimestamp.
```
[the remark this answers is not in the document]
linzhou-db: These are delta sharing related fields.
Added more comment.

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -135,3 +135,126 @@ object Action {
...
+ *   Ignoring 1 field: tags.
```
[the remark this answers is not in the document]
linzhou-db: Question: are these fields all require minReaderVersion > 2?
If so we would skip them in the oss server change for now, since this change focuses on responseFormat=delta, we could add another PR to support DV, etc.

server/src/main/protobuf/protocol.proto
```diff
@@ -63,6 +63,9 @@ message QueryTableRequest {
     // Query all data change files until endingVersion, inclusive. Only used when startingVersion
     // is set, otherwise will be ignored.
     optional int64 endingVersion = 7;
+
+    // TODO: update this. The format of the response, supported formats: parquet, delta.
+    optional bool queryDeltaLog = 8;
```
chakankardb, line 68: Should this be an enum?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -268,12 +268,30 @@ class DeltaSharingService(serverConfig: ServerConfig) {
     HttpResponse.of(headers)
   }
 
+  private def getDeltaSharingCapabilitiesMap(headerString: String): Map[String, String] = {
+    if (headerString == null) {
+      return Map.empty[String, String]
+    }
+    headerString.toLowerCase().split(",").map { capbility =>
```
chakankardb, line 275: typo: capability?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -283,18 +301,22 @@ class DeltaSharingService(serverConfig: ServerConfig) {
       version = None,
       timestamp = None,
       startingVersion = None,
-      endingVersion = None
-    )
+      endingVersion = None,
+      responseFormat = capabilitiesMap.get(DELTA_SHARING_RESPONSE_FORMAT).getOrElse("parquet"))
```
chakankardb, line 305: Should we consolidate default values in some central config?
conf.getDeltaSharingResponseFormat()

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -135,3 +135,130 @@ object Action {
...
+
+/**
+ * DeltaAddFile used in delta sharing protocol, copied from AddFile in delta.
+ *   Adding 4 delta sharing related fields: id/version/timestamp/expirationTimestamp,
+ *       which will be redacted at the client side before writing to the delta log.
+ *   Ignoring 1 field: tags.
+ */
+case class DeltaAddFile(
+    path: String,
+    id: String,
```
chakankardb, line 170: Question: why do we want the server to generate this id? It adds load on the server. Can the client generate the id as needed?

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -135,3 +135,130 @@ object Action {
...
+    expirationTimestamp: Long) extends DeltaAction {
+  require(path.nonEmpty)
+
+  override def wrap: DeltaSingleAction = DeltaSingleAction(add = this)
+}
+
+/**
+ * DeltaRemoveFile used in delta sharing protocol, copied from RemoveFile in delta.
+ *   Adding 4 delta sharing related fields: id/version/timestamp/expirationTimestamp,
+ *       which will be redacted at the client side before writing to the delta log.
```
chakankardb, line 189: Consider adding a comment on why we are sending these.

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -135,3 +135,130 @@ object Action {
...
+    version: Long,
+    timestamp: Long,
+    expirationTimestamp: Long) extends DeltaAction {
+  override def wrap: DeltaSingleAction = DeltaSingleAction(remove = this)
+}
+
+/**
+ * DeltaAddCDCFile used in delta sharing protocol, copied from AddCDCFile in delta.
+ *   Adding 4 delta sharing related fields: id/version/timestamp/expirationTimestamp,
+ *       which will be redacted at the client side before writing to the delta log.
```
chakankardb, line 209: Consider expanding on why we are sending these?

This comment assumes a certain client implementation (the one we are working on currently).
We should perhaps generalize this comment:
   "If the client uses delta kernel, it should redact these fields as needed".

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -180,6 +181,160 @@ class DeltaSharedTable(
...
 
+  private def getResponseProtocol(p: Protocol, responseFormat: String): Object = {
+    if (responseFormat == "delta") {
+      model.DeltaProtocol(p.minReaderVersion).wrap
+    } else {
+      model.Protocol(p.minReaderVersion).wrap
+    }
+  }
+
+  private def getResponseMetadata(
```
chakankardb, line 192: consider adding headers for these functions.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -180,6 +181,160 @@ class DeltaSharedTable(
...
+        partitionValues = addFile.partitionValues,
+        size = addFile.size,
+        modificationTime = addFile.modificationTime,
+        dataChange = addFile.dataChange,
+        stats = addFile.stats,
+        version = version,
+        timestamp = timestamp
+      ).wrap
+    } else if (returnAddFileForCDF) {
+      model.AddFileForCDF(
```
chakankardb, line 252: Question: why do have AddFileForCDF which is not needed in delta? considering adding a comment.

server/src/main/scala/io/delta/sharing/server/model.scala
linzhou-db: good point, it assumes client side operatoins, but I think these will be part of the protocol, as client need these fields to construct the delta log correctly.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
linzhou-db: Because, 
1) we need to serialize these actions as json, and when we develop cdf, AddFile is used as "file", so we need another action to be "add". 
2) And "file" suppose not to contain fields such as version/timestamp.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
linzhou-db: Added response header, is that what you are asking for?

prakharjain09 (approved):

server/src/main/scala/io/delta/sharing/server/model.scala
chakankardb: I think it is fine to have it in the server. Just tested this and we generate these at 2M/sec.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
linzhou-db: added

server/src/main/scala/io/delta/sharing/server/model.scala
linzhou-db: thanks for checking the rate!

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -560,6 +568,10 @@ object DeltaSharingService {
     endingTimestamp.map(DeltaDataSource.CDF_END_TIMESTAMP_KEY -> _)).toMap
   }
 
+  private[server] def getResponseFormat(headerCapabilities: Map[String, String]): String = {
+    headerCapabilities.get(DELTA_SHARING_RESPONSE_FORMAT).getOrElse("parquet")
```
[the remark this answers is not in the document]
linzhou-db: updated.

PR #338, opened 2023-06-30
Support delta-sharing-capabilities header in DeltaSharingRestClient 

linzhou-db:
Support delta-sharing-capabilities header in DeltaSharingRestClient, if `responseFormat=delta`, then we don't parse the returned lines, just return them as a Seq[String]. In the tests, we are still trying to check the returned fields by directly matching the string.
For tests on exceptions, we use `DeltaSharingRestClientSuite` as the responseFormat shouldn't matter. For tests with successfully returned rpcs, we add a new `DeltaSharingRestClientDeltaSuite` to check the returned lines. 

Please only review the changes under `client/` in this PR, changes under `server/` should be reviewed with https://github.com/delta-io/delta-sharing/pull/335.

This is part 2 for issue https://github.com/delta-io/delta-sharing/issues/341

TODO: check whether what we should do with random capabilities header, error or ignore.


chakankardb:
@wchau Can you please also look at this change?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -205,7 +208,16 @@ class DeltaSharingRestClient(
     val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
     val target = getTargetUrl(
       s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/metadata")
-    val (version, lines) = getNDJson(target)
+    val (version, respondedFormat, lines) = getNDJson(target)
+    if (responseFormat != respondedFormat) {
+      logWarning(s"RespondedFormat($respondedFormat) is different from requested responseFormat(" +
```
chakankardb, line 213: Perhaps add a comment that this can happen if the DS server does not support the requested format.

Ideally, this should only happen if we request delta format but get back parquet. The reverse is not expected right?

linzhou-db: Right!

And this happens on the two recipient on uc use cases, where the delta sharing server is an open servers, and recipients are on databricks DBR, so recipients would any way use this code to ask for "delta" format, but may get parquet in response. And we should still be able to handle that case.

chakankardb (approved):

PR #339, opened 2023-06-30
Support delta-sharing-capabilities header in DeltaSharingClient

linzhou-db:
Support delta-sharing-capabilities header in DeltaSharingClient


linzhou-db:
not needed.

PR #342, opened 2023-07-10
[Branch-0.6] Double check sortedFetchedFiles and set previousOffset

linzhou-db:
1. Double check sortedFetchedFiles, to ensure we return the correct file set asked by the spark streaming engine.
2. set previousOffset even the returned DataFrame is empty.

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -229,8 +229,6 @@ case class DeltaSharingSource(
 
     val currentLatestVersion = getOrUpdateLatestTableVersion
     if (fromVersion > currentLatestVersion) {
-      logWarning(s"The asked version($fromVersion) is after currentLatestVersion(" +
-        s"$currentLatestVersion), for table(id:$tableId, name:${deltaLog.table.toString})")
       // If true, it means that there's no new data from the delta sharing server.
       return
     }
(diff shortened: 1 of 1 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -546,8 +565,11 @@ case class DeltaSharingSource(
 
     val fileActions = sortedFetchedFiles.takeWhile {
       case IndexedFile(version, index, _, _, _, _) =>
-        version < endOffset.tableVersion ||
-          (version == endOffset.tableVersion && index <= endOffset.index)
+        // Ensure (version, index) is in the range of
+        // [(startVersion, startIndex), (endVersion, endOffset)]
+        (version > startVersion || (version == startVersion && (index == -1 ||
+          index >= startIndex))) && (version < endOffset.tableVersion ||
+          (version == endOffset.tableVersion && index <= endOffset.index))
```
chakankardb, line 572: Question: Can these changes be unit tested?
Could we call the DeltaSharingSource apis from a test that can mimic various scenarios?

linzhou-db: It's hard to test because spark doesn't allow `getBatch` to be called alone: https://github.com/delta-io/delta-sharing/blob/main/spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala#L170

But I'll run integration tests in the notebook. I'll ping you specific tests before merging it.

Issue #351, opened 2023-07-21
Support Pagination in QueryTable and QueryTableChanges APIs


charlenelyu-db:
This is a proposal to add pagination support in [QueryTable API](https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#read-data-from-a-table) and [QueryTableChanges API](https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#read-change-data-feed-from-a-table).

## Motivation

Currently, we don't have a mechanism to restrict the number of files returned per query table request. When reading from tables that contain millions of files, the server may not be able to process such a substantial volume of files, leading to issues like timeouts or exceeding resource limits. Additionally, the client may struggle to handle large responses efficiently. This limitation becomes a performance bottleneck for the Delta Sharing service.

By introducing pagination in data access APIs, we can control the number of files returned in each API call. This will result in a more scalable Delta Sharing server and client solution.

## Protocol Change

We propose the following protocol changes:

### QueryTable

<table>
<tr>
<th>HTTP Request</th>
<th>Value</th>
</tr>
<tr>
<td>Method</td>
<td>

`POST`

</td>
</tr>
<tr>
<td>Headers</td>
<td>

`Authorization: Bearer {token}`
`Content-Type: application/json; charset=utf-8`

</td>
</tr>
<tr>
<td>URL</td>
<td>

`{prefix}/shares/{share}/schemas/{schema}/tables/{table}/query`

</td>
</tr>
<tr>
<td>URL Parameters</td>
<td>

No Change

</td>
</tr>
<tr>
<td>Body</td>
<td>

Add two optional fields:
- **maxFiles** (type: Int, optional): the maximum number of files to return in one page. If the number of available files is larger than `maxFiles`, the response will provide a `nextPageToken` that can be used to get the next page of results in subsequent requests. This is a hint for the server, and the server may not honor it. The server that supports pagination should return no more than this limit, but it can return fewer. The client should check `nextPageToken` in the response to determine if there are more available. Must be positive.
-  **pageToken** (type: String, optional): specifies the page token to use to retrieve the subsequent page. Set `pageToken` to the `nextPageToken` returned by a previous request to get the next page of results.

</td>
</tr>
</table>

Example:

`POST {prefix}/shares/vaccine_share/schemas/acme_vaccine_data/tables/vaccine_patients/query`

```json
{
  "maxFiles": 123,
  "pageToken": "..."
}
```

<details>
[...]

PR #354, opened 2023-07-24
Pagination for QueryTableChanges

charlenelyu-db:
- [Part 1](https://github.com/delta-io/delta-sharing/pull/352): Pagination for QueryTable at snapshot
  - [Part 2](https://github.com/delta-io/delta-sharing/pull/353): Pagination for QueryTable from starting_version
    - [Part 3](https://github.com/delta-io/delta-sharing/pull/354): Pagination for QueryTableChanges
      - [Part 4](https://github.com/delta-io/delta-sharing/pull/356): Client change [[files changed](https://github.com/delta-io/delta-sharing/pull/356/files/c7271be600f1e8645d53a7c12d2578cceda4deee..b05024e06af268ecb7c2f9bd541071be3eb319ee)]

---

Support pagination for QueryTableChanges in reference server.

This is part 3 for issue https://github.com/delta-io/delta-sharing/issues/351.

### DIFF
--- a/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
+++ b/server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
@@ -663,22 +663,43 @@ class DeltaSharedTable(
   def queryCDF(
       cdfOptions: Map[String, String],
       includeHistoricalMetadata: Boolean = false,
+      maxFiles: Option[Int],
+      pageToken: Option[String],
       responseFormat: String = DeltaSharedTable.RESPONSE_FORMAT_PARQUET
   ): (Long, Seq[Object]) = withClassLoader {
-    val actions = ListBuffer[Object]()
+    // Step 1: validate pageToken if it's specified
+    lazy val queryParamChecksum = computeChecksum(
+      QueryParamChecksum(
+        version = None,
+        timestamp = None,
+        startingVersion = cdfOptions.get(DeltaDataSource.CDF_START_VERSION_KEY).map(_.toLong),
+        startingTimestamp = cdfOptions.get(DeltaDataSource.CDF_START_TIMESTAMP_KEY),
+        endingVersion = cdfOptions.get(DeltaDataSource.CDF_END_VERSION_KEY).map(_.toLong),
+        endingTimestamp = cdfOptions.get(DeltaDataSource.CDF_END_TIMESTAMP_KEY),
+        predicateHints = Nil,
+        jsonPredicateHints = None,
+        limitHint = None,
+        includeHistoricalMetadata = Some(includeHistoricalMetadata)
+      )
+    )
+    val pageTokenOpt = pageToken.map(decodeAndValidatePageToken(_, queryParamChecksum))
 
-    // First: validate cdf options are greater than startVersion
+    // Step 2: validate cdfOptions
     val cdcReader = new DeltaSharingCDCReader(deltaLog, conf)
-    val latestVersion = tableVersion
+    // For subsequent page calls, instead of using the current latestVersion, use latestVersion in
+    // the pageToken (which is equal to the latestVersion when the first page call is received),
+    // in case the latestVersion changes after the first page call.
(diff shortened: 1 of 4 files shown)


server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -530,89 +537,169 @@ class DeltaSharedTable(
...
+          // Skip actions that are already processed in previous pages
+          indexedVersionActions =
+            indexedVersionActions.drop(pageTokenOpt.get.getStartingActionIndex)
+        }
+        indexedVersionActions.foreach {
+          case (a: AddFile, idx) if a.dataChange =>
+            // Return early if we already have enough files in the current page
+            if (pageSizeOpt.contains(numSignedFiles)) {
+              actions.append(tokenGenerator(v, idx))
+              return actions.toSeq
```
chakankardb, line 611: Question: It seems first we will generate all the changes, and then apply pagination? So pagination will not save any processing on the server side right? Is the same true for DS server in UC as well? @linzhou-db

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -530,89 +537,169 @@ class DeltaSharedTable(
...
+              )
+            }
+          case _ => ()
+        }
       }
+    // Return an empty `nextPageToken` object only when `maxFiles` is specified for
+    // backwards compatibility.
+    if (maxFilesOpt.isDefined) {
+      actions.append(model.NextPageToken(null).wrap)
     }
```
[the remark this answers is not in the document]
charlenelyu-db: Yeah I'm stacking PRs, sorry for the confusion. Please use the [files changed] link in the description to view incremental changes. I will clean things up before merging.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
```diff
@@ -622,71 +709,108 @@ class DeltaSharedTable(
...
-            cdcDataSpec.version,
-            cdcDataSpec.timestamp.getTime,
-            responseFormat
-          )
-        )
+    var numSignedFiles = 0
+    // We override (start, end) in subsequent page calls because:
+    // - Versions that are processed in previous pages can be skipped.
+    // - Versions that are committed after the first page call should be ignored, especially
+    //   when the endingVersion is not specified and resolved to latestVersion.
```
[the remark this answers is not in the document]
charlenelyu-db: By override I meant we should use (start, end) from the page token instead of the original request. I will rephrase it.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
charlenelyu-db: We're saving some process: the `start` will be equal to the `startingVersion` in the page token, so we won't unpack delta logs for versions that are already processed in previous pages.

However, we still generate all changes until the `endingVersion` each time. I guess this part can be optimized in UC. Two options I can think of: 
1. we could push down `actionListener` to `getChanges()`, so we read logs and process actions together and it'll return in the middle
2. we don't change the interface of `getChanges()`, we call it to unpack one version at a time and process version by version.

WDYT?

chakankardb: I think both options should work; the first one is similar to what we do for normal query, so may be preferable for consistency reason.
In the future, we will likely process json files in parallel as well (similar to checkpoint parquet).

Let's open a ticket for now to address this in UC.

chakankardb (approved):
Perhaps wait for Lin's review as well.

linzhou-db (approved):

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTableLoader.scala
linzhou-db: let's have a separate discussion about how to handle this in UC.

PR #361, opened 2023-07-31
Fix two build errors: use a static Cast init function and isinstance for type comparison in python 

linzhou-db:
Fix two build errors: 

1. use a static Cast init function to allow delta-sharing-spark work for multiple spark versions. 
2. use isinstance for type comparison in python.

### DIFF
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -183,7 +183,7 @@ class DeltaSharingReader:
 
         if for_cdf:
             # Add the change type col name to non cdc actions.
-            if type(action) != AddCdcFile:
+            if not isinstance(action, AddCdcFile):
                 pdf[DeltaSharingReader._change_type_col_name()] = action.get_change_type_col_value()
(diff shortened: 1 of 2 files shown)


python/delta_sharing/reader.py
```diff
@@ -183,7 +183,7 @@ def _to_pandas(
 
         if for_cdf:
             # Add the change type col name to non cdc actions.
-            if type(action) != AddCdcFile:
```
zsxwing, line 186: unrelated change?

zsxwing:
Could you also update the PR title and description?

python/delta_sharing/reader.py
linzhou-db: Build and Test /python (3.8) complains about this type comparison so I made the fix.

linzhou-db:
> Could you also update the PR title and description?

updated.

zsxwing (approved):
LGTM

PR #368, opened 2023-08-03
Add timestamp support in spark connector for V2 predicates.

chakankardb:
In this PR, we add timestamp support in V2 predicates.

### DIFF
--- a/spark/src/test/scala/io/delta/sharing/spark/filters/OpConverterSuite.scala
+++ b/spark/src/test/scala/io/delta/sharing/spark/filters/OpConverterSuite.scala
@@ -46,7 +46,8 @@ import org.apache.spark.sql.types.{
   FloatType => SqlFloatType,
   IntegerType => SqlIntegerType,
   LongType => SqlLongType,
-  StringType => SqlStringType
+  StringType => SqlStringType,
+  TimestampType => SqlTimestampType
 }
 
 class OpConverterSuite extends SparkFunSuite {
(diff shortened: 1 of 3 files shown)


spark/src/test/scala/io/delta/sharing/spark/filters/OpConverterSuite.scala
```diff
@@ -296,6 +297,25 @@ class OpConverterSuite extends SparkFunSuite {
...
+  test("Timestamp test") {
+    val formatter = java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME
+    val tsStr = "2023-06-07T04:27:03.234Z"
+    val tsMicros = java.time.OffsetDateTime.parse(tsStr, formatter).toInstant.toEpochMilli * 1000L
+    assert(tsMicros == 1686112023234000L)
+    val sqlColumn = SqlAttributeReference("ts", SqlTimestampType)()
+    val sqlLiteral = SqlLiteral(tsMicros, SqlLongType)
+    val sqlEq = SqlEqualTo(
+      SqlCast(sqlColumn, SqlTimestampType),
+      SqlCast(sqlLiteral, SqlTimestampType)
```
linzhou-db, line 309: question:in a real query, at which line of code are we applying this conversion? How do we know that we need to convert it to `SqlTimestampType`?

linzhou-db (approved):

spark/src/test/scala/io/delta/sharing/spark/filters/OpConverterSuite.scala
chakankardb: The OpConverter converts a SqlExpression tree into json predicate tree. The convertor gets this from spark, which constructs the sql expressions for data and partition predicates from query predicates. This test mimics how a sql timestamp would be part of such an expression.

PR #369, opened 2023-08-04
Splits DeltaSharingSourceSuite and DeltaSharingSourceLimitSuite

linzhou-db:
Splits DeltaSharingSourceSuite and DeltaSharingSourceLimitSuite so that each Suite could finish in shorter time, as the TestDeltaSharingServer is started for limited time, to avoid connection timeout issue.


spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceLimitSuite.scala
```diff
@@ -37,7 +37,7 @@ class DeltaSharingSourceLimitSuite extends QueryTest
   // VERSION 3: UPDATE 1 row, 1 remove file and 1 add file
   lazy val tablePath = testProfileFile.getCanonicalPath + "#share8.default.cdf_table_cdf_enabled"
 
-  val streamingTimeout = 30.seconds
+  val streamingTimeout = 20.seconds
```
chakankardb, line 40: This suite is also experiencing flakiness. Do we need to split this one as well?

chakankardb (approved):

PR #384, opened 2023-08-31
Fix bug where to_pandas doesn't properly handle the case when column names between Delta and parquet metadata are different in casing

andyl-db:
There's a bug where calling `to_pandas` on a partitioned table with differing Delta and parquet metadata, the resulting pandas dataframe would result in the columns being None. Specifically the difference in the metadata is the casing in the column names. 

ex: 
Delta columns: ["MARKET", "DATA", "branch"]
Parquet columns: ["market", "data", "branch"]

This PR fixes that issue by performing a case-insensitive lookup of the columns.

### DIFF
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -106,12 +106,18 @@ class DeltaSharingReader:
                 if left == 0:
                     break
 
-        return pd.concat(
+        merged = pd.concat(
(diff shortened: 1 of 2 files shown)


charlenelyu-db (approved):
LGTM

linzhou-db (approved):

Issue #383, opened 2023-08-30
Delta Sharing client needs to refresh urls for the correct table version


charlenelyu-db:
# Issue

Delta sharing spark connector could support long running (> 45 minutes) queries by making an additional RPC with the same parameters, to avoid the expiration of pre-signed urls. Though the refresh may fail under the following conditions:
- The spark query runs longer than the minimum url expiration; **AND**
- The query RPC is a [QueryTable RPC](https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#read-data-from-a-table) on the latest snapshot (`version`/`timestamp` is not specified); **AND**
- Between the initial RPC and the refresh RPC, there are new table versions which made deletion/updates to the table, causing some original files not returned in the refresh RPC.

# Solution

For table shared with history, this issue may be fixed by making the refresh RPC to query the previous version. But for table that are not shared with history, we need a mechanism to allow the refresh RPC to query a historical version and return the same set of files as the initial RPC.

We could introduce a refresh token to the `QueryTable` request. The refresh token will be returned from the initial RPC if requested, and the refresh RPC could specify the refresh token to retrieve the same set of files with refreshed urls. The protocol changes are as follow:

<table>
<tr>
<th>HTTP Request</th>
<th>Value</th>
</tr>
<tr>
<td>Method</td>
<td>

`POST`

</td>
</tr>
<tr>
<td>Headers</td>
<td>

`Authorization: Bearer {token}`
`Content-Type: application/json; charset=utf-8`

</td>
</tr>
<tr>
<td>URL</td>
<td>

`{prefix}/shares/{share}/schemas/{schema}/tables/{table}/query`

</td>
</tr>
<tr>
<td>URL Parameters</td>
<td>

No Change

</td>
</tr>
<tr>
<td>Body</td>
<td>

Add two optional fields:
- **includeRefreshToken** (type: Boolean, optional): whether or not to return a refresh token in the response. This should be set to `true` only when querying the latest snapshot (`version`/`timestamp` is not specified).
-  **refreshToken** (type: String, optional): specifies the refresh token to use to retrieve the same set of files with refreshed urls.

</td>
</tr>
</table>

Example:

`POST {prefix}/shares/vaccine_share/schemas/acme_vaccine_data/tables/vaccine_patients/query`

```json
{
  "includeRefreshToken": true,
  "refreshToken": "..."
}
```

<details>
<summary><b>200: The tables were successfully returned.</b></summary>

<table>
[...]

PR #387, opened 2023-09-01
CachedTableManager use refresh token to refresh urls

charlenelyu-db:
Use this [link](https://github.com/delta-io/delta-sharing/pull/387/files) to review incremental changes.

- [stack/refresh-token-server](https://github.com/delta-io/delta-sharing/pull/385)
  - [stack/refresh-token-client](https://github.com/delta-io/delta-sharing/pull/386)
    - [stack/refresh-token-spark](https://github.com/delta-io/delta-sharing/pull/387)

---

This pr modifies `CachedTableManager` to support refresh token:
- Change `refresher` func to take `refreshToken: Option[String]`, and return `(idToUrlMap, minUrlExpiration, newRefreshToken)`
- The `refreshToken` will be stored in the `CachedTable`
- When refresh is needed for a cached table, call `cachedTable.refresher(cachedTable.refreshToken)` to update urls.

Note that the above change only applies to latest snapshot queries. For other types of queries, `refreshToken` will always be `None`.

This is part 3 of https://github.com/delta-io/delta-sharing/issues/383.

### DIFF
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -35,15 +35,18 @@ import io.delta.sharing.client.DeltaSharingProfileProvider
  *             remove the cached table from our cache.
  * @param lastAccess When the table was accessed last time. We will remove old tables that are not
  *                   accessed after `expireAfterAccessMs` milliseconds.
- * @param refresher the function to generate a new file id to pre sign url map, as long as the new
- *                  expiration timestamp of the urls.
+ * @param refresher the function to generate a new file id to pre sign url map, with the new
+ *                  expiration timestamp of the urls and the new refresh token.
+ * @param refreshToken the optional refresh token that can be used by the refresher to retrieve
+ *                     the same set of files with refreshed urls.
  */
 class CachedTable(
     val expiration: Long,
     val idToUrl: Map[String, String],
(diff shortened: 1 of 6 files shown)


client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -35,15 +35,18 @@ import io.delta.sharing.client.DeltaSharingProfileProvider
...
+ * @param refreshToken the optional refresh token that can be used by the refresher to retrieve
+ *                     the same set of files with refreshed urls.
  */
 class CachedTable(
     val expiration: Long,
     val idToUrl: Map[String, String],
     val refs: Seq[WeakReference[AnyRef]],
     @volatile var lastAccess: Long,
-    val refresher: () => (Map[String, String], Option[Long]))
+    val refresher: Option[String] => (Map[String, String], Option[Long], Option[String]),
```
linzhou-db, line 48: I should do it in previous change: Instead of returning a 3 elements tuple, let's use a case class and return the class from refresher function, I don't have a good name, maybe `RefreshedResult`/`RefreshedUrlsAndMetadata`?

linzhou-db (approved):

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
charlenelyu-db: Done!

PR #388, opened 2023-09-02
Use case class ParsedDeltaSharingTablePath as return type for parsePath

linzhou-db:
Use case class ParsedDeltaSharingTablePath instead of tuple as return type for parsePath, this is part of https://github.com/delta-io/delta-sharing/issues/341

PR #378, opened 2023-08-21
Support versionAsOf and timestampAsOf for getMetadata in spark

linzhou-db:
Support versionAsOf and timestampAsOf for getMetadata in spark, updated existing unit tests.

Please only review changes under spark/ in this PR, all others should be covered by https://github.com/delta-io/delta-sharing/pull/377

PR #377, opened 2023-08-19
Support version and timestamp for DeltaSharingClient.getMetadata

linzhou-db:
Support version and timestamp for DeltaSharingClient.getMetadata, added new unit tests.

Please only review changes under client/ repo, server changes should be covered by https://github.com/delta-io/delta-sharing/pull/376

PR #392, opened 2023-09-11
Group fix comment in previous PRs

linzhou-db:
Group fix comment in previous PRs:
https://github.com/delta-io/delta-sharing/pull/377 
https://github.com/delta-io/delta-sharing/pull/378
https://github.com/delta-io/delta-sharing/pull/388


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -952,7 +952,13 @@ case class DeltaSharingSource(
     new AdmissionLimits().toReadLimit
   }
 
+  def debug(str: String): Unit = {
+    // scalastyle:off println
+    Console.println(s"----[linzhou]----SharingSource: $str")
```
chakankardb, line 957: should we turn the style back on after print?

Consider printing "DeltaSharingSourceDebug" instead on linzhou?

chakankardb (approved):

PR #396, opened 2023-09-14
Support json predicates in python/pandas connector.

chakankardb:
In this PR, we add support for json predicates in the python connector.
It also adds an integration test.

### DIFF
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -34,6 +34,7 @@ class DeltaSharingReader:
         rest_client: DataSharingRestClient,
         *,
         predicateHints: Optional[Sequence[str]] = None,
+        jsonPredicateHints: Optional[str] = None,
         limit: Optional[int] = None,
         version: Optional[int] = None,
         timestamp: Optional[str] = None,
@@ -45,6 +46,7 @@ class DeltaSharingReader:
             assert isinstance(predicateHints, Sequence)
             assert all(isinstance(predicateHint, str) for predicateHint in predicateHints)
         self._predicateHints = predicateHints
+        self._jsonPredicateHints = jsonPredicateHints
 
         if limit is not None:
             assert isinstance(limit, int) and limit >= 0, "'limit' must be a non-negative int"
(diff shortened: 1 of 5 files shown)


server/src/test/scala/io/delta/sharing/server/TestDeltaSharingServer.scala
```diff
@@ -37,6 +37,7 @@ object TestDeltaSharingServer {
     FileUtils.writeStringToFile(pidFile, pid)
     if (sys.env.get("AWS_ACCESS_KEY_ID").exists(_.length > 0)) {
       val serverConfigPath = TestResource.setupTestTables().getCanonicalPath
+      println("Abhijit loading server config from " + serverConfigPath)
```
linzhou-db, line 40: remove

server/src/universal/conf/delta-sharing-server.yaml.template
```diff
@@ -56,7 +56,7 @@ stalenessAcceptable: false
 # Whether to evaluate user provided `predicateHints`
 evaluatePredicateHints: false
 # Whether to evaluate user provided `jsonPredicateHints`
-evaluateJsonPredicateHints: false
+evaluateJsonPredicateHints: true
```
linzhou-db, line 59: To confirm this is intended.

python/dev/pytest
```diff
@@ -17,6 +17,8 @@
 #
 
 PYTHON_EXECUTABLE="${PYTHON_EXECUTABLE:-python3}"
+echo $PYTHON_EXECUTABLE
```
linzhou-db, line 20: keep or remove?

linzhou-db (approved):
I see the example of jsonPredicateHints here: https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#json-predicates-for-filtering

Wonder if we need comment in the json file to lead the user to the link above.

server/src/universal/conf/delta-sharing-server.yaml.template
chakankardb: Yes, this is needed to enable the predicates.

server/src/test/scala/io/delta/sharing/server/TestDeltaSharingServer.scala
chakankardb: Done

python/dev/pytest
chakankardb: Removed.

chakankardb:
> I see the example of jsonPredicateHints here: https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#json-predicates-for-filtering
> 
> Wonder if we need comment in the json file to lead the user to the link above.

Which json file should we add the comment in?

linzhou-db:
> > I see the example of jsonPredicateHints here: https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#json-predicates-for-filtering
> > Wonder if we need comment in the json file to lead the user to the link above.
> 
> Which json file should we add the comment in?

"python file" not "json file"... But I think it's ok to skip since looking for an example in the PROTOCOL page is straightforward.

PR #401, opened 2023-09-18
Improve error message on unrecognized file from server, and add two configs in 0.6

linzhou-db:
Improve error message when seeing unrecognized file from server: output the line instead of file.
And add two configs:

- "spark.delta.sharing.streaming.queryTableVersionIntervalSeconds", cannot be less than 30 seconds
- "spark.delta.sharing.ignoreUnparsedActions"

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
@@ -272,11 +272,14 @@ private[spark] class DeltaSharingRestClient(
     val addFiles = ArrayBuffer[AddFileForCDF]()
     val removeFiles = ArrayBuffer[RemoveFile]()
     val additionalMetadatas = ArrayBuffer[Metadata]()
-    lines.drop(2).map(line => JsonUtils.fromJson[SingleAction](line).unwrap).foreach{
-      case a: AddFileForCDF => addFiles.append(a)
-      case r: RemoveFile => removeFiles.append(r)
-      case m: Metadata => additionalMetadatas.append(m)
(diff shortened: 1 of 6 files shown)


spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
```diff
@@ -272,11 +279,14 @@ private[spark] class DeltaSharingRestClient(
...
-      case r: RemoveFile => removeFiles.append(r)
-      case m: Metadata => additionalMetadatas.append(m)
-      case f => throw new IllegalStateException(s"Unexpected File:${f}")
+    lines.drop(2).foreach { line =>
+      val action = JsonUtils.fromJson[SingleAction](line).unwrap
+      action match {
+        case a: AddFileForCDF => addFiles.append(a)
+        case r: RemoveFile => removeFiles.append(r)
+        case m: Metadata => additionalMetadatas.append(m)
+        case _ => throw new IllegalStateException(s"Unexpected Line:${line}")
```
chakankardb, line 288: If we want to be more lenient (since the client will consume from multiple servers), we can consider ignoring such lines (maybe based on some config: spark.delta.sharing.ignore_unrecognized_actions, default false)

chakankardb (approved):

spark/src/main/scala/io/delta/sharing/spark/DeltaSharingClient.scala
linzhou-db: added. 
Also added the rpc interval conf for streaming. 
Please take a look again.

charlenelyu-db (approved):

PR #403, opened 2023-09-19
Improve error message on unrecognized file from server, and add two configs

linzhou-db:
Improve error message when seeing unrecognized file from server: output the line instead of file.
And add two configs:

- "spark.delta.sharing.streaming.queryTableVersionIntervalSeconds", cannot be less than 30 seconds
- "spark.delta.sharing.ignoreUnparsedActions"

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -334,7 +334,7 @@ class DeltaSharingRestClient(
     checkProtocol(protocol)
     val metadata = JsonUtils.fromJson[SingleAction](lines(1)).metaData
     val files = ArrayBuffer[AddFile]()
-    files.drop(2).foreach { line =>
+    lines.drop(2).foreach { line =>
       val action = JsonUtils.fromJson[SingleAction](line)
       if (action.file != null) {
         files.append(action.file)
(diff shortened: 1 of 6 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -389,11 +397,16 @@ class DeltaSharingRestClient(
...
-      case f => throw new IllegalStateException(s"Unexpected File:${f}")
+    lines.drop(2).foreach { line =>
+      val action = JsonUtils.fromJson[SingleAction](line).unwrap
+      action match {
+        case a: AddFileForCDF => addFiles.append(a)
+        case r: RemoveFile => removeFiles.append(r)
+        case m: Metadata => additionalMetadatas.append(m)
+        case _ => if (!ConfUtils.ignoreUnparsedActions(SparkSession.active.sessionState.conf)) {
+          throw new IllegalStateException(s"Unexpected Line:${line}")
+        }
```
chakankardb, line 408: Perhaps log an error if we are ignoring?

linzhou-db: added.

charlenelyu-db (approved):

chakankardb (approved):

PR #420, opened 2023-10-07
Allow different response format

linzhou-db:
Allow different response format, will let the caller decide the behavior

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -239,16 +239,12 @@ class DeltaSharingRestClient(
     if (responseFormat != respondedFormat) {
       // This could only happen when the asked format is delta and the server doesn't support
       // the requested format.
-      logError(s"RespondedFormat($respondedFormat) is different from requested responseFormat(" +
+      logWarning(s"RespondedFormat($respondedFormat) is different from requested responseFormat(" +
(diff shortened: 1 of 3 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -239,16 +239,12 @@ class DeltaSharingRestClient(
     if (responseFormat != respondedFormat) {
       // This could only happen when the asked format is delta and the server doesn't support
       // the requested format.
-      logError(s"RespondedFormat($respondedFormat) is different from requested responseFormat(" +
+      logWarning(s"RespondedFormat($respondedFormat) is different from requested responseFormat(" +
         s"$responseFormat) for getMetadata.${table.share}.${table.schema}.${table.name}.")
-      throw new IllegalArgumentException("The responseFormat returned from the delta sharing " +
-        s"server doesn't match the requested responseFormat: respondedFormat($respondedFormat)" +
-        s" != requestedFormat($responseFormat)." +
-        s"This is likely because the server is not upgraded to support delta format sharing.")
```
chakankardb, line 247: Question: If the requestedFormat=parquet and responseFormat=delta, should we still error out? 
In the other case, we can fallback?

chakankardb (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: updated.

charlenelyu-db (approved):

PR #421, opened 2023-10-07
RemoteDeltaLog accept input Metadata

linzhou-db:
RemoteDeltaLog can be initialized with `initDeltaTableMetadata`.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
@@ -42,9 +42,10 @@ private[sharing] class RemoteDeltaLog(
   val table: DeltaSharingTable,
   val path: Path,
   val client: DeltaSharingClient,
-  inputDeltaTableMetadata: Option[DeltaTableMetadata] = None) {
+  initDeltaTableMetadata: Option[DeltaTableMetadata] = None) {
 
-  @volatile private var currentSnapshot: RemoteSnapshot = new RemoteSnapshot(path, client, table)
+  @volatile private var currentSnapshot: RemoteSnapshot = new RemoteSnapshot(
+    path, client, table, initDeltaTableMetadata = initDeltaTableMetadata)
 
   def snapshot(
       versionAsOf: Option[Long] = None,
@@ -52,7 +53,7 @@ private[sharing] class RemoteDeltaLog(
     if (versionAsOf.isEmpty && timestampAsOf.isEmpty) {
       currentSnapshot
     } else {
-      new RemoteSnapshot(path, client, table, versionAsOf, timestampAsOf, inputDeltaTableMetadata)
+      new RemoteSnapshot(path, client, table, versionAsOf, timestampAsOf, initDeltaTableMetadata)
     }
   }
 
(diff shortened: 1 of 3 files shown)


client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -41,17 +41,19 @@ import io.delta.sharing.spark.perf.DeltaSharingLimitPushDown
 private[sharing] class RemoteDeltaLog(
```
chakankardb, line 41: Consider adding comments on how the init metadata is used.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -165,7 +169,9 @@ class RemoteSnapshot(
   }
 
   private def getTableMetadata: (Metadata, Protocol, Long) = {
```
chakankardb, line 171: Consider adding some comments about initDeltaTableMetadata?

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -165,7 +169,9 @@ class RemoteSnapshot(
...
-    val tableMetadata = client.getMetadata(table, versionAsOf, timestampAsOf)
+    val tableMetadata = initDeltaTableMetadata.getOrElse(
+      client.getMetadata(table, versionAsOf, timestampAsOf)
+    )
```
chakankardb, line 174: Question: 
Should we cache the metadata as well?

We could have a local var that can be initialzed from initDeltaTableMetadata and update it if not set, which will avoid a refetch across multiple calls even if initDeltaTableMetadata is not specified.

linzhou-db: updated. using val as there's no re-assignment to it i the code.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
linzhou-db: added.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
linzhou-db: added

spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
```diff
@@ -505,4 +491,34 @@ class RemoteDeltaLogSuite extends SparkFunSuite with SharedSparkSession {
...
+
+  test("RemoteDeltaLog Initialized with metadata") {
+    val path = new Path("profileFile")
+    val table = Table(share = "share", schema = "schema", name = "table")
+    val client = new TestDeltaSharingClient()
+    client.clear()
+
+    def checkGetMetadataCalledOnce(versionAsOf: Option[Long] = None, nullable: Boolean): Unit = {
+      var deltaTableMetadata = client.getMetadata(table, versionAsOf, None)
+      assert(TestDeltaSharingClient.numMetadataCalled == 1)
```
chakankardb, line 503: Could we test that even if create the log without deltaTableMetadata specified, getMetadata will be only called once due to caching?

chakankardb (approved):

spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
linzhou-db: tested.

PR #426, opened 2023-10-20
Allow list of resopnseFormat

linzhou-db:
Allow list of responseFormat in the request http header, separated by comma(`responseFormat=parquet,delta`), but require only one decided responseformat in the response.
And throw error if the respondedFormat is not in the list of the request.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -227,9 +227,9 @@ class DeltaSharingRestClient(
   }
 
   /**
-   * Compare requestedFormat and respondedFormat, only error out when requested parquet but got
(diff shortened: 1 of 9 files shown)


charlenelyu-db (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -228,20 +231,13 @@ class DeltaSharingRestClient(
    * delta in response. The client allows backward compatibility: requested delta but got parquet
    * in response.
```
chakankardb, line 232: Update this comment?

chakankardb (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: updated.

PR #456, opened 2024-02-06
Adds more logging in DeltaSharingFileSystem

linzhou-db:
Adds more logging in DeltaSharingFileSystem, to understand # of calls on pre-signed urls.


client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
```diff
@@ -154,6 +163,7 @@ private[sharing] class DeltaSharingFileSystem extends FileSystem {
     throw new UnsupportedOperationException("mkdirs")
 
   override def getFileStatus(f: Path): FileStatus = {
+    logInfo(s"Checking delta sharing file status for path: $f.")
```
chakankardb, line 166: Question: Is it possible to guard these with some debug level?

linzhou-db: updated to logDebug

chakankardb (approved):

PR #471, opened 2024-04-09
Do not retry on ConnectionClosedException in branch 0.7

linzhou-db:
Do not retry on ConnectionClosedException, because it can be caused by invalid json returned from the delta sharing server.

The same PR as https://github.com/delta-io/delta-sharing/pull/467

### DIFF
--- a/spark/src/main/scala/io/delta/sharing/spark/util/RetryUtils.scala
+++ b/spark/src/main/scala/io/delta/sharing/spark/util/RetryUtils.scala
@@ -71,6 +71,9 @@ private[sharing] object RetryUtils extends Logging {
           false
         }
       case _: java.net.SocketTimeoutException => true
+      // do not retry on ConnectionClosedException because it can be caused by invalid json returned
+      // from the delta sharing server.
+      case _: org.apache.http.ConnectionClosedException => false
       case _: InterruptedException => false
       case _: InterruptedIOException => false
       case _: IOException => true


spark/src/main/scala/io/delta/sharing/spark/util/RetryUtils.scala
```diff
@@ -71,6 +71,9 @@ private[sharing] object RetryUtils extends Logging {
           false
         }
       case _: java.net.SocketTimeoutException => true
+      // do not retry on ConnectionClosedException because it can be caused by invalid json returned
+      // from the delta sharing server.
+      case _: org.apache.http.ConnectionClosedException => false
```
chakankardb, line 76: Are there other cases where this should be retried (for example, some networking error) which may be more common?

linzhou-db: that's a good question, I'm not sure about the side effect of this change. As it's the client side change, we don't have good logs/stats on what are all the cases that trigger this error. 

let me see if the current code works with a server side change, if not we can add more detailed check here but may still need it in some way.

charlenelyu-db (approved):

spark/src/main/scala/io/delta/sharing/spark/util/RetryUtils.scala
chakankardb: SG.

PR #478, opened 2024-04-21
Add queryID in userAgent for spark delta sharing queries

linzhou-db:
Add queryID in userAgent for spark delta sharing queries

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -173,6 +173,7 @@ case class DeltaSharingSource(
     if (lastGetVersionTimestamp == -1 ||
       (currentTimeMillis - lastGetVersionTimestamp) >= QUERY_TABLE_VERSION_INTERVAL_MILLIS) {
       val serverVersion = deltaLog.client.getTableVersion(deltaLog.table)
+      logInfo(s"Got table version $serverVersion from Delta Sharing Server.")
       if (serverVersion < 0) {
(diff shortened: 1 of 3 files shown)


client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -1124,7 +1125,10 @@ case class DeltaSharingSource(
...
       Some(v)
     } else if (options.startingTimestamp.isDefined) {
-      Some(deltaLog.client.getTableVersion(deltaLog.table, options.startingTimestamp))
+      val version = deltaLog.client.getTableVersion(deltaLog.table, options.startingTimestamp)
+      logInfo(s"Got table version $version for timestamp ${options.startingTimestamp} " +
+        s"from Delta Sharing Server.")
+      Some(version)
     } else {
       None
     }
```
chakankardb, line 1134: Can this change be unit tested?

linzhou-db: there's no real code change, just added a logInfo.

chakankardb: Aren't we changing the user agent header being sent by this client?
Can we test that the user agent as various fields? (perhaps a test was missing earlier).

linzhou-db: I see, taking a look.

linzhou-db: added

chakankardb (approved):

PR #487, opened 2024-05-17
use async api in delta sharing client for view query

jadewang-db:
Add async api support in delta-sharing client

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -163,7 +163,7 @@ class DeltaSharingRestClient(
     maxFilesPerReq: Int = 100000,
     enableAsyncQuery: Boolean = false,
     asyncQueryPollIntervalMillis: Long = 10000L,
-    asyncQueryMaxDuration: Long = Long.MaxValue
+    asyncQueryMaxDuration: Long = 600000L
   ) extends DeltaSharingClient with Logging {
 
   logError(s"DeltaSharingRestClient with enableAsyncQuery $enableAsyncQuery")
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -21,7 +21,7 @@ import java.net.{URL, URLEncoder}
 import java.nio.charset.StandardCharsets.UTF_8
 import java.sql.Timestamp
 import java.time.LocalDateTime
-import java.time.format.DateTimeFormatter.ISO_DATE_TIME
+import java.time.format.DateTimeFormatter.{ISO_DATE, ISO_DATE_TIME}
 import java.util.UUID
 
 import scala.collection.mutable.{ArrayBuffer, ListBuffer}
@@ -499,8 +499,13 @@ class DeltaSharingRestClient(
     val start = System.currentTimeMillis()
     var numPages = 1
 
-    val (version, respondedFormat, lines, queryIdOpt)
-    = getNDJsonWithAsync(table, targetUrl, request)
+    val (version, respondedFormat, lines, queryIdOpt) = if (enableAsyncQuery) {
+      getNDJsonWithAsync(table, targetUrl, request)
+    } else {
+      val (version, respondedFormat, lines) = getNDJson(targetUrl, request)
+      (version, respondedFormat, lines, None)
+    }
+
     var (filteredLines, endStreamAction) = maybeExtractEndStreamAction(lines)
     if (endStreamAction.isEmpty) {
       logWarning("EndStreamAction is not returned in the response for paginated query.")
@@ -821,11 +826,6 @@ class DeltaSharingRestClient(
         throw new IllegalStateException("QueryId is not returned in the response.")
       }
 
-      if (queryStatus.status != QUERY_PENDING_TRUE) {
-        throw new IllegalStateException("Unexpected query status: " + queryStatus.status)
-      }
-
-      logInfo(s"Query is timed out. QueryId: ${queryStatus.queryId}")
       (lines.drop(1), Some(queryStatus.queryId), true)
     }
   }
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -361,8 +361,7 @@ class DeltaSharingRestClient(
       pageToken = None,
       includeRefreshToken = Some(includeRefreshToken),
       refreshToken = refreshToken,
-      idempotencyKey = idempotencyKey,
-      requestedColumns = Seq.empty[String]
+      idempotencyKey = idempotencyKey
     )
 
     val updatedRequest = if (queryTablePaginationEnabled) {
@@ -434,8 +433,7 @@ class DeltaSharingRestClient(
       pageToken = None,
       includeRefreshToken = None,
       refreshToken = None,
-      idempotencyKey = None,
-      requestedColumns = Seq.empty[String]
+      idempotencyKey = None
     )
     val (version, respondedFormat, lines) = if (queryTablePaginationEnabled) {
       logInfo(
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -16,7 +16,7 @@
 
 package io.delta.sharing.client
 
-import java.io.{BufferedReader, InputStream, InputStreamReader}
+import java.io.{BufferedReader, InputStreamReader}
 import java.net.{URL, URLEncoder}
 import java.nio.charset.StandardCharsets.UTF_8
 import java.sql.Timestamp
@@ -24,7 +24,6 @@ import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter.ISO_DATE_TIME
 import java.util.UUID
 
-import scala.collection.JavaConverters._
 import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 
 import org.apache.commons.io.IOUtils
@@ -33,7 +32,7 @@ import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.util.VersionInfo
 import org.apache.http.{HttpHeaders, HttpHost, HttpStatus}
 import org.apache.http.client.config.RequestConfig
-import org.apache.http.client.methods.{HttpGet, HttpHead, HttpPost, HttpRequestBase}
+import org.apache.http.client.methods.{HttpGet, HttpPost, HttpRequestBase}
 import org.apache.http.client.protocol.HttpClientContext
 import org.apache.http.conn.ssl.{SSLConnectionSocketFactory, SSLContextBuilder, TrustSelfSignedStrategy}
 import org.apache.http.entity.StringEntity
@@ -88,6 +87,16 @@ private[sharing] trait PaginationResponse {
   def nextPageToken: Option[String]
 }
 
+private[sharing] trait PaginationRequest {
+  def maxFiles: Option[Int]
+  def pageToken: Option[String]
+
+  def clone(
+     maxFiles: Option[Int],
+     pageToken: Option[String],
+     unsetRefreshToken: Boolean = false): PaginationRequest
+}
+
 private[sharing] case class QueryTableRequest(
   predicateHints: Seq[String],
   limitHint: Option[Long],
@@ -99,8 +108,39 @@ private[sharing] case class QueryTableRequest(
   maxFiles: Option[Int],
   pageToken: Option[String],
   includeRefreshToken: Option[Boolean],
-  refreshToken: Option[String]
-)
+  refreshToken: Option[String],
+  idempotencyKey: Option[String],
+  requestedColumns: Seq[String]
+) extends PaginationRequest {
+  override def clone(
+        maxFiles: Option[Int],
+        pageToken: Option[String],
+        unsetRefreshToken: Boolean = false): PaginationRequest = {
+    if (unsetRefreshToken) {
+      this.copy(
+        maxFiles = maxFiles,
+        pageToken = pageToken,
+        refreshToken = None,
+        includeRefreshToken = None)
+    }
+    else {
+      this.copy(maxFiles = maxFiles, pageToken = pageToken)
+    }
+  }
+}
+
+private[sharing] case class GetQueryInfoRequest(
+  queryId: String,
+  maxFiles: Option[Int],
+  pageToken: Option[String]
+  ) extends PaginationRequest {
+  override def clone(
+      maxFiles: Option[Int],
+      pageToken: Option[String],
+      unsetRefreshToken: Boolean = false): PaginationRequest = {
+    this.copy(maxFiles = maxFiles, pageToken = pageToken)
+  }
+}
 
 private[sharing] case class ListSharesResponse(
     items: Seq[Share],
@@ -112,16 +152,19 @@ private[sharing] case class ListAllTablesResponse(
 
 /** A REST client to fetch Delta metadata from remote server. */
 class DeltaSharingRestClient(
-    profileProvider: DeltaSharingProfileProvider,
-    timeoutInSeconds: Int = 120,
-    numRetries: Int = 10,
-    maxRetryDuration: Long = Long.MaxValue,
-    sslTrustAll: Boolean = false,
-    forStreaming: Boolean = false,
-    responseFormat: String = DeltaSharingRestClient.RESPONSE_FORMAT_PARQUET,
-    readerFeatures: String = "",
-    queryTablePaginationEnabled: Boolean = false,
-    maxFilesPerReq: Int = 100000
+      profileProvider: DeltaSharingProfileProvider,
+      timeoutInSeconds: Int = 120,
+      numRetries: Int = 10,
+      maxRetryDuration: Long = Long.MaxValue,
+      sslTrustAll: Boolean = false,
+      forStreaming: Boolean = false,
+      responseFormat: String = DeltaSharingRestClient.RESPONSE_FORMAT_PARQUET,
+      readerFeatures: String = "",
+      queryTablePaginationEnabled: Boolean = false,
+      maxFilesPerReq: Int = 100000,
+      enableAsyncQuery: Boolean = false,
+      asyncQueryPollIntervalMillis: Long = 1000L,
+      asyncQueryMaxDuration: Long = Long.MaxValue
   ) extends DeltaSharingClient with Logging {
 
   import DeltaSharingRestClient._
@@ -264,6 +307,7 @@ class DeltaSharingRestClient(
       table = s"${table.share}.${table.schema}.${table.name}"
     )
 
+
     if (respondedFormat == RESPONSE_FORMAT_DELTA) {
       return DeltaTableMetadata(version, lines = lines, respondedFormat = respondedFormat)
     }
@@ -300,6 +344,13 @@ class DeltaSharingRestClient(
     val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
     val target = getTargetUrl(
       s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/query")
+
+    val idempotencyKey = if (enableAsyncQuery) {
+      Some(UUID.randomUUID().toString)
+    } else {
+      None
+    }
+
     val request: QueryTableRequest = QueryTableRequest(
       predicateHints = predicates,
       limitHint = limit,
@@ -311,16 +362,18 @@ class DeltaSharingRestClient(
       maxFiles = None,
       pageToken = None,
       includeRefreshToken = Some(includeRefreshToken),
-      refreshToken = refreshToken
+      refreshToken = refreshToken,
+      idempotencyKey = idempotencyKey,
+      requestedColumns = Seq.empty[String]
     )
     val (version, respondedFormat, lines, refreshTokenOpt) = if (queryTablePaginationEnabled) {
       logInfo(
         s"Making paginated queryTable requests for table " +
         s"${table.share}.${table.schema}.${table.name} with maxFiles=$maxFilesPerReq"
       )
-      getFilesByPage(target, request)
+      getFilesByPage(table, target, request)
     } else {
-      val (version, respondedFormat, lines) = getNDJson(target, request)
+      val (version, respondedFormat, lines, _) = getNDJsonWithAsync(table, target, request)
       val (filteredLines, endStreamAction) = maybeExtractEndStreamAction(lines)
       val refreshTokenOpt = endStreamAction.flatMap { e =>
         Option(e.refreshToken).flatMap { token =>
@@ -391,14 +444,16 @@ class DeltaSharingRestClient(
       maxFiles = None,
       pageToken = None,
       includeRefreshToken = None,
-      refreshToken = None
+      refreshToken = None,
+      idempotencyKey = None,
+      requestedColumns = Seq.empty[String]
     )
     val (version, respondedFormat, lines) = if (queryTablePaginationEnabled) {
       logInfo(
         s"Making paginated queryTable from version $startingVersion requests for table " +
         s"${table.share}.${table.schema}.${table.name} with maxFiles=$maxFilesPerReq"
       )
-      val (version, respondedFormat, lines, _) = getFilesByPage(target, request)
+      val (version, respondedFormat, lines, _) = getFilesByPage(table, target, request)
       (version, respondedFormat, lines)
     } else {
       getNDJson(target, request)
@@ -442,6 +497,7 @@ class DeltaSharingRestClient(
   // Send paginated queryTable requests. Loop internally to fetch and concatenate all pages,
   // then return (version, respondedFormat, actions, refreshToken) tuple.
   private def getFilesByPage(
+      table: Table,
       targetUrl: String,
       request: QueryTableRequest): (Long, String, Seq[String], Option[String]) = {
     val allLines = ArrayBuffer[String]()
@@ -449,12 +505,17 @@ class DeltaSharingRestClient(
     var numPages = 1
 
     // Fetch first page
-    var updatedRequest = request.copy(maxFiles = Some(maxFilesPerReq))
-    val (version, respondedFormat, lines) = getNDJson(targetUrl, updatedRequest)
+    val updatedRequest = request.copy(
+      maxFiles = Some(maxFilesPerReq),
+      pageToken = request.pageToken)
+
+    val (version, respondedFormat, lines, queryIdOpt)
+    = getNDJsonWithAsync(table, targetUrl, updatedRequest)
     var (filteredLines, endStreamAction) = maybeExtractEndStreamAction(lines)
     if (endStreamAction.isEmpty) {
       logWarning("EndStreamAction is not returned in the response for paginated query.")
     }
+
     val protocol = filteredLines(0)
     val metadata = filteredLines(1)
     // Extract refresh token if available
@@ -473,16 +534,32 @@ class DeltaSharingRestClient(
       endStreamAction.get.nextPageToken != null &&
       endStreamAction.get.nextPageToken.nonEmpty) {
       numPages += 1
-      updatedRequest = updatedRequest.copy(
-        pageToken = Some(endStreamAction.get.nextPageToken),
-        // Unset includeRefreshToken and refreshToken because they can only be used in
-        // the first page request.
-        includeRefreshToken = None,
-        refreshToken = None
-      )
+
+      val (pagingRequest, nextPageUrl) = if (queryIdOpt.isDefined) {
+        (
+          GetQueryInfoRequest(
+            queryId = queryIdOpt.get,
+            maxFiles = Some(maxFilesPerReq),
+            pageToken = Some(endStreamAction.get.nextPageToken)),
+          getQueryInfoTargetUrl(table, queryIdOpt.get)
+        )
+      } else {
+        (
+          updatedRequest.clone(
+            maxFiles = Some(maxFilesPerReq),
+            pageToken = Some(endStreamAction.get.nextPageToken),
+            // Unset includeRefreshToken and refreshToken
+            // because they can only be used in
+            // the first page request.
+            unsetRefreshToken = true
+          ),
+          targetUrl
+        )
+      }
+
       val res = fetchNextPageFiles(
-        targetUrl = targetUrl,
-        requestBody = Some(updatedRequest),
+        targetUrl = nextPageUrl,
+        requestBody = Some(pagingRequest),
         expectedVersion = version,
         expectedRespondedFormat = respondedFormat,
         expectedProtocol = protocol,
@@ -629,7 +706,7 @@ class DeltaSharingRestClient(
   // if it's not returned in the response.
   private def fetchNextPageFiles(
       targetUrl: String,
-      requestBody: Option[QueryTableRequest],
+      requestBody: Option[PaginationRequest],
       expectedVersion: Long,
       expectedRespondedFormat: String,
       expectedProtocol: String,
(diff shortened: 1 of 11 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -88,6 +87,16 @@ private[sharing] trait PaginationResponse {
   def nextPageToken: Option[String]
 }
 
+private[sharing] trait PaginationRequest {
```
linzhou-db, line 90: Why extracting this trait out?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -311,16 +362,18 @@ class DeltaSharingRestClient(
...
     val (version, respondedFormat, lines, refreshTokenOpt) = if (queryTablePaginationEnabled) {
       logInfo(
         s"Making paginated queryTable requests for table " +
         s"${table.share}.${table.schema}.${table.name} with maxFiles=$maxFilesPerReq"
       )
-      getFilesByPage(target, request)
+      getFilesByPage(table, target, request)
     } else {
-      val (version, respondedFormat, lines) = getNDJson(target, request)
+      val (version, respondedFormat, lines, _) = getNDJsonWithAsync(table, target, request)
```
linzhou-db, line 376: Will this affect the existing queries?

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -111,6 +123,44 @@ object ConfUtils {
     maxDur
   }
 
+  def asyncQueryPollIntervalMillis(conf: Configuration): Long = {
```
linzhou-db, line 126: Why do we implement each twice? Are they needed?

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientDeltaSuite.scala
```diff
@@ -93,7 +93,7 @@ class DeltaSharingRestClientDeltaSuite extends DeltaSharingIntegrationTest {
     }
   }
 
-  integrationTest("getFiles") {
+  integrationTest("getFiles include async") {
```
linzhou-db, line 96: why this change?

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
```diff
@@ -293,7 +293,68 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
...
+          size = 573,
+          stats =
+            """{"numRecords":1,"minValues":{"eventTime":"2021-04-28T23:33:48.719Z"},"maxValues":{"eventTime":"2021-04-28T23:33:48.719Z"},"nullCount":{"eventTime":0}}"""
+        )
+      )
+      assert(expectedFiles == tableFiles.files.toList)
+      assert(tableFiles.files(0).expirationTimestamp > System.currentTimeMillis())
+    }
+
+    Seq(false, true).foreach { paginationEnabled => {}
```
linzhou-db, line 333: remove the {} at the end of line?

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
```diff
@@ -340,40 +401,47 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
...
-              limit = None,
-              versionAsOf = None,
-              timestampAsOf = None,
-              jsonPredicateHints = None,
-              refreshToken = tableFiles.refreshToken
-            )
-          verifyTableFiles(refreshedTableFiles)
-        } finally {
-          client.close()
+      Seq(false).foreach { paginationEnabled =>
```
linzhou-db, line 404: do we need to update L404 and L405?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -325,6 +325,45 @@ class DeltaSharingService(serverConfig: ServerConfig) {
     streamingOutput(Some(queryResult.version), queryResult.responseFormat, queryResult.actions)
   }
 
+
+  @Post("/shares/{share}/schemas/{schema}/tables/{table}/queries/{queryId}")
+  @ConsumesJson
+  def getQueryStatus(
+                 req: HttpRequest,
```
linzhou-db, line 332: fix the indent.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -396,28 +435,47 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+      val queryId = s"${share}_${schema}_${table}"
+
+      streamingOutput(
+        Some(0),
+        "parquet",
+        Seq(
+        SingleAction(queryStatus = QueryStatus(queryId, "pending"))
+      ))
+    }
+    else {
```
linzhou-db, line 455: nit: put in the same line as the above line?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -396,28 +435,47 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+        )
+      }
+
+      val queryId = s"${share}_${schema}_${table}"
+
+      streamingOutput(
+        Some(0),
+        "parquet",
+        Seq(
+        SingleAction(queryStatus = QueryStatus(queryId, "pending"))
```
linzhou-db, line 452: So this is always returning "pending"?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +796,77 @@ class DeltaSharingRestClient(
...
+    val shareName = URLEncoder.encode(table.share, "UTF-8")
+    val schemaName = URLEncoder.encode(table.schema, "UTF-8")
+    val tableName = URLEncoder.encode(table.name, "UTF-8")
+    val encodedQueryId = URLEncoder.encode(queryId, "UTF-8")
+    val target = getTargetUrl(
+      s"/shares/$shareName/schemas/$schemaName/tables/$tableName/queries/$encodedQueryId")
+    target
+  }
+
+  private def getQueryInfo(
```
linzhou-db, line 809: Is this getQueryInfo or queryView? or asyncQueryTable?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -396,28 +435,47 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
-      request.refreshToken,
-      responseFormatSet = responseFormatSet)
-    if (queryResult.version < tableConfig.startVersion) {
-      throw new DeltaSharingIllegalArgumentException(
-        s"You can only query table data since version ${tableConfig.startVersion}."
-      )
+
+    if(getAsyncQuery(capabilitiesMap)) {
+      if (!request.idempotencyKey.isDefined) {
+        throw new DeltaSharingIllegalArgumentException(
```
linzhou-db, line 441: could you add a test for this?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: this is to make GetQueryInfo api to be able to support the pagination when retrieving the query result. so extract some common properties.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -99,8 +108,39 @@ private[sharing] case class QueryTableRequest(
...
+        unsetRefreshToken: Boolean = false): PaginationRequest = {
+    if (unsetRefreshToken) {
+      this.copy(
+        maxFiles = maxFiles,
+        pageToken = pageToken,
+        refreshToken = None,
+        includeRefreshToken = None)
+    }
+    else {
+      this.copy(maxFiles = maxFiles, pageToken = pageToken)
```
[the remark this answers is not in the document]
jadewang-db: yes, copy will preserve all properties not being override in the parameters?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: this is designed to be compatible with existing sync query path, and existing test cases succeed.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: it's getQueryInfo, this where we pool the query status using the new getQueryInfo api.

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
jadewang-db: I am following the existing pattern, looks like we want to support 2 different type of config system.

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
jadewang-db: fixed, thank you

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
jadewang-db: done

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
jadewang-db: for this demo service, we already return pending if user start async.

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientDeltaSuite.scala
jadewang-db: reverted.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +795,77 @@ class DeltaSharingRestClient(
...
+      if (System.currentTimeMillis() - startTime > asyncQueryMaxDuration) {
+        throw new IllegalStateException("Query is still pending after " +
+          s"${asyncQueryMaxDuration} ms. Please try again later.")
+      }
+    Thread.sleep(asyncQueryPollIntervalMillis)
+
+      val queryId = queryIdOpt.get
+      val (currentVersion, currentRespondedFormat, currentLines)
+        = getQueryInfo(table, queryId, request.maxFiles, request.pageToken)
+      val (newLines, _, newQueryPending) = checkQueryPending(currentLines)
```
chakankardb, line 857: Question: will the server send the queryId back?
We should perhaps check that if it does, then the queryId must match.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -910,7 +1057,13 @@ class DeltaSharingRestClient(
     if (responseFormatSet.contains(RESPONSE_FORMAT_DELTA) && readerFeatures.nonEmpty) {
       capabilities = capabilities :+ s"$READER_FEATURES=$readerFeatures"
     }
-    capabilities.mkString(DELTA_SHARING_CAPABILITIES_DELIMITER)
+
+    if(enableAsyncQuery) {
```
chakankardb, line 1061: nit: space after if ?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +795,77 @@ class DeltaSharingRestClient(
...
+
+    // Check if the query is still pending
+    var (lines, queryIdOpt, queryPending) = checkQueryPending(initialLines)
+
+    var version = initialVersion
+    var respondedFormat = initialRespondedFormat
+
+    val startTime = System.currentTimeMillis()
+    // Loop while the query is pending
+    while (queryPending) {
```
chakankardb, line 847: I think we are assuming that the query is either pending, or it has succeeded.
Could the server send back an error in the query (for example if MO errors out, or is cancelled due to large size)?

We should perhaps consider: [pending, finished, error, cancelled] as possible response types. These would need to be added to the protocol.

client/src/main/scala/io/delta/sharing/client/model.scala
```diff
@@ -93,6 +94,13 @@ private[sharing] case class SingleAction(
   }
 }
 
+case class QueryStatus(
+      queryId: String = null,
+      status: String = null
```
chakankardb, line 99: status should ideally be an enum?
We should perhaps add an error message for server to be able to send errors back to the client.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +795,77 @@ class DeltaSharingRestClient(
...
+        (lines, None, false)
+      } else {
+      (lines.drop(1), Some(queryStatus.queryId), queryStatus.status == QUERY_PENDING_TRUE)
+    }
+  }
+
+  private def getNDJsonWithAsync(
+                table: Table,
+                target: String,
+                request: QueryTableRequest): (Long, String, Seq[String], Option[String]) = {
```
chakankardb, line 835: Consider documenting this method, specifically what is being returned.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -112,16 +152,19 @@ private[sharing] case class ListAllTablesResponse(
...
-    numRetries: Int = 10,
-    maxRetryDuration: Long = Long.MaxValue,
-    sslTrustAll: Boolean = false,
-    forStreaming: Boolean = false,
-    responseFormat: String = DeltaSharingRestClient.RESPONSE_FORMAT_PARQUET,
-    readerFeatures: String = "",
-    queryTablePaginationEnabled: Boolean = false,
-    maxFilesPerReq: Int = 100000
+      profileProvider: DeltaSharingProfileProvider,
+      timeoutInSeconds: Int = 120,
```
chakankardb, line 156: nit: indent seems off?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
chakankardb: I think it does make the code a bit risky; if there is a bug in getNDJsonWithAsync, it will impact existing queries.
Should we do an if (enableAsync) switch here instead?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -442,19 +496,25 @@ class DeltaSharingRestClient(
...
 
     // Fetch first page
-    var updatedRequest = request.copy(maxFiles = Some(maxFilesPerReq))
-    val (version, respondedFormat, lines) = getNDJson(targetUrl, updatedRequest)
+    val updatedRequest = request.copy(
+      maxFiles = Some(maxFilesPerReq),
+      pageToken = request.pageToken)
+
+    val (version, respondedFormat, lines, queryIdOpt)
+    = getNDJsonWithAsync(table, targetUrl, updatedRequest)
```
chakankardb, line 512: Same question. Should we do an if (enableAsync) here as well?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: the error is sent back through http error code and expcetion, there will be no query status for error

jadewang-db: let's know your thoughts

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -88,6 +87,24 @@ private[sharing] trait PaginationResponse {
   def nextPageToken: Option[String]
 }
 
+/*
+abstract the requirement into a trait so that we can
```
linzhou-db, line 91: update the comment? add comma at proper places?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -442,19 +504,25 @@ class DeltaSharingRestClient(
...
     val allLines = ArrayBuffer[String]()
     val start = System.currentTimeMillis()
     var numPages = 1
 
     // Fetch first page
-    var updatedRequest = request.copy(maxFiles = Some(maxFilesPerReq))
-    val (version, respondedFormat, lines) = getNDJson(targetUrl, updatedRequest)
+    val updatedRequest = request.copy(
+      maxFiles = Some(maxFilesPerReq),
+      pageToken = request.pageToken)
```
linzhou-db, line 517: nit: why do we need this override?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: +1

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -112,16 +160,19 @@ private[sharing] case class ListAllTablesResponse(
...
+      timeoutInSeconds: Int = 120,
+      numRetries: Int = 10,
+      maxRetryDuration: Long = Long.MaxValue,
+      sslTrustAll: Boolean = false,
+      forStreaming: Boolean = false,
+      responseFormat: String = DeltaSharingRestClient.RESPONSE_FORMAT_PARQUET,
+      readerFeatures: String = "",
+      queryTablePaginationEnabled: Boolean = false,
+      maxFilesPerReq: Int = 100000,
+      enableAsyncQuery: Boolean = false,
```
linzhou-db, line 173: Is it possible to have cleaner separate on the code path to handle sync v.s. async queries? 

To make it easier to read and make sure the sync queries are not affected.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +803,107 @@ class DeltaSharingRestClient(
...
+    var (lines, queryIdOpt, queryPending) = checkQueryPending(initialLines)
+
+    var version = initialVersion
+    var respondedFormat = initialRespondedFormat
+
+    val startTime = System.currentTimeMillis()
+    // Loop while the query is pending
+    while (queryPending) {
+      if (System.currentTimeMillis() - startTime > asyncQueryMaxDuration) {
+        throw new IllegalStateException("Query is still pending after " +
```
andyl-db, line 881: nit: say timed out?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +803,107 @@ class DeltaSharingRestClient(
...
+  * be a query status object if the query is still pending.
+  * The method return (queryResultLines, queryId, queryPending) tuple. If the queryPending
+  * is false it means the query is finished and the queryResultLines contains the result.
+   */
+  private def checkQueryPending(lines: Seq[String]): (Seq[String], Option[String], Boolean) = {
+    val queryStatus = JsonUtils.fromJson[SingleAction](lines(0)).queryStatus
+
+    if(queryStatus == null) {
+        (lines, None, false)
+      } else {
```
andyl-db, line 841: nit: indentation

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
andyl-db: I agree that this seems a bit risky to do

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
andyl-db: +1

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -99,8 +116,39 @@ private[sharing] case class QueryTableRequest(
   maxFiles: Option[Int],
   pageToken: Option[String],
   includeRefreshToken: Option[Boolean],
-  refreshToken: Option[String]
-)
+  refreshToken: Option[String],
+  idempotencyKey: Option[String],
+  requestedColumns: Seq[String]
+) extends PaginationRequest {
+  override def clone(
```
andyl-db, line 123: This feels confusing...can you not just set refreshToken = None in the code instead of having this `unsetRefreshToken` parameter?

andyl-db: Why don't we get rid of this clone method and directly use copy in the function?

andyl-db (changes requested):
Mostly nits, but my biggest take away is similar to what Lin said about separating out the logic for async and sync queries for both readability and maintainability.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
chakankardb: That should work too; we will need to ensure we return a non-retriable error code and document it in the protocol so that clients do not keep retrying.

client/src/main/scala/io/delta/sharing/client/model.scala
chakankardb: Not needed if  the rpc itself returns error.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -99,8 +108,39 @@ private[sharing] case class QueryTableRequest(
...
+        refreshToken = None,
+        includeRefreshToken = None)
+    }
+    else {
+      this.copy(maxFiles = maxFiles, pageToken = pageToken)
+    }
+  }
+}
+
+private[sharing] case class GetQueryInfoRequest(
```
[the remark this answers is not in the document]
chakankardb: ?

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
```diff
@@ -293,7 +293,95 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
...
+          predicates = Nil,
+          limit = None,
+          versionAsOf = None,
+          timestampAsOf = None,
+          jsonPredicateHints = None,
+          refreshToken = None
+        )
+      verifyTableFiles(tableFiles)
+    }
+    }
```
chakankardb, line 381: nit: indent seems off.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -325,6 +325,49 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+  @ConsumesJson
+  def getQueryStatus(
+     req: HttpRequest,
+     @Param("share") share: String,
+     @Param("schema") schema: String,
+     @Param("table") table: String,
+     @Param("queryId") queryId: String,
+     request: GetQueryInfoRequest): HttpResponse = processRequest {
+
+    if (table == "table1") {
```
chakankardb, line 339: consider using a more descriptive name: "tableWithAsyncQueryError" ?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -325,6 +325,49 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+      false,
+      None,
+      responseFormatSet = responseFormatSet)
+    if (queryResult.version < tableConfig.startVersion) {
+      throw new DeltaSharingIllegalArgumentException(
+        s"You can only query table data since version ${tableConfig.startVersion}."
+      )
+    }
+
+    streamingOutput(Some(queryResult.version), queryResult.responseFormat, queryResult.actions)
```
chakankardb, line 368: Consider sending back a pending status with 50% likelihood. 
This would allow testing the wait loop on the client side.

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -30,6 +30,18 @@ object ConfUtils {
   val MAX_RETRY_DURATION_CONF = "spark.delta.sharing.network.maxRetryDuration"
   val MAX_RETRY_DURATION_DEFAULT_MILLIS = 10L * 60L* 1000L /* 10 mins */
 
+  val ASYNC_QUERY_POLL_INTERVAL_CONF = "spark.delta.sharing.network.asyncQueryRetryInterval"
+  val ASYNC_QUERY_POLL_INTERVAL_DEFAULT_MILLIS = 10L * 1000L /* 10 seconds */
+
+  val ASYNC_QUERY_MAX_RETRY_COUNT_CONF = "spark.delta.sharing.network.asyncQueryMaxRetryCount"
+  val ASYNC_QUERY_MAX_RETRY_COUNT_DEFAULT = 10
```
chakankardb, line 37: is this used? if not perhaps delete it?

chakankardb (approved):
Some minor nits; but mostly looks good.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -99,8 +115,32 @@ private[sharing] case class QueryTableRequest(
   maxFiles: Option[Int],
   pageToken: Option[String],
   includeRefreshToken: Option[Boolean],
-  refreshToken: Option[String]
-)
+  refreshToken: Option[String],
+  idempotencyKey: Option[String],
+  requestedColumns: Seq[String]
```
linzhou-db, line 120: When/who will set `requestedColumns`?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +781,102 @@ class DeltaSharingRestClient(
...
+  * is false it means the query is finished and the queryResultLines contains the result.
+   */
+  private def checkQueryPending(lines: Seq[String]): (Seq[String], Option[String], Boolean) = {
+    val queryStatus = JsonUtils.fromJson[SingleAction](lines(0)).queryStatus
+
+    if (queryStatus == null) {
+      (lines, None, false)
+    } else {
+      if (queryStatus.queryId == null) {
+        throw new IllegalStateException("QueryId is not returned in the response.")
```
linzhou-db, line 821: do we want to include the full queryStatus in the error message for debugging?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +781,102 @@ class DeltaSharingRestClient(
...
+      val (currentVersion, currentRespondedFormat, currentLines)
+        = getTableQueryInfo(table, queryId, request.maxFiles, request.pageToken)
+      val (newLines, returnedQueryId, newQueryPending) = checkQueryPending(currentLines)
+
+      if(newQueryPending && returnedQueryId.get != queryId) {
+        throw new IllegalStateException("QueryId is not consistent in the response. " +
+          s"Expected: $queryId, Actual: ${returnedQueryId.get}")
+      }
+
+      version = currentVersion
```
linzhou-db, line 870: more of a question, what does `version` mean for shared views?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +781,102 @@ class DeltaSharingRestClient(
...
+
+    // Check if the query is still pending
+    var (lines, queryIdOpt, queryPending) = checkQueryPending(initialLines)
+
+    var version = initialVersion
+    var respondedFormat = initialRespondedFormat
+
+    val startTime = System.currentTimeMillis()
+    // Loop while the query is pending
+    while (queryPending) {
```
linzhou-db, line 852: The timeout is 10 minutes, do we want to log some info for debugging every 10s or 1 minute?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -325,6 +327,60 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+     @Param("schema") schema: String,
+     @Param("table") table: String,
+     @Param("queryId") queryId: String,
+     request: GetQueryInfoRequest): HttpResponse = processRequest {
+
+    if (table == "tableWithAsyncQueryError") {
+      throw new DeltaSharingIllegalArgumentException("expected error")
+    }
+
+    if(rand.nextInt(100) > 50) {
```
linzhou-db, line 345: Could you add comment for this?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -374,50 +430,70 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
-    if (numVersionParams > 0) {
-      if (!tableConfig.historyShared) {
-        throw new DeltaSharingIllegalArgumentException("Reading table by version or timestamp is" +
-          " not supported because history sharing is not enabled on table: " +
-          s"$share.$schema.$table")
-      }
-      if (request.version.exists(_ < tableConfig.startVersion) ||
-        request.startingVersion.exists(_ < tableConfig.startVersion)) {
+
+    if(getAsyncQuery(capabilitiesMap)) {
```
linzhou-db, line 434: nit: do we want to put these two if clause into one?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: we current don't have it supported, remove, and will add later.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -391,14 +433,16 @@ class DeltaSharingRestClient(
       maxFiles = None,
       pageToken = None,
       includeRefreshToken = None,
-      refreshToken = None
+      refreshToken = None,
+      idempotencyKey = None,
+      requestedColumns = Seq.empty[String]
```
linzhou-db: need to remove requestedColumns from here as well?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -311,27 +360,20 @@ class DeltaSharingRestClient(
       maxFiles = None,
       pageToken = None,
       includeRefreshToken = Some(includeRefreshToken),
-      refreshToken = refreshToken
+      refreshToken = refreshToken,
+      idempotencyKey = idempotencyKey,
+      requestedColumns = Seq.empty[String]
```
linzhou-db: need to remove requestedColumns from here as well?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -719,6 +780,104 @@ class DeltaSharingRestClient(
...
+    val startTime = System.currentTimeMillis()
+    // Loop while the query is pending
+    while (queryPending) {
+      if (System.currentTimeMillis() - startTime > asyncQueryMaxDuration) {
+        throw new IllegalStateException("Query is timed out after " +
+          s"${asyncQueryMaxDuration} ms. Please try again later.")
+      }
+
+      val queryId = queryIdOpt.get
+      logInfo(s"Query is still pending. Polling queryId: ${queryId}")
```
linzhou-db, line 857: I worry this results in too many logging as it's one log every second.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
linzhou-db: why do we need this 50% randomness? 
Is this covered by any tests?

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -64,6 +67,12 @@ case class Metadata(
   override def wrap: SingleAction = SingleAction(metaData = this)
 }
 
+case class QueryStatus(
+      queryId: String = null
+      ) extends  Action {
```
linzhou-db, line 72: nit: the lint seems weird.

linzhou-db: Also, doesn't queryStatus contain any status? why it only contains id?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: this is the protocol version, and this code path is used for both table and view queries.

server/src/main/scala/io/delta/sharing/server/model.scala
jadewang-db: currently, we throw exception when there is error happens. and return result if query is success, so getting query status means it's pending, there was a status but I removed after Abhijit's comments, which I think it makes sense.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: this basically a config value, and the default config value is 10s, I changed the default value of the constructor here also to make it consistent.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
jadewang-db: it suppose that the client should cover both cases.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -391,14 +434,15 @@ class DeltaSharingRestClient(
       maxFiles = None,
       pageToken = None,
       includeRefreshToken = None,
-      refreshToken = None
+      refreshToken = None,
+      idempotency_key = None
```
linzhou-db, line 438: nit: use camel case?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -300,6 +344,13 @@ class DeltaSharingRestClient(
     val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
     val target = getTargetUrl(
       s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/query")
+
+    val idempotency_key = if (enableAsyncQuery) {
```
linzhou-db, line 348: nit: use camel case?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
chakankardb: @linzhou-db I had made this suggestion.
The idea is that the server will return that query is still in progress sometimes, which will test the async code path.
@jadewang-db If there is a deterministic way of doing this (maybe return kPending for the first few calls, that is fine as well).

linzhou-db: sorry for the delay on this...

It seems to me that returning "pending" for the first few(3) calls is better than this.

jadewang-db: easist way to do is to make the first call pending

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
jadewang-db: I found out this issue in the integration testing, because the proto we added was with _ , so this will make it consistant with the proto field.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
jadewang-db: @linzhou-db do you feel strong on this? I think this is the best way of returning some pending status, and this is not production code, I think we should be ok here?

linzhou-db (approved):

PR #494, opened 2024-06-05
Code refactor to split logic for DeltaSharedTableLoader and DeltaSharedTable

pranavsuku-db:
Part of effort for supporting Delta Sharing Python to Read Shared DV/CM tables. Refactoring code so that a new DeltaSharedTableKernel can be created and the loader does not care about the DeltaSharedTableImplementation. 

In this refactor I brought the DeltaSharedTableLoader to its own file in the server package. I renamed the old file to DeltaSharedTable


charlenelyu-db (approved):
LGTM

PR #497, opened 2024-06-10
Fix test and protocol description about delta sharing streaming rpc internal

linzhou-db:
Fix test and protocol description about delta sharing streaming rpc internal


pranavsuku-db:
thanks for the fix, lgtm!

pranavsuku-db (approved):

PR #498, opened 2024-06-11
DTS-3397: do not log URLs on access errors

PatrickJin-db:
When the client accesses the URI of a file in a shared table, it no longer logs the URI. This addresses the security concerns of logging presigned URLs.


linzhou-db (approved):
The change itself looks good, I'm a bit concerned that it affects our ability to debug issues....

client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
```diff
@@ -167,7 +167,8 @@ private[sharing] class RandomAccessHttpInputStream(
             }
           }
           throw new UnexpectedHttpStatus(
```
linzhou-db, line 169: could you add logError with the error message that includes the uri?

PR #507, opened 2024-06-18
Fix flakiness test

jadewang-db:
fix the flakiness server test by only random query status for a specific table which which used by the rest client.


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.1.0"
+version in ThisBuild := "1.1.1-SNAPSHOT"
```
linzhou-db, line 1: revert this?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -364,9 +364,10 @@ class DeltaSharingService(serverConfig: ServerConfig) {
     }
 
     // simulate async query with 50% chance of return
-    // asynchronously client should be able to hand both cases
-    // this is for test purpose only and shouldn't cause flakiness
-    if(rand.nextInt(100) > 50) {
+    // asynchronously for a specific table
+    // client should be able to hand both cases and for server
```
linzhou-db, line 368: nit: hand -> handle.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -166,7 +166,7 @@ class DeltaSharingRestClient(
     asyncQueryMaxDuration: Long = 600000L
   ) extends DeltaSharingClient with Logging {
 
-  logError(s"DeltaSharingRestClient with enableAsyncQuery $enableAsyncQuery")
+  logInfo(s"DeltaSharingRestClient with enableAsyncQuery $enableAsyncQuery")
```
linzhou-db, line 169: is this intended?

jadewang-db: yes

linzhou-db (approved):
Thanks for the fix!

PR #509, opened 2024-06-20
Feature fix load as spark

### LINKED ISSUE #508: load_as_spark() returns error  "No active SparkSession was found. load_as_spark requires running in a PySpark application." when used in Django Rest Api.
Hi all I am trying to load a delta table using load_as_spark() from the delta-sharing library in a rest API in the Django app. The issue is that when I run my Django service a spark session ( from a  different app in the same project ) starts on startup and when I hit the rest API for reading the delta table using delta protocol it gives me an error:
"No active SparkSession was found. `load_as_spark` requires running in a PySpark application." 
Due to a condition check in load_as_spark():
`spark = SparkSession.getActiveSession()
assert spark is not None, (
    "No active SparkSession was found. "
    "`load_as_spark` requires running in a PySpark application."
)` 

the SparkSession.getActiveSession() returns a spark session only if the spark session is from the current thread which in my case was started in a different thread and therefore I am stuck in loading data from delta table.

Shabbir-Khan-12:
Closes  #508

Used SparkSession.active() instead of SparkSession.getActiveSession() in Delta-Sharing.py which returns a spark session regardless of any active session being present or not.

### DIFF
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -150,7 +150,11 @@ def load_as_spark(
     except ImportError:
         raise ImportError("Unable to import pyspark. `load_as_spark` requires PySpark.")
 
-    spark = SparkSession.active()
+    spark = SparkSession.getActiveSession()
+    assert spark is not None, (
+        "No active SparkSession was found. "
+        "`load_as_spark` requires running in a PySpark application."
+    )
     df = spark.read.format("deltaSharing")
     if version is not None:
         df.option("versionAsOf", version)
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -150,11 +150,7 @@ def load_as_spark(
     except ImportError:
         raise ImportError("Unable to import pyspark. `load_as_spark` requires PySpark.")
 
-    spark = SparkSession.getActiveSession()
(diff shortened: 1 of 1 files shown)


python/delta_sharing/delta_sharing.py
```diff
@@ -150,11 +150,7 @@ def load_as_spark(
     except ImportError:
         raise ImportError("Unable to import pyspark. `load_as_spark` requires PySpark.")
 
-    spark = SparkSession.getActiveSession()
-    assert spark is not None, (
```
linzhou-db, line 154: do we still want to keep the assert?

Shabbir-Khan-12: It will not be beneficial to keep it. If any case occurs with issues in spark session fetching then this assert will handle that.

Shabbir-Khan-12: I want to hear your thoughts as to what should be the better approach.

linzhou-db: Sorry for the delay, I'm reading this, it seems it will default to `None`, so I think this assert is still helpful.  https://spark.apache.org/docs/latest/api/java/org/apache/spark/sql/SparkSession.html

linzhou-db:
feel free to reopen

PR #513, opened 2024-06-21
limit hints for query supported by delta-kernel

pranavsuku-db:
limit hints for query supported by delta-kernel


linzhou-db (approved):

linzhou-db:
please fix related tests before merging.

PR #517, opened 2024-06-25
column mapping functionality added to server

pranavsuku-db:
added column mapping functionality to server and wrote 2 corresponding tests

### DIFF
--- a/server/src/main/scala/io/delta/common/SnapshotChecker.scala
+++ b/server/src/main/scala/io/delta/common/SnapshotChecker.scala
@@ -38,11 +38,8 @@ object SnapshotChecker {
    *                      properties of the table.
    * @param tableVersion  The table vesrion associated with the configuration.
    * @param clientReaderFeatures The set of reader features supported by the delta sharing client
-   * @param flagReaderFeatures The set of reader features allowed by the flag, used to enabling
-   *                           flags for providers to share tables with specific reader features.
    */
   def assertTableProperties(
-      isProviderRpc: Boolean,
       configuration: Map[String, String],
       tableVersion: Option[Long],
       clientReaderFeatures: Set[String]): Unit = {
--- a/server/src/main/scala/io/delta/common/SnapshotChecker.scala
+++ b/server/src/main/scala/io/delta/common/SnapshotChecker.scala
@@ -1,0 +1,138 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.common
+
+import io.delta.sharing.server.DeltaSharingUnsupportedOperationException
+import io.delta.sharing.server.actions.{ColumnMappingTableFeature, DeltaAction}
+
+
+object SnapshotChecker {
+
+  /**
+   * Assert that the client is up to date with the protocol and allowed
+   * to read the table that is using this Snapshot's `protocol`.
+   */
+  def assertProtocolRead(tblMinReaderVersion: Int): Unit = {
+    val clientVersion = DeltaAction.readerVersion
+    if (clientVersion < tblMinReaderVersion) {
+      throw new DeltaSharingUnsupportedOperationException("Unsupported Delta Reader Version")
+    }
+  }
+
+  /**
+   * Assert all properties present in the table are covered/supported either by clientReaderFeatures
+   * or flagReaderFeatures.
+   *
+   * If a table property is covered by clientReaderFeatures or flagReaderFeatures, then it's
+   * considered as supported, if not, check if the property value is disabled,
+   * if not it's considered unsupported, and should throw an error.
+   *
+   * This should fail all parquet format requests on tables with property values not in
+   * tablePropertiesWithDisabledValues, because clientReaderFeatures is empty for parquet format
+   * sharing rpcs, which won't filter out any properties in tablePropertiesWithDisabledValues.
+   *
+   * @param configuration The "configuration" field of a delta Metadata, which contains the
+   *                      properties of the table.
+   * @param tableVersion  The table vesrion associated with the configuration.
+   * @param clientReaderFeatures The set of reader features supported by the delta sharing client
+   * @param flagReaderFeatures The set of reader features allowed by the flag, used to enabling
+   *                           flags for providers to share tables with specific reader features.
+   */
+  def assertTableProperties(
+                             isProviderRpc: Boolean,
+                             configuration: Map[String, String],
+                             tableVersion: Option[Long],
+                             clientReaderFeatures: Set[String],
+                             flagReaderFeatures: Set[String] = Set.empty): Unit = {
+    // An unsupported table property can be supported if it is in part of the client supported
+    // table features.
+    def propertySupportedByClientOrFlag(property: String): Boolean = {
+      // TODO: @pranavsuku-db add deletionvector logic
+      if (property == DeltaAction.columnMappingProperty.property) {
+        ColumnMappingTableFeature.isInSet(clientReaderFeatures) ||
+          ColumnMappingTableFeature.isInSet(flagReaderFeatures)
+      } else {
+        // We should not reject any other table properties as they can contain arbitrary keys.
+        true
+      }
+    }
+    val unsupportedPropertiesByClient = DeltaAction.tablePropertiesWithDisabledValues
+      .flatMap {
+        case pr @ DeltaAction.PropertyAllowedValues(property, allowedValues)
+          if !propertySupportedByClientOrFlag(property) =>
+          configuration.get(property).filterNot(allowedValues.contains(_)).map(_ => pr)
+        case _ => None
+      }
+    if (unsupportedPropertiesByClient.nonEmpty) {
+      throw new DeltaSharingUnsupportedOperationException("Unsupported Delta Table Features")
+    }
+  }
+
+  def getTableFeaturesFromProperties(properties: Map[String, String]): Set[String] =
+    properties.keys
+      .filter(_.contains("delta.feature."))
+      .map { featureProp =>
+        val startIndx = "delta.feature.".length
+        featureProp.substring(startIndx)
+      }
+      .toSet
+
+  /**
+   * Assert all table features are supported by both the server and the client.
+   *
+   * This will NOT fail parquet format requests on tables with advanced features, because
+   * features like DV will persist in protocol.readerFeatures even if it's turned off by
+   * setting table properties. We rely on assertTableProperties to fail the request if
+   * the table properties are actually enabled.
+   */
+  def assertTableFeatures(
+      isProviderRpc: Boolean,
+      kernelEnabled: Boolean,
+      tableFeatures: Set[String],
+      tableVersion: Option[Long],
+      clientReaderFeatures: Set[String] = Set.empty): Unit = {
+    // Check all table features are supported by the server.
+    val supportedTableFeatures = if (kernelEnabled) {
+      DeltaAction.kernelSupportedTableFeatures
+    } else {
+      DeltaAction.supportedTableFeatures
+    }
+    // scalastyle:off caselocale
+    val readerUnsupportedFeatures = tableFeatures.map(_.toLowerCase) --
+      supportedTableFeatures.map(_.name.toLowerCase)
+    // scalastyle:on caselocale
+    if (readerUnsupportedFeatures.nonEmpty) {
+      throw new DeltaSharingUnsupportedOperationException("Unsupported Delta Table Features")
+    }
+
+    // Check all table features are supported by the client.
+    // We only need to check DV and CM here because other features does not require
+    // support from the delta sharing client.
+    val clientUnsupportedFeatures =
+      tableFeatures
+        .filter { feature =>
+          // TODO @pranavsuku-db add deletionvector table feature here
+          feature == ColumnMappingTableFeature.name
+        }
+        // scalastyle:off caselocale
+        .map(_.toLowerCase) -- clientReaderFeatures.map(_.toLowerCase)
+    // scalastyle:on caselocale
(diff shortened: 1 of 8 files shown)


server/src/main/scala/io/delta/common/SnapshotChecker.scala
```diff
@@ -0,0 +1,138 @@
...
+import io.delta.sharing.server.actions.{ColumnMappingTableFeature, DeltaAction}
+
+
+object SnapshotChecker {
+
+  /**
+   * Assert that the client is up to date with the protocol and allowed
+   * to read the table that is using this Snapshot's `protocol`.
+   */
+  def assertProtocolRead(tblMinReaderVersion: Int): Unit = {
```
linzhou-db, line 29: Actually we need to add test for these functions.

server/src/main/scala/io/delta/common/SnapshotChecker.scala
```diff
@@ -0,0 +1,138 @@
...
+   *
+   * @param configuration The "configuration" field of a delta Metadata, which contains the
+   *                      properties of the table.
+   * @param tableVersion  The table vesrion associated with the configuration.
+   * @param clientReaderFeatures The set of reader features supported by the delta sharing client
+   * @param flagReaderFeatures The set of reader features allowed by the flag, used to enabling
+   *                           flags for providers to share tables with specific reader features.
+   */
+  def assertTableProperties(
+                             isProviderRpc: Boolean,
```
linzhou-db, line 56: nit: lint

server/src/main/scala/io/delta/common/SnapshotChecker.scala
```diff
@@ -0,0 +1,138 @@
...
+                             configuration: Map[String, String],
+                             tableVersion: Option[Long],
+                             clientReaderFeatures: Set[String],
+                             flagReaderFeatures: Set[String] = Set.empty): Unit = {
```
linzhou-db, line 60: There are no flags in the oss code, you can just remove all flag related code.

server/src/main/scala/io/delta/common/SnapshotChecker.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  /**
+   * Assert all table features are supported by both the server and the client.
+   *
+   * This will NOT fail parquet format requests on tables with advanced features, because
+   * features like DV will persist in protocol.readerFeatures even if it's turned off by
+   * setting table properties. We rely on assertTableProperties to fail the request if
+   * the table properties are actually enabled.
+   */
+  def assertTableFeatures(
```
linzhou-db, line 102: similarly we need to add test for these functions, otherwise we can remove them.

server/src/main/scala/io/delta/kernelsharedtable/DeltaSharedTableKernel.scala
```diff
@@ -741,7 +761,6 @@ private case class SharedTableSnapshot(
     version: Long,
     protocol: KernelProtocol,
     metadata: KernelMetadata,
-    // versionStats: Option[VersionStats],
```
linzhou-db, line 744: why are we removing versionStats?

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -1410,6 +1414,165 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
     verifyPreSignedUrl(actualFiles(1).url, 573)
   }
 
+  integrationTest("column mapping query test 1 - /shares/{share}/schemas/{schema}/tables/{table}/metadata") {
```
linzhou-db, line 1417: could you update the test names to match the actual query you are using? `/metadata` -> `/query`

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -1410,6 +1414,165 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
...
+          )
+        )
+      )
+    )
+    assert(expectedFiles == actualFiles.toList)
+    assert(actualFiles.count(_.expirationTimestamp != null) == 1)
+    verifyPreSignedUrl(actualFiles(0).deltaSingleAction.add.path, 688)
+  }
+
+  integrationTest("column mapping query test 2 - /shares/{share}/schemas/{schema}/tables/{table}/metadata") {
```
linzhou-db, line 1479: could you update the test names to match the actual query you are using? `/metadata` -> `/query`

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -1410,6 +1414,165 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
...
+        )
+      )
+    )
+    assert(expectedFiles == actualFiles.toList)
+    assert(actualFiles.count(_.expirationTimestamp != null) == 3)
+    verifyPreSignedUrl(actualFiles(0).deltaSingleAction.add.path, 653)
+    verifyPreSignedUrl(actualFiles(1).deltaSingleAction.add.path, 652)
+    verifyPreSignedUrl(actualFiles(2).deltaSingleAction.add.path, 653)
+  }
+
```
linzhou-db, line 1575: And add two tests for `/metadata` on the two tables.

pranavsuku-db: @linzhou-db is this necessary since I am checking the metadata and protocol in the query function anyways? If I do write a test just for query it will just be the top half of this test copied and pasted directly

server/src/main/scala/io/delta/kernelsharedtable/DeltaSharedTableKernel.scala
pranavsuku-db: this was never being used and was always commented out. I am just doing a code cleanup since I noticed it now. I should have removed it in a previous PR, sorry about that

linzhou-db: could you double check with @charlenelyu-db on whether it's expected to be empty?

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
linzhou-db: Yes this is necessary, as we need to make sure the rpc works as expected. 

And actually, we want to add test with different header values for `deltaSharingCapabilities` and verify the server behavior.

server/src/main/scala/io/delta/common/SnapshotChecker.scala
pranavsuku-db: fixed

server/src/main/scala/io/delta/common/SnapshotChecker.scala
pranavsuku-db: fixed

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
pranavsuku-db: fixed

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
pranavsuku-db: fixed

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
pranavsuku-db: Ok I will add the 2 metadata tests

For the other part, should I pass nothing into the readerFeatures (as opposed to passing in readerFeatures = columnMapping) and see if an error is being thrown (it should throw an error in assertTableProperties and this would address the other comment you made right)?

linzhou-db: For the other part, we should test:
- responseFormat=parquet;readerFeatures=columnMapping -> this should result in error
- responseFormat=parquet,delta;readerFeatures=columnMapping -> this should result in delta format response
- responseFormat=deltal; -> with empty reader features, which should throw error. BUT, it may not throw exception as we expected, and we may want to skip it for now and double check why the reader features is empty in the tables I created for testing.

pranavsuku-db: added tests for these 3 and it all worked as expected

server/src/main/scala/io/delta/common/SnapshotChecker.scala
pranavsuku-db: I think assertTableFeatures is not useful and here is why I think it can be removed:

For this function there is a comment 

> Assert all table features are supported by both the server and the client.
>   This will NOT fail parquet format requests on tables with advanced features, because
>    features like DV will persist in protocol.readerFeatures even if it's turned off by
>    setting table properties. We rely on assertTableProperties to fail the request if
>    the table properties are actually enabled.

assertTableProperties checks if features are not supported by the client. The assertTableFeatures is mainly checking if something is not handled by the server, but since we are using kernel, column mapping and deletion vectors will be supported by the server. The tests I wrote in the most recent commit go through assertTableProperties and fail there since the client cannot support them.

server/src/main/scala/io/delta/common/SnapshotChecker.scala
pranavsuku-db: How can I write a test where the client is not up to date with the protocol? Would I need to write a client test for this in Python?

server/src/main/scala/io/delta/kernelsharedtable/DeltaSharedTableKernel.scala
pranavsuku-db: VersionStats is used for shadow traffic and for the size and numFiles in metadata response (which are both not needed I believe). So I am just not bothering to set it when getting the snapshot. Confirmation on this would be appreciated though

server/src/main/scala/io/delta/common/SnapshotChecker.scala
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * @param configuration The "configuration" field of a delta Metadata, which contains the
+   *                      properties of the table.
+   * @param tableVersion  The table vesrion associated with the configuration.
+   * @param clientReaderFeatures The set of reader features supported by the delta sharing client
+   * @param flagReaderFeatures The set of reader features allowed by the flag, used to enabling
+   *                           flags for providers to share tables with specific reader features.
+   */
+  def assertTableProperties(
+      isProviderRpc: Boolean,
```
linzhou-db, line 45: remove all the reference to `isProviderRpc`, it's not useful in the oss code. 

And update the `param` list in the function header.

server/src/main/scala/io/delta/sharing/server/DeltaAction.scala
```diff
@@ -34,11 +34,53 @@ object DeltaAction {
...
+   * However, they are not supported by the Delta standalone reader at this point.
+   * Delta standalone reader should still try to read these tables given the properties are not set.
+   */
+
+  val columnMappingProperty: PropertyAllowedValues =
+    PropertyAllowedValues("delta.columnMapping.mode", Seq("none"))
+  val tablePropertiesWithDisabledValues: Seq[PropertyAllowedValues] =
+    Seq(columnMappingProperty)
+
+  val supportedTableFeatures: Set[TableFeature] = Set(
```
linzhou-db, line 50: is this and L56 used anywhere?

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -1410,6 +1412,393 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
...
+        configuration = expectedConfigMap,
+        createdTime = Option(1719346404750L)
+      )
+    ).wrap
+    assert(expectedMetadata == JsonUtils.fromJson[DeltaResponseSingleAction](metadata))
+  }
+
+  integrationTest("column mapping metadata test 3 - /shares/{share}/schemas/{schema}/tables/{table}/metadata") {
+    val parquetAndDeltaResponse = RESPONSE_FORMAT_PARQUET + "," + RESPONSE_FORMAT_DELTA
+    // scalastyle:off println
```
linzhou-db, line 1484: clean up these.

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -1410,6 +1412,393 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
...
+        partitionColumns = Seq("date"),
+        configuration = expectedConfigMap,
+        createdTime = Option(1719346404750L)
+      )
+    ).wrap
+    assert(expectedMetadata == JsonUtils.fromJson[DeltaResponseSingleAction](metadata))
+  }
+
+  integrationTest("column mapping metadata test 3 - /shares/{share}/schemas/{schema}/tables/{table}/metadata") {
+    val parquetAndDeltaResponse = RESPONSE_FORMAT_PARQUET + "," + RESPONSE_FORMAT_DELTA
```
linzhou-db, line 1483: Add a comment on what it is testing.

linzhou-db (approved):
Maybe in a separate PR: consider use some utils functions to make the test changes shorter

server/src/main/scala/io/delta/common/SnapshotChecker.scala
pranavsuku-db: done

server/src/main/scala/io/delta/sharing/server/DeltaAction.scala
pranavsuku-db: not being used right now I will delete and reintroduce for the DV PR if needed

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
pranavsuku-db: done

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
pranavsuku-db: done

linzhou-db (approved):

PR #519, opened 2024-06-28
Support for Delta Format RPC in Python Client

pranavsuku-db:
Set headers for delta format and wrote test for querying delta protocol/metadata from a DV table

### DIFF
--- a/python/delta_sharing/protocol.py
+++ b/python/delta_sharing/protocol.py
@@ -184,7 +184,7 @@ class DeltaProtocol:
     writer_features: list[str]
 
     @staticmethod
-    def from_json(json) -> "Protocol":
+    def from_json(json) -> "DeltaProtocol":
         if isinstance(json, (str, bytes, bytearray)):
             json = loads(json)
         return DeltaProtocol(
@@ -256,7 +256,7 @@ class DeltaMetadata:
     created_time: Optional[int] = None
 
     @staticmethod
-    def from_json(json) -> "Metadata":
+    def from_json(json) -> "DeltaMetadata":
         if isinstance(json, (str, bytes, bytearray)):
             json = loads(json)
         if "configuration" in json["deltaMetadata"]:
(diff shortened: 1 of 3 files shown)


python/delta_sharing/protocol.py
```diff
@@ -176,6 +176,24 @@ def from_json(json) -> "Protocol":
         return Protocol(min_reader_version=int(json["minReaderVersion"]))
 
 
+@dataclass(frozen=True)
+class DeltaProtocol:
```
charlenelyu-db, line 180: Could you remind me when do we need to parse the json for delta protocol/metadata? I thought all we need to do is to dump the raw json into temp files, then kernel-rust will read from those files and parse the json instead. 

I'm not familiar with kernel-rust interface though, we should define these classes only when the inner fields are needed when we use the kernel-rust API.

python/delta_sharing/rest_client.py
```diff
@@ -71,6 +73,13 @@ class QueryTableMetadataResponse:
     metadata: Metadata
 
 
+@dataclass(frozen=True)
+class QueryTableDeltaMetadataResponse:
```
charlenelyu-db, line 77: I feel we could just add fields into existing `QueryTableMetadataResponse`, and add one more string field indicating the responded format.

python/delta_sharing/protocol.py
pranavsuku-db: Mainly for testing it helps to parse the object into JSON so that each property can be properly checked. 

I am also going to use the JSON for autoresolving the query format:
- if the response can be parsed into Protocol then we stick to original codepath
- if the response cannot be parsed into Protocol but can be parsed into DeltaProtocol we know we need to use the new codepath using kernel-rust

python/delta_sharing/rest_client.py
pranavsuku-db: I was thinking about this but logically I felt the response should only contain what is needed for the response type. If it is not using delta, why have empty fields for it when we can create a new object to handle it. I am fine either way though

PR #523, opened 2024-07-08
Delta-Kernel-Rust Integration (old)

pranavsuku-db:
Store delta log locally and send temporary file to delta-kernel-rust


pranavsuku-db:
Still a WIP, not ready for formal review yet

pranavsuku-db:
@linzhou-db I tried adding the maturin develop to setup.py but I think this command is not being run in the package itself since I am still getting an error for the unresolved import in python tests. I took a look into this, but don't know how to put the OS command in the setup() function in setup.py which I think might be the solution.

python/delta-kernel-python/src/lib.rs
```diff
@@ -0,0 +1,170 @@
+use std::sync::Arc;
```
linzhou-db, line 1: @nicklan  Could you help review this?

python/delta_sharing/reader.py
```diff
@@ -85,11 +89,73 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
             timestamp=self._timestamp
         )
 
+    def to_pandas_kernel(self):
+        self._rest_client.set_delta_format_header()
```
linzhou-db, line 93: let's correct the name of this function. it is actually setting both format in the header.

linzhou-db: and if needed, create a function `set_delta_format_header` which only set delta format in the header.

python/delta_sharing/reader.py
```diff
@@ -85,11 +89,73 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
...
+
+        return result
+
     def to_pandas(self) -> pd.DataFrame:
         response_format = self._rest_client.autoresolve_query_format(self._table)
 
         if (response_format == DataSharingRestClient.DELTA_FORMAT):
-            raise Exception("Delta format not supported in query yet.")
+            return self.to_pandas_kernel()
+            # raise Exception("Delta format not supported in query yet.")
```
linzhou-db, line 158: clean up this comment?

python/delta_sharing/rest_client.py
```diff
@@ -408,16 +409,28 @@ def list_files_in_table(
...
 
-            return ListFilesInTableResponse(
-                delta_table_version=int(headers.get(
-                    DataSharingRestClient.DELTA_TABLE_VERSION_HEADER)),
-                protocol=Protocol.from_json(protocol_json["protocol"]),
-                metadata=Metadata.from_json(metadata_json["metaData"]),
-                add_files=[AddFile.from_json(json.loads(file)["file"]) for file in lines],
-            )
+            if (DataSharingRestClient.CAPABILITIES_HEADER in headers and
+                    "responseformat=delta" in headers[DataSharingRestClient.CAPABILITIES_HEADER]):
```
linzhou-db, line 414: is there a better way than directly comparing the string "responseformat=delta"?

python/delta_sharing/reader.py
pranavsuku-db: done

python/delta_sharing/reader.py
pranavsuku-db: done changed it to set_sharing_capabilities_header()

python/delta_sharing/rest_client.py
pranavsuku-db: the only other way I can think of is to try to serialize it into a json and try catch the serialization process. This might be extra overhead though

python/delta_sharing/reader.py
```diff
@@ -85,11 +89,72 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
             timestamp=self._timestamp
         )
 
+    def to_pandas_kernel(self):
```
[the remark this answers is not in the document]
pranavsuku-db: wdym function header? sure i can make it private

python/delta_sharing/reader.py
```diff
@@ -85,11 +89,72 @@ def limit(self, limit: Optional[int]) -> "DeltaSharingReader":
...
+        self._rest_client.set_sharing_capabilities_header()
```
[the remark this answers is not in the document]
pranavsuku-db: done

python/delta-kernel-python/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+                            .into_any()
+                            .downcast::<ArrowEngineData>()
+                            .map_err(|_| {
+                                ArrowError::CastError("Couldn't cast to ArrowEngineData".to_string())
+                            })?
+                            .into();
+                        if let Some(mask) = res.mask {
+                            let filtered_batch = filter_record_batch(&record_batch, &mask.into())?;
+                            Ok(filtered_batch)
+                        } else {
```
nicklan, line 123:
```suggestion
                        if let Some(mask) = res.mask {
                            let extra_rows = record_batch.num_rows() - mask.len();
                            if extra_rows > 0 {
                                // we need to extend the mask here in case it's too short
                                mask.extend(std::iter::repeat(true).take(extra_rows));
                            }
                            let filtered_batch = filter_record_batch(&record_batch, &mask.into())?;
                            Ok(filtered_batch)
                        } else {
```

python/delta_sharing/reader.py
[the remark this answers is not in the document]
linzhou-db: comments of the function, what it does, input, output, etc.

linzhou-db (approved):

PR #524, opened 2024-07-08
refactored server-side code to cleaner packages

pranavsuku-db:
Moved `Codec.scala`, `DeletionVectorDescriptor.scala`, `DeltaAction.scala` into `actions` package in `common`. Moved `CloudFileSigner.scala` and `DeltaAction.scala` into `common`. This meant changing the imports in some other files. 

This PR is purely a refactor and does not change any core logic. Does not require any additional testing since it is just a refactor

### DIFF
--- a/server/src/main/scala/io/delta/common/CloudFileSigner.scala
+++ b/server/src/main/scala/io/delta/common/CloudFileSigner.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.delta.sharing.server
+package io.delta.common
 
 import java.net.URI
 import java.util.Date
--- a/server/src/main/scala/io/delta/common/SnapshotChecker.scala
+++ b/server/src/main/scala/io/delta/common/SnapshotChecker.scala
@@ -16,8 +16,9 @@
 
 package io.delta.common
 
+import io.delta.common.actions.{ColumnMappingTableFeature, DeletionVectorsTableFeature, DeltaAction}
+
 import io.delta.sharing.server.DeltaSharingUnsupportedOperationException
-import io.delta.sharing.server.actions.{ColumnMappingTableFeature, DeletionVectorsTableFeature, DeltaAction}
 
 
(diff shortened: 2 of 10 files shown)


pranavsuku-db:
The tests that are failing are connection timeouts and these tests have been flaky. Not related to to this work

server/src/main/scala/io/delta/common/CloudFileSigner.scala
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.delta.sharing.server
+package io.delta.common
```
linzhou-db, line 17: should it be io.delta.sharing.server.common?

pranavsuku-db: fixed

pranavsuku-db:
@linzhou-db I will probably address this in the JSON predicate hints PR where I add more files related to kernel, but do you have any ideas for what I should name the current package `kernelsharedtable` to? It is not consistent with the rest of the package names and feels a bit clunky. The issue is I cannot rename it to `kernel.internal` similar to how `standalone.internal` is named since that will conflict with the actual import from kernel.internal. One idea is we rename standalone.internal to `sharing.standalone.internal `and do the same for the new `sharing.kernel.internal`

linzhou-db (approved):

linzhou-db:
> @linzhou-db I will probably address this in the JSON predicate hints PR where I add more files related to kernel, but do you have any ideas for what I should name the current package `kernelsharedtable` to? It is not consistent with the rest of the package names and feels a bit clunky. The issue is I cannot rename it to `kernel.internal` similar to how `standalone.internal` is named since that will conflict with the actual import from kernel.internal. One idea is we rename standalone.internal to `sharing.standalone.internal `and do the same for the new `sharing.kernel.internal`

`sharing.standalone.internal `and `sharing.kernel.internal` sounds good to me.

PR #525, opened 2024-07-10
JSON Predicate Hints

pranavsuku-db:
Server side push down JSON Predicate Hints


server/src/main/scala/io/delta/kernelsharedtable/DeltaSharedTableKernel.scala
```diff
@@ -324,9 +325,18 @@ class DeltaSharedTableKernel(
     )
 
     if (includeFiles) {
-      if (predicateHints.isEmpty && jsonPredicateHints.isEmpty) {
+      if (predicateHints.isEmpty) {
```
linzhou-db, line 328: nit: what if we do a reverse check, throw the exception is this is not empty. 
Then we can put the rest of the logic outside the if clause.

server/src/main/scala/io/delta/kernelsharedtable/JsonPredicatePruner.scala
```diff
@@ -0,0 +1,172 @@
+/*
```
linzhou-db, line 1: I assume this is mostly copied and don't have much changes?

server/src/main/scala/io/delta/kernelsharedtable/JsonPredicates.scala
```diff
@@ -0,0 +1,492 @@
+/*
```
linzhou-db, line 1: Isn't this already available in server/src/main/scala/io/delta/standalone/internal/JsonPredicates.scala?

linzhou-db:
Could you also check the test failures?

server/src/main/scala/io/delta/kernelsharedtable/JsonPredicatePruner.scala
pranavsuku-db: yup, no changes

server/src/main/scala/io/delta/kernelsharedtable/DeltaSharedTableKernel.scala
pranavsuku-db: I can actually just delete this if statement entirely since the kernel function doesn't get executed if predicateHints.isEmpty since that is being caught in DeltaSharingService.scala. Should I remove the double check here?

server/src/main/scala/io/delta/kernelsharedtable/JsonPredicates.scala
pranavsuku-db: Yeah it seems very similar, let me pull the version from internal into the common folder, remove this one and see what happens

pranavsuku-db:
@linzhou-db the test failures are just those flaky timeouts, nothing introduced by the code changes here

server/src/main/scala/io/delta/kernelsharedtable/JsonPredicates.scala
pranavsuku-db: done, this worked and I moved the standalone JsonPredicates.scala to common and now the kernel and standalone code uses this same file

server/src/main/scala/io/delta/kernelsharedtable/DeltaSharedTableKernel.scala
pranavsuku-db: Alternatively if we want the double check in case the kernel implementation gets called when it shouldn't and we want to throw an error, I should be consistent and do the reverse check on predicateHints, maxFiles, startingVersion, and pageToken

linzhou-db: either way works. 
I'd prefer let's remove this check and add a comment saying that predicateHints is not supported.

linzhou-db (approved):

PR #529, opened 2024-07-18
Increase delta-kernel-rust version to 0.2

pranavsuku-db:
Support for DV and column mapping with column partition

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, 3.10]
+        python-version: [3.8, 3.9]
         include:
           - pandas-version: 2.0.3
             pyarrow-version: 16.1.0
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,7 +34,7 @@ jobs:
       matrix:
         python-version: [3.7, 3.8, 3.9]
         include:
-          - pandas-version: 1.2.4
+          - pandas-version: 2.2.2
             pyarrow-version: 16.1.0
     env:
       PYTHON_VERSION: ${{ matrix.python-version }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9, 3.10]
         include:
           - pandas-version: 2.0.3
             pyarrow-version: 16.1.0
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,7 +34,7 @@ jobs:
       matrix:
         python-version: [3.7, 3.8, 3.9]
         include:
-          - pandas-version: 2.2.2
+          - pandas-version: 2.0.3
             pyarrow-version: 16.1.0
     env:
       PYTHON_VERSION: ${{ matrix.python-version }}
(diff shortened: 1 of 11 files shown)


python/delta-kernel-python/src/lib.rs
```diff
@@ -88,59 +87,38 @@ impl Scan {
         &self,
         engine_interface: &PythonInterface,
     ) -> DeltaPyResult<PyArrowType<Box<dyn RecordBatchReader + Send>>> {
-        let mut results = self.0.execute(engine_interface.0.as_ref())?;
-        match results.pop() {
-            Some(last) => {
-                // TODO(nick): This is a terrible way to have to get the schema
-                let rb: RecordBatch = last
-                    .raw_data?
+        let result_schema: SchemaRef =
```
linzhou-db, line 90: I assume these changes are all suggested by @nicklan

server/src/main/scala/io/delta/sharing/kernel/internal/DeltaSharedTableKernel.scala
```diff
@@ -703,27 +703,24 @@ class DeltaSharedTableKernel(
...
-      // scalastyle:off caselocale
       if (!DeletionVectorsTableFeature.isInSet(clientReaderFeaturesSet)) {
         throw new DeltaSharingUnsupportedOperationException("Deletion Vector property disabled")
       }
-      // scalastyle:on caselocale
       val dvAbsolutePath = dvDescriptor.absolutePath(dataPath)
       val presignedDvUrl = fileSigner.sign(dvAbsolutePath)
       minUrlExpirationTimestamp = minUrlExpirationTimestamp.min(presignedDvUrl.expirationTimestamp)
       dvDescriptor = dvDescriptor.copy(pathOrInlineDv = presignedDvUrl.url, storageType = "p")
+      numRecords -= dvDescriptor.cardinality
```
linzhou-db, line 721: can we add a server side test for this?

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -484,32 +485,229 @@ def test_get_table_protocol(profile_path: str):
...
+            ),
+            id="dv",
+        ),
+        pytest.param(
+            "share8.default.deletion_vectors_with_dvs_dv_property_on",
+            6,
+            None,
+            pd.DataFrame(
+                {
+                    "id": [3, 4, 5, 6, 7, 8],
```
linzhou-db, line 541: How did you confirm that the returned dataframe is correct?

linzhou-db (commented):
And we need

linzhou-db (approved):
Approving, but let's do 2 follow-ups:
- add that server side testing.
- test what will happen if we add 3.10 and 3.11

python/delta-kernel-python/src/lib.rs
pranavsuku-db: yup

python/delta_sharing/tests/test_delta_sharing.py
pranavsuku-db: i printed it out on dogfood

server/src/main/scala/io/delta/sharing/kernel/internal/DeltaSharedTableKernel.scala
pranavsuku-db: yup will add

.github/workflows/build-and-test.yml
```diff
@@ -32,9 +32,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
```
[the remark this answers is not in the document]
pranavsuku-db: done

PR #533, opened 2024-07-23
Add python release script

linzhou-db:
Add python release script


python/setup.py
```diff
@@ -78,11 +79,10 @@
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: OS Independent",
         "Topic :: Software Development :: Libraries :: Python Modules",
-        'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
```
pranavsuku-db, line 83: Do we want to add 3.10 and 3.11 here?

python/setup.py
```diff
@@ -41,14 +41,15 @@
...
         'pandas',
-        'pyarrow>=4.0.0',
+        'pyarrow>=16.1.0',
         'fsspec>=0.7.4',
         'requests',
         'aiohttp',
-        'dataclasses;python_version<"3.7"',
+        'dataclasses;python_version<"3.8"',
         'yarl>=1.6.0',
         'maturin',
```
pranavsuku-db, line 54: we might be able to remove maturin here since it is only used to build delta_lernel_rust_sharing_wrapper but this is now being done in python_release.sh. Not totally sure though but if we can avoid having user install an extra package this might be a good thing

pranavsuku-db:
lgtm but I have not tested this on my local machine yet

pranavsuku-db (approved):

python/setup.py
linzhou-db: updated

python/setup.py
linzhou-db: added

PR #535, opened 2024-07-25
Support Empty Table Sharing with Kernel

pranavsuku-db:
Support query of tables with no data (includes LIMIT 0 queries)

### DIFF
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -32,8 +32,6 @@ from delta_sharing.rest_client import DataSharingRestClient
 
 
 class DeltaSharingReader:
-    FILE_NUMBER_WARNING_THRESH = 5000
-
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


python/delta_sharing/reader.py
```diff
@@ -146,11 +150,20 @@ def __to_pandas_kernel(self):
         # Close the file
         json_file.close()
 
+        if (num_files > DeltaSharingReader.FILE_NUMBER_WARNING_THRESH):
+            print("Warning: Queries with a large number of files take a large ammount of time")
```
pranavsuku-db, line 154: @linzhou-db  is there a better way to display the warning to the user?

linzhou-db: as discussed offline, we don't need this.

linzhou-db (approved):
LGTM, let's remove the warning.

PR #536, opened 2024-07-26
load_table_as_pandas() allows user to determine whether or not to use kernel

pranavsuku-db:
This can lead to a 4x speedup of reading tables even without DV/CM support


python/delta_sharing/delta_sharing.py
```diff
@@ -104,6 +104,7 @@ def load_as_pandas(
     version: Optional[int] = None,
     timestamp: Optional[str] = None,
     jsonPredicateHints: Optional[str] = None,
+    useKernel: Optional[bool] = None,
```
pranavsuku-db, line 107: @linzhou-db I was not sure whether to name this as use_kernel or useKernel. the rest of variables in python are not in camelCase but jsonPredicateHints is. Should I make both of them use the normal python_underscore_naming_convention?

python/delta_sharing/reader.py
```diff
@@ -172,11 +174,19 @@ def __to_pandas_kernel(self):
         return result
 
     def to_pandas(self) -> pd.DataFrame:
-        response_format = self._rest_client.autoresolve_query_format(self._table)
+        response_format = ""
+        # If client indicates to use delta format use it, otherwise autoresolve the format
+        if (self._useKernel is not None):
```
linzhou-db, line 179: nit: does this work? `if (self._user_delta_format is not None and self._use_delta_format):`

python/delta_sharing/delta_sharing.py
linzhou-db: jsonPredicataHints is considered a full string. so let's use the underscore naming. 

though, I'm thinking of calling it `use_delta_format`

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -710,6 +710,10 @@ def test_load_as_pandas_success_dv_and_cm(
     expected['partition_col'] = expected['partition_col'].astype('int32')
     pd.testing.assert_frame_equal(pdf, expected)
 
+    # Test client specifying explicit delta format
+    pdf = load_as_pandas(f"{profile_path}#{fragments}", limit, version, None, None, True)
```
linzhou-db, line 714: did you have output to ensure that it is using the new code path?

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -912,6 +949,33 @@ def test_load_as_pandas_exception_kernel(
...
+        pytest.param(
+            "share8.default.table_with_cm_id",
+            None,
+            None,
+            "Unsupported Delta Table Properties",
+            id="column mapping id not supported",
+        ),
+    ],
+)
+def test_load_as_pandas_exception_client_delta_kernel_disabled_with_delta_table(
```
linzhou-db, line 965: nice test coverage.

python/delta_sharing/tests/test_delta_sharing.py
pranavsuku-db: yup i logged it and it is using the new path

linzhou-db: o, and could you also hack the profile file and test against our production server to verify the server respects this config?

python/delta_sharing/reader.py
pranavsuku-db: yeah None is false in Pyhton so i can actually just do: `if self._useKernel`

linzhou-db (approved):
This change looks good! 

Though let's test the production behavior before merging.

PR #537, opened 2024-07-29
Release Updates

pranavsuku-db:
pr for new release process


.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
+name: Build wheels
```
linzhou-db, line 1: let's rename the file, and change this name to be more meaningful.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,74 @@
...
+
+on:
+  push:
+    branches:
```
linzhou-db, line 5: and see if we can have more conditions in this.

PR #543, opened 2024-08-01
add openssl as dependencies in Cargo

linzhou-db:
add openssl as dependencies in Cargo, so that it can support cross release on linux arm64


pranavsuku-db (approved):
lgtm!

PR #545, opened 2024-08-01
Fix typos and formatting in protocol.md

c27kwan:
Exactly as titled: Fix typos in the protocol doc.


PROTOCOL.md
```diff
@@ -2460,11 +2460,11 @@ responses to be compatible with legacy servers.
 ### responseFormat
 Indicates the format to expect in the [API Response Format in Parquet](#api-response-format-in-parquet), two values are supported.
```
c27kwan, line 2461: Is this still valid? It seems confusing to talk about the format being either `parquet` or `delta` for API Response Format in Parquet.

PR #553, opened 2024-08-05
added support for OAuth Client Credentials grant in the Scala client

moderakh:
**Added OAuth Client Credentials Support in Scala Client**
Overview:
This PR introduces support for OAuth Client Credentials in the Scala client. 

Adds a CredentialProvider trait and two implementations: 
* BearerTokenAuthProvider and OAuthClientCredentialsAuthProvider.

`CredentialProvider` Trait:
Added the CredentialProvider trait which includes a method CredentialProvider#addAuthHeader(httpRequest: HttpRequestBase). This method is responsible for adding the authentication header to the HTTP request.
Implementations of CredentialProvider:

`BearerTokenAuthProvider`: A simple implementation that uses a provided bearer token.

`OAuthClientCredentialsAuthProvider`: This implementation handles OAuth client credential grants. It manages the access token internally and refreshes it when it is about to expire (within the next 10 minutes), using the clientId and clientSecret to obtain a new token.
Functionality:

The DeltaShringClient now invokes credentialProvider.addAuthHeader() before making any requests, ensuring that the appropriate authentication header is included.
OAuth Token Management:

The OAuthClientCredentialsAuthProvider efficiently manages the OAuth access token. It only exchanges the clientId and clientSecret for a new access token when the current token is nearing expiration

New format for the profile share file:
```
{
    "shareCredentialsVersion": 2,
    "type": "persistent_oauth2.0",
    "endpoint": "https://example.com/delta-sharing/",
    "tokenEndpoint": "https://example.com/oauth/v1/token",
    "clientId": "ABCE",
    "clientSecret": "XYZE",
    "scope" : "abc" // optional parameter
}

```

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -1213,8 +1213,11 @@ object DeltaSharingRestClient extends Logging {
         classOf[Int],
         classOf[Boolean],
         classOf[Long],
-        classOf[Long]
-      ).newInstance(profileProvider,
+        classOf[Long],
+        classOf[Int],
+        classOf[Int],
+        classOf[Int]
+    ).newInstance(profileProvider,
         java.lang.Integer.valueOf(timeoutInSeconds),
         java.lang.Integer.valueOf(numRetries),
         java.lang.Long.valueOf(maxRetryDurationMillis),
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -40,7 +40,7 @@ import org.apache.http.impl.client.{HttpClientBuilder, HttpClients}
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.SparkSession
 
-import io.delta.sharing.client.auth.AuthCredentialProviderFactory
+import io.delta.sharing.client.auth.{AuthConfig, AuthCredentialProviderFactory}
 import io.delta.sharing.client.model._
 import io.delta.sharing.client.util.{ConfUtils, JsonUtils, RetryUtils, UnexpectedHttpStatus}
 
@@ -164,7 +164,10 @@ class DeltaSharingRestClient(
     maxFilesPerReq: Int = 100000,
     enableAsyncQuery: Boolean = false,
     asyncQueryPollIntervalMillis: Long = 10000L,
-    asyncQueryMaxDuration: Long = 600000L
+    asyncQueryMaxDuration: Long = 600000L,
+    tokenExchangeMaxRetries: Int = 5,
+    tokenExchangeMaxRetryDurationInSeconds: Int = 60,
+    tokenRenewalThresholdInSeconds: Int = 600
   ) extends DeltaSharingClient with Logging {
 
   logInfo(s"DeltaSharingRestClient with enableAsyncQuery $enableAsyncQuery")
@@ -206,6 +209,8 @@ class DeltaSharingRestClient(
 
   private lazy val authCredentialProvider = AuthCredentialProviderFactory.createCredentialProvider(
     profileProvider.getProfile,
+    AuthConfig(tokenExchangeMaxRetries,
+      tokenExchangeMaxRetryDurationInSeconds, tokenRenewalThresholdInSeconds),
     client
   )
 
@@ -1188,6 +1193,11 @@ object DeltaSharingRestClient extends Logging {
     val asyncQueryMaxDurationMillis = ConfUtils.asyncQueryTimeout(sqlConf)
     val asyncQueryPollDurationMillis = ConfUtils.asyncQueryPollIntervalMillis(sqlConf)
 
+    val tokenExchangeMaxRetries = ConfUtils.tokenExchangeMaxRetries(sqlConf)
+    val tokenExchangeMaxRetryDurationInSeconds =
+      ConfUtils.tokenExchangeMaxRetryDurationInSeconds(sqlConf)
+    val tokenRenewalThresholdInSeconds = ConfUtils.tokenRenewalThresholdInSeconds(sqlConf)
+
     val clientClass = ConfUtils.clientClass(sqlConf)
     Class.forName(clientClass)
       .getConstructor(
@@ -1216,7 +1226,10 @@ object DeltaSharingRestClient extends Logging {
         java.lang.Integer.valueOf(maxFilesPerReq),
         java.lang.Boolean.valueOf(useAsyncQuery),
         java.lang.Long.valueOf(asyncQueryPollDurationMillis),
-        java.lang.Long.valueOf(asyncQueryMaxDurationMillis)
+        java.lang.Long.valueOf(asyncQueryMaxDurationMillis),
+        java.lang.Integer.valueOf(tokenExchangeMaxRetries),
+        java.lang.Integer.valueOf(tokenExchangeMaxRetryDurationInSeconds),
+        java.lang.Integer.valueOf(tokenRenewalThresholdInSeconds)
       ).asInstanceOf[DeltaSharingClient]
   }
 }
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -204,7 +204,7 @@ class DeltaSharingRestClient(
     client
   }
 
-  private lazy val credentialProvider = AuthCredentialProviderFactory.createCredentialProvider(
+  private lazy val authCredentialProvider = AuthCredentialProviderFactory.createCredentialProvider(
     profileProvider.getProfile,
     client
   )
@@ -936,19 +936,8 @@ class DeltaSharingRestClient(
     new HttpHost(url.getHost, port, protocol)
   }
 
-  private def tokenExpired(profile: DeltaSharingProfile): Boolean = {
-    profile match {
-      case bt @ BearerTokenDeltaSharingProfile(_, _, _, expirationTime) =>
-        if (expirationTime == null) return false
-        try {
-          val expirationTime = Timestamp.valueOf(
-            LocalDateTime.parse(bt.expirationTime, ISO_DATE_TIME))
-          expirationTime.before(Timestamp.valueOf(LocalDateTime.now()))
-        } catch {
-          case _: Throwable => false
-        }
-      case _ => false
-    }
+  private def tokenExpired(): Boolean = {
+    authCredentialProvider.isExpired()
   }
 
   private[client] def prepareHeaders(httpRequest: HttpRequestBase): HttpRequestBase = {
@@ -965,7 +954,7 @@ class DeltaSharingRestClient(
       DELTA_SHARING_CAPABILITIES_HEADER -> getDeltaSharingCapabilities()
     ) ++ customeHeaders
     headers.foreach(header => httpRequest.setHeader(header._1, header._2))
-    credentialProvider.addAuthHeader(httpRequest)
+    authCredentialProvider.addAuthHeader(httpRequest)
 
     httpRequest
   }
@@ -1031,9 +1020,9 @@ class DeltaSharingRestClient(
         if (!(statusCode == HttpStatus.SC_OK ||
           (allowNoContent && statusCode == HttpStatus.SC_NO_CONTENT))) {
           var additionalErrorInfo = ""
-          if (statusCode == HttpStatus.SC_UNAUTHORIZED && tokenExpired(profile)) {
+          if (statusCode == HttpStatus.SC_UNAUTHORIZED && tokenExpired()) {
             additionalErrorInfo = s"It may be caused by an expired token as it has expired " +
-              s"at ${profile.asInstanceOf[BearerTokenDeltaSharingProfile].expirationTime}"
+              s"at ${authCredentialProvider.getExpirationTime()}"
           }
           // Only show the last 100 lines in the error to keep it contained.
           val responseToShow = lines.drop(lines.size - 100).mkString("\n")
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -29,6 +29,7 @@ import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 import org.apache.commons.io.IOUtils
 import org.apache.commons.io.input.BoundedInputStream
 import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.security.alias.CredentialProviderFactory
 import org.apache.hadoop.util.VersionInfo
 import org.apache.http.{HttpHeaders, HttpHost, HttpStatus}
 import org.apache.http.client.config.RequestConfig
@@ -41,6 +42,7 @@ import org.apache.spark.internal.Logging
 import org.apache.spark.sql.SparkSession
 
 import io.delta.sharing.client.model._
+import io.delta.sharing.client.auth.AuthCredentialProviderFactory
 import io.delta.sharing.client.util.{ConfUtils, JsonUtils, RetryUtils, UnexpectedHttpStatus}
 
 /** An interface to fetch Delta metadata from remote server. */
@@ -202,6 +204,11 @@ class DeltaSharingRestClient(
     created = true
     client
   }
+
+  private lazy val credentialProvider = AuthCredentialProviderFactory.createCredentialProvider(
+    profileProvider.getProfile,
+    client
+  )
 
   override def getProfileProvider: DeltaSharingProfileProvider = profileProvider
 
@@ -931,13 +938,18 @@ class DeltaSharingRestClient(
   }
 
   private def tokenExpired(profile: DeltaSharingProfile): Boolean = {
-    if (profile.expirationTime == null) return false
-    try {
-      val expirationTime = Timestamp.valueOf(
-        LocalDateTime.parse(profile.expirationTime, ISO_DATE_TIME))
-      expirationTime.before(Timestamp.valueOf(LocalDateTime.now()))
-    } catch {
-      case _: Throwable => false
+    profile match {
+      case bt @  BearerTokenDeltaSharingProfile(shareCredentialsVersion, endpoint, bearerToken, expirationTime) => {
+        if (expirationTime == null) return false
+        try {
+          val expirationTime = Timestamp.valueOf(
+            LocalDateTime.parse(bt.expirationTime, ISO_DATE_TIME))
+          expirationTime.before(Timestamp.valueOf(LocalDateTime.now()))
+        } catch {
+          case _: Throwable => false
+        }
+      }
+      case _ => false
     }
   }
 
@@ -951,11 +963,11 @@ class DeltaSharingRestClient(
       )
     }
     val headers = Map(
-      HttpHeaders.AUTHORIZATION -> s"Bearer ${profileProvider.getProfile.bearerToken}",
       HttpHeaders.USER_AGENT -> getUserAgent(),
       DELTA_SHARING_CAPABILITIES_HEADER -> getDeltaSharingCapabilities()
     ) ++ customeHeaders
     headers.foreach(header => httpRequest.setHeader(header._1, header._2))
+    credentialProvider.addAuthHeader(httpRequest)
 
     httpRequest
   }
@@ -1023,7 +1035,7 @@ class DeltaSharingRestClient(
           var additionalErrorInfo = ""
           if (statusCode == HttpStatus.SC_UNAUTHORIZED && tokenExpired(profile)) {
             additionalErrorInfo = s"It may be caused by an expired token as it has expired " +
-              s"at ${profile.expirationTime}"
+              s"at ${profile.asInstanceOf[BearerTokenDeltaSharingProfile].expirationTime}"
           }
           // Only show the last 100 lines in the error to keep it contained.
           val responseToShow = lines.drop(lines.size - 100).mkString("\n")
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -29,7 +29,6 @@ import scala.collection.mutable.{ArrayBuffer, ListBuffer}
 import org.apache.commons.io.IOUtils
 import org.apache.commons.io.input.BoundedInputStream
 import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.security.alias.CredentialProviderFactory
 import org.apache.hadoop.util.VersionInfo
 import org.apache.http.{HttpHeaders, HttpHost, HttpStatus}
 import org.apache.http.client.config.RequestConfig
@@ -41,8 +40,8 @@ import org.apache.http.impl.client.{HttpClientBuilder, HttpClients}
 import org.apache.spark.internal.Logging
 import org.apache.spark.sql.SparkSession
 
-import io.delta.sharing.client.model._
 import io.delta.sharing.client.auth.AuthCredentialProviderFactory
+import io.delta.sharing.client.model._
 import io.delta.sharing.client.util.{ConfUtils, JsonUtils, RetryUtils, UnexpectedHttpStatus}
 
 /** An interface to fetch Delta metadata from remote server. */
@@ -939,7 +938,7 @@ class DeltaSharingRestClient(
 
   private def tokenExpired(profile: DeltaSharingProfile): Boolean = {
     profile match {
-      case bt @  BearerTokenDeltaSharingProfile(shareCredentialsVersion, endpoint, bearerToken, expirationTime) => {
+      case bt @ BearerTokenDeltaSharingProfile(_, _, _, expirationTime) =>
         if (expirationTime == null) return false
         try {
           val expirationTime = Timestamp.valueOf(
@@ -948,7 +947,6 @@ class DeltaSharingRestClient(
         } catch {
           case _: Throwable => false
         }
-      }
       case _ => false
     }
   }
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
@@ -22,7 +22,7 @@ import java.util.Locale
 import com.fasterxml.jackson.core.JsonParser
 import com.fasterxml.jackson.databind.{DeserializationContext, JsonDeserializer}
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
-import com.fasterxml.jackson.databind.node.ObjectNode
+import com.fasterxml.jackson.databind.node.{BaseJsonNode, ObjectNode}
 import org.apache.commons.io.IOUtils
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.Path
@@ -37,21 +37,21 @@ sealed trait DeltaSharingProfile {
   val endpoint: String
 
 
-  private [client] def validate(profile: DeltaSharingProfile): Unit = {
-    if (profile.shareCredentialsVersion.isEmpty) {
+  private [client] def validate(): Unit = {
+    if (shareCredentialsVersion.isEmpty) {
       throw new IllegalArgumentException(
         "Cannot find the 'shareCredentialsVersion' field in the profile file")
     }
 
-    if (profile.shareCredentialsVersion.get > DeltaSharingProfile.CURRENT) {
+    if (shareCredentialsVersion.get > DeltaSharingProfile.CURRENT) {
       throw new IllegalArgumentException(
         s"'shareCredentialsVersion' in the profile is " +
-          s"${profile.shareCredentialsVersion.get} which is too new. The current release " +
+          s"${shareCredentialsVersion.get} which is too new. The current release " +
           s"supports version ${DeltaSharingProfile.CURRENT} and below. Please upgrade to a newer " +
           s"release.")
     }
 
-    validateNotNullAndEmpty(profile.endpoint, "endpoint")
+    validateNotNullAndEmpty(endpoint, "endpoint")
   }
 
 }
@@ -61,7 +61,10 @@ case class BearerTokenDeltaSharingProfile(override val shareCredentialsVersion: 
                                           bearerToken: String = null,
                                           expirationTime: String = null
                                          ) extends DeltaSharingProfile {
-  super.validate(this)
+  super.validate()
+  if (shareCredentialsVersion.get != 1) {
+    throw new IllegalArgumentException("BearerTokenDeltaSharingProfile only supports version 1")
+  }
   DeltaSharingProfile.validateNotNullAndEmpty(bearerToken, "bearerToken")
 }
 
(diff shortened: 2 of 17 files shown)


client/src/main/scala/io/delta/sharing/client/auth/CredentialProvider.scala
```diff
@@ -0,0 +1,91 @@
...
+    if (currentToken.isEmpty || needsRefresh(currentToken.get)) {
+      refreshToken()
+    }
+
+    // assert currentToken is not empty
+    httpRequest.setHeader(HttpHeaders.AUTHORIZATION, s"Bearer ${currentToken.get.accessToken}")
+  }
+
+  // Method to set the current token for testing purposes
+  private[auth] def setCurrentToken(token: OAuthClientCredentials): Unit = {
```
chakankardb: Nit: call this setCurrentTokenForTesting

client/src/main/scala/io/delta/sharing/client/auth/CredentialProvider.scala
```diff
@@ -0,0 +1,91 @@
...
+                                   profile: OAuthClientCredentialsDeltaSharingProfile)
+  extends CredentialProvider {
+
+  private[auth] lazy val oauthClient = new OAuthClient(client,
+    profile.tokenEndpoint, profile.clientId, profile.clientSecret, profile.scope)
+
+  private var currentToken: Option[OAuthClientCredentials] = None
+
+  override def addAuthHeader(httpRequest: HttpRequestBase): Unit = {
+    if (currentToken.isEmpty || needsRefresh(currentToken.get)) {
```
chakankardb: Nit: refreshToken already checks for this?

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
```diff
@@ -0,0 +1,95 @@
...
+                                  clientSecret: String,
+                                  scope: Option[String] = None) {
+
+  def clientCredentials(): OAuthClientCredentials = {
+    val credentials = Base64.getEncoder.encodeToString(s"$clientId:$clientSecret".getBytes("UTF-8"))
+
+    val post = new HttpPost(tokenEndpoint)
+    post.setHeader("accept", "application/json")
+    post.setHeader("authorization", s"Basic $credentials")
+    post.setHeader("content-type", "application/x-www-form-urlencoded")
```
chakankardb, line 47: Is there a doc we can link to that explains why these headers are needed?

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    val scopeParam = scope.map(s => s"&scope=$s").getOrElse("")
+    val body = s"grant_type=client_credentials$scopeParam"
+    post.setEntity(new StringEntity(body))
+
+    var response: CloseableHttpResponse = null
+    try {
+      response = httpClient.execute(post)
+      val responseString = getResponseAsString(response.getEntity)
+      if (response.getStatusLine.getStatusCode != 200 || responseString == null) {
```
chakankardb, line 57: Nit: responseString is already validated in the parse*

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
```diff
@@ -0,0 +1,95 @@
...
+        throw new RuntimeException(s"Failed to get OAuth token from token endpoint: " +
+          s"Token Endpoint responded: ${response.getStatusLine} with response: $responseString")
+      }
+
+      parseOAuthTokenResponse(responseString)
+
+    } finally {
+      if (response != null) response.close()
+    }
```
chakankardb, line 66: Question: where are retries handled? Consider adding a comment.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -931,13 +937,17 @@ class DeltaSharingRestClient(
...
+      case bt @ BearerTokenDeltaSharingProfile(_, _, _, expirationTime) =>
+        if (expirationTime == null) return false
+        try {
+          val expirationTime = Timestamp.valueOf(
+            LocalDateTime.parse(bt.expirationTime, ISO_DATE_TIME))
+          expirationTime.before(Timestamp.valueOf(LocalDateTime.now()))
+        } catch {
+          case _: Throwable => false
+        }
+      case _ => false
```
linzhou-db, line 950: shall we add a comment on what are other cases, and why it's always false for them.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1023,7 +1033,7 @@ class DeltaSharingRestClient(
           var additionalErrorInfo = ""
           if (statusCode == HttpStatus.SC_UNAUTHORIZED && tokenExpired(profile)) {
             additionalErrorInfo = s"It may be caused by an expired token as it has expired " +
-              s"at ${profile.expirationTime}"
+              s"at ${profile.asInstanceOf[BearerTokenDeltaSharingProfile].expirationTime}"
```
linzhou-db, line 1036: why is it safe to call asInstanceOf here?

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
-    bearerToken: String = null,
-    expirationTime: String = null)
+@JsonDeserialize(using = classOf[DeltaSharingProfileDeserializer])
+sealed trait DeltaSharingProfile {
+  val shareCredentialsVersion: Option[Int]
+  val endpoint: String
+}
+
+case class BearerTokenDeltaSharingProfile(override val shareCredentialsVersion: Option[Int],
+                                          override val endpoint: String = null,
```
linzhou-db, line 40: nit: fix the indent here and else where.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+      OAuthClientCredentialsDeltaSharingProfile(
+        shareCredentialsVersion = Option(node.get("shareCredentialsVersion")).map(_.asInt()),
+        endpoint = Option(node.get("endpoint")).map(_.asText()).orNull,
+        tokenEndpoint = Option(node.get("tokenEndpoint")).map(_.asText()).orNull,
+        clientId = Option(node.get("clientId")).map(_.asText()).orNull,
+        clientSecret = Option(node.get("clientSecret")).map(_.asText()).orNull,
+        scope = Option(node.get("scope")).map(_.asText())
+      )
+
+    } else {
```
linzhou-db, line 100: could there be other cases?

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+      endpoint, bearerToken, expirationTime)
+  }
+
+  private[client] def validate(profile: DeltaSharingProfile): Unit = {
+    if (profile.shareCredentialsVersion.isEmpty) {
+      throw new IllegalArgumentException(
+        "Cannot find the 'shareCredentialsVersion' field in the profile file")
+    }
+
+    if (profile.shareCredentialsVersion.get > DeltaSharingProfile.CURRENT) {
```
linzhou-db, line 131: do we need to check that the version is 1 for `BearerTokenDeltaSharingProfile`?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -203,6 +204,11 @@ class DeltaSharingRestClient(
     client
   }
 
+  private lazy val credentialProvider = AuthCredentialProviderFactory.createCredentialProvider(
```
linzhou-db, line 207: do we have to use a `CredentialProvider` class? what if we just define a `addAuthHeader` function from each profile class?

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+case class
+OAuthClientCredentialsDeltaSharingProfile(override
+                                          val shareCredentialsVersion: Option[Int],
+                                          override
+                                          val endpoint: String = null,
+                                          tokenEndpoint: String = null,
+                                          clientId: String = null,
+                                          clientSecret: String = null,
+                                          scope: Option[String] = None
+                                         ) extends DeltaSharingProfile
```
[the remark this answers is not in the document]
moderakh: addressed.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+                                          bearerToken: String = null,
+                                          expirationTime: String = null
+                                         ) extends DeltaSharingProfile
+
+object BearerTokenDeltaSharingProfile {
+  private[client] def validate(profile: BearerTokenDeltaSharingProfile): Unit = {
+    if (profile.bearerToken == null) {
+      throw new IllegalArgumentException(
+        "Cannot find the 'bearerToken' field in the profile file")
+    }
```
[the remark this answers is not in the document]
moderakh: addressed for bearerToken. expiration is not mandatory field.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+  }
+}
+
 
 object DeltaSharingProfile {
-  val CURRENT = 1
+  val CURRENT = 2
+
+  // constructor to ensure backward compatibility for DeltaSharingProfile instantiation
```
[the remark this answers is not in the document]
moderakh: addressed

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+    if (profile.bearerToken == null) {
+      throw new IllegalArgumentException(
+        "Cannot find the 'bearerToken' field in the profile file")
+    }
+  }
+}
+
+object OAuthClientCredentialsDeltaSharingProfile {
+  private[client] def validate(profile: OAuthClientCredentialsDeltaSharingProfile): Unit = {
+    if (profile.tokenEndpoint == null) {
```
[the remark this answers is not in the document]
moderakh: addressed.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
+    if (profile.endpoint == null) {
+      throw new IllegalArgumentException("Cannot find the 'endpoint' field in the profile file")
+    }
+
+    profile match {
+      case bearerTokenProfile: BearerTokenDeltaSharingProfile =>
+        BearerTokenDeltaSharingProfile.validate(bearerTokenProfile)
+      case oauthProfile: OAuthClientCredentialsDeltaSharingProfile =>
+        OAuthClientCredentialsDeltaSharingProfile.validate(oauthProfile)
+    }
```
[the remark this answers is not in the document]
moderakh: addressed

client/src/main/scala/io/delta/sharing/client/auth/CredentialProvider.scala
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  // Method to set the current token for testing purposes
+  private[auth] def setCurrentToken(token: OAuthClientCredentials): Unit = {
+    synchronized {
+      currentToken = Some(token)
+    }
+  }
+
+  private def refreshToken(): Unit = {
```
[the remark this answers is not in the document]
moderakh: addressed

client/src/main/scala/io/delta/sharing/client/auth/CredentialProvider.scala
moderakh: This is intentional to account for multi-threading before acquiring the synchronization lock. This ensures that if another thread refreshes the token before this thread acquires the synchronization lock, this thread will reuse the refreshed token.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
moderakh: added link to the spec doc

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
moderakh: That approach could work. However, I would argue that the role of a profile should primarily be to serve as a container for configuration rather than to implement the authentication logic. Moreover, if we need to introduce additional authentication-related methods in the future, the profile class could become bloated.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -71,24 +187,9 @@ private[sharing] class DeltaSharingFileProfileProvider(
...
-          s"supports version ${DeltaSharingProfile.CURRENT} and below. Please upgrade to a newer " +
-          s"release.")
-    }
-    if (profile.endpoint == null) {
-      throw new IllegalArgumentException("Cannot find the 'endpoint' field in the profile file")
-    }
-    if (profile.bearerToken == null) {
-      throw new IllegalArgumentException("Cannot find the 'bearerToken' field in the profile file")
-    }
+    DeltaSharingProfile.validate(profile)
```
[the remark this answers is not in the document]
moderakh: addressed.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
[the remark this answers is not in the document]
moderakh: addressed

moderakh: added a default throw for other cases

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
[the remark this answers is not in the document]
moderakh: I can see argument both ways. anyway moved this to a method inside `AuthCredentialProvider`

client/src/main/scala/io/delta/sharing/client/auth/CredentialProvider.scala
moderakh: addressed

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
moderakh: moved this to a method to the authProvider

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
moderakh: did abstraction using the authCredentailProvider.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
moderakh: Thanks for the comment. addressed.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,136 @@
...
 
-case class DeltaSharingProfile(
-    shareCredentialsVersion: Option[Int],
-    endpoint: String = null,
-    bearerToken: String = null,
-    expirationTime: String = null)
+@JsonDeserialize(using = classOf[DeltaSharingProfileDeserializer])
+sealed trait DeltaSharingProfile {
+  val shareCredentialsVersion: Option[Int]
+  val endpoint: String
```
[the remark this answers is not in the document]
moderakh: added type to the parent trait

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -51,7 +165,9 @@ trait DeltaSharingProfileProvider {
   // `refresher` takes an optional refreshToken, and returns
   // (idToUrlMap, minUrlExpirationTimestamp, refreshToken)
   def getCustomRefresher(
-      refresher: Option[String] => TableRefreshResult): Option[String] => TableRefreshResult = {
+                          refresher:
+                          Option[String] => TableRefreshResult):
+  Option[String] => TableRefreshResult = {
```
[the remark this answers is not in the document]
moderakh: sounds like the existing code is not happy with the style rule If I change it back to what it was before then I would get this style error:
```
[error] /Users/Moe.Derakhshani/github/delta-sharing/client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala:171: maxLineLength.message
```

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
moderakh: I think the indent is correct and style check is passing already.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
moderakh: done.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,138 @@
...
+  val endpoint: String
+  val `type` : String
+
+  private [client] def validate(): Unit = {
+    if (shareCredentialsVersion.isEmpty) {
+      throw new IllegalArgumentException(
+        "Cannot find the 'shareCredentialsVersion' field in the profile file")
+    }
+
+    if (shareCredentialsVersion.get > DeltaSharingProfile.CURRENT) {
```
chakankardb, line 46: perhaps also check that it is >= 1?

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,138 @@
...
 import io.delta.sharing.client.util.JsonUtils
 
-case class DeltaSharingProfile(
-    shareCredentialsVersion: Option[Int],
-    endpoint: String = null,
+@JsonDeserialize(using = classOf[DeltaSharingProfileDeserializer])
+sealed trait DeltaSharingProfile {
+  val shareCredentialsVersion: Option[Int]
+  val endpoint: String
+  val `type` : String
```
chakankardb, line 38: call this profileType (to avoid backticks)?

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -17,22 +17,138 @@
...
+  }
+}
+
+case class BearerTokenDeltaSharingProfile(
+    override val shareCredentialsVersion: Option[Int],
+    override val endpoint: String = null,
     bearerToken: String = null,
     expirationTime: String = null)
+  extends DeltaSharingProfile {
+  override val `type`: String = "bearer_token"
```
chakankardb, line 64: consider defining these as constants: BEARER_TOKEN

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
```diff
@@ -0,0 +1,108 @@
...
+      EntityUtils.toString(httpEntity)
+    } else {
+      null
+    }
+  }
+}
+
+private[auth] object OAuthClient {
+  val numRetries = 3
+  val maxRetryDurationInMillis = 30000 // 30 seconds
```
chakankardb, line 107: Can we use these from a config?

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  private[auth] lazy val oauthClient = new OAuthClient(client,
+    profile.tokenEndpoint, profile.clientId, profile.clientSecret, profile.scope)
+
+  private var currentToken: Option[OAuthClientCredentials] = None
+
+  override def addAuthHeader(httpRequest: HttpRequestBase): Unit = {
+    if (currentToken.isEmpty || needsRefresh(currentToken.get)) {
+      maybeRefreshToken()
+    }
```
chakankardb, line 38: Perhaps directly call maybeRefreshToken, since these conditions are anyways checked there and this access is not synchronized so there may be concurrent access issues.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+
+private[client] case class OAuthClientCredentialsAuthProvider(
+                                           client: CloseableHttpClient,
+                                           profile: OAuthClientCredentialsDeltaSharingProfile)
+  extends AuthCredentialProvider {
+
+  private[auth] lazy val oauthClient = new OAuthClient(client,
+    profile.tokenEndpoint, profile.clientId, profile.clientSecret, profile.scope)
+
+  private var currentToken: Option[OAuthClientCredentials] = None
```
chakankardb, line 33: Perhaps add a comment that access to this should be synchronized.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  override def addAuthHeader(httpRequest: HttpRequestBase): Unit = {
+    if (currentToken.isEmpty || needsRefresh(currentToken.get)) {
+      maybeRefreshToken()
+    }
+
+    httpRequest.setHeader(HttpHeaders.AUTHORIZATION, s"Bearer ${currentToken.get.accessToken}")
```
chakankardb, line 40: Do we need to protect the get with lock?
Perhaps maybeRefreshToken can return the token? 
Maybe we can call it mayRefreshAndGetToken?

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+  private def maybeRefreshToken(): Unit = {
+    synchronized {
+      if (currentToken.isEmpty || needsRefresh(currentToken.get)) {
+        val newToken = oauthClient.clientCredentials()
+        currentToken = Some(newToken)
+      }
+    }
+  }
+
+  private[auth] def needsRefresh(token: OAuthClientCredentials): Boolean = {
```
chakankardb, line 59: Perhaps add a comment that this needs to be called with the lock held.
Perhaps call this needsRefreshLocked?

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+    val expirationTime = token.creationTimestamp + token.expiresIn * 1000
+    expirationTime - now < OAuthClientCredentialsAuthProvider.expirationThresholdInSeconds * 1000
+  }
+
+  override def getExpirationTime(): Option[String] = None
+}
+
+private[client] object OAuthClientCredentialsAuthProvider {
+  // We will refresh the token if it expires in less than 10 minutes.
+  private val expirationThresholdInSeconds = 60 * 10 // 10 minutes
```
chakankardb, line 70: Question: Typically how long is the token valid? Is 10 mins too high (consider putting this in a conf)? What if the token itself it valid of <= 10 mins?

One thought is to do something like: expirationTime - now <  min(K mins, 0.5 * tokenExpiresIn)

moderakh: usually access-tokens are valid for at least 1 hour or more.

linzhou-db (commented):
Almost looking good, I still need some time to read OAuth* changes.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
moderakh: We already have a check for this in the sublcass please see here:
https://github.com/delta-io/delta-sharing/pull/553/files#diff-41000a1f5b33604158b33b25b259a9892ecc0ed468733bc4a71704e89b91fe56R68

and here https://github.com/delta-io/delta-sharing/pull/553/files#diff-41000a1f5b33604158b33b25b259a9892ecc0ed468733bc4a71704e89b91fe56R86

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
moderakh: sure, renamed.

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
moderakh: sure. added constant for that.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClient.scala
moderakh: sure, I added new configs.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
```diff
@@ -0,0 +1,71 @@
...
+    val now = System.currentTimeMillis()
+    val expirationTime = token.creationTimestamp + token.expiresIn * 1000
+    expirationTime - now < OAuthClientCredentialsAuthProvider.expirationThresholdInSeconds * 1000
+  }
+
+  override def getExpirationTime(): Option[String] = None
+}
+
+private[client] object OAuthClientCredentialsAuthProvider {
+  // We will refresh the token if it expires in less than 10 minutes.
```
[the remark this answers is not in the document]
moderakh: updated comment.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
moderakh: I moved the conf to the config. I would rather we do not make this heuristic.
tokens are typically valid for at least 1 hour. As this is configurable now the end user can change the threshold if they want.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
moderakh: this does not need to be invoked with lock held.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
moderakh: no, we do not need to hold the lock here.

moderakh: The synchronization logic in the PR with [Double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) was correct. However for better readability I changed that to use a reentrant lock. That should improve readability.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
moderakh: The synchronization logic in the PR with [Double-checked locking](https://en.wikipedia.org/wiki/Double-checked_locking) was correct. However for better readability I changed that to use a reentrant lock. That should improve readability.

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
chakankardb: Yes, you are right.

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileProfileProviderSuite.scala
```diff
@@ -255,7 +255,7 @@ class DeltaSharingFileProfileProviderSuite extends SparkFunSuite {
       scope = Some("testScope")
     )
 
-    assert(profile.`type` == "oauth_client_credentials")
+    assert(profile.`profileType` == "oauth_client_credentials")
```
chakankardb, line 258: Are these backticks still needed?

client/src/main/scala/io/delta/sharing/client/auth/OAuthClientCredentialsAuthProvider.scala
chakankardb: sg

chakankardb (approved):

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileProfileProviderSuite.scala
moderakh: no. thanks for catching. addressed.

client/src/main/scala/io/delta/sharing/client/auth/AuthCredentialProviderFactory.scala
```diff
@@ -0,0 +1,38 @@
...
+
+private[client] object AuthCredentialProviderFactory {
+  def createCredentialProvider(profile: DeltaSharingProfile,
+                               authConfig: AuthConfig,
+                               client: CloseableHttpClient): AuthCredentialProvider = {
+    profile match {
+      case oauthProfile: OAuthClientCredentialsDeltaSharingProfile =>
+        OAuthClientCredentialsAuthProvider(client, authConfig, oauthProfile)
+      case BearerTokenDeltaSharingProfile(_, _, bearerToken, expirationTime) =>
+        BearerTokenAuthProvider(bearerToken, expirationTime)
```
linzhou-db, line 35: nit: should we throw internal exception if it didn't match either.

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -216,6 +227,47 @@ object ConfUtils {
...
+  }
+
+  def tokenExchangeMaxRetryDurationInSeconds(conf: SQLConf): Int = {
+    val maxDur = conf.getConfString(
+      OAUTH_MAX_RETRY_DURATION_CONF, OAUTH_MAX_RETRY_DURATION_SECONDS_DEFAULT.toString).toInt
+    validatePositive(maxDur, OAUTH_MAX_RETRY_DURATION_CONF)
+    maxDur
+  }
+
+  def tokenRenewalThresholdInSeconds(conf: Configuration): Int = {
```
linzhou-db, line 257: we may not need both functions for each conf?

PR #559, opened 2024-08-08
Add logic in DeltaSharingFileSystem to conditionally downgrade https requests to http

taiga-db:
Conditionally allow https requests (to access storage buckets) in `DeltaSharingFileSystem` to be downgraded to http when a proxy is specified (introduced [in this PR](https://github.com/delta-io/delta-sharing/pull/451)) through a new spark config, `spark.delta.sharing.never.use.https`

verified using unit test, `DeltaSharingFileSystemSuite` and integration test, `table1 with storage proxy`. The latter validates that the pre-signed url file-access code path is successful when requests go through a proxy server (where the request is upgraded from http -> https).


client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
```diff
@@ -69,7 +70,31 @@ private[sharing] class DeltaSharingFileSystem extends FileSystem with Logging {
...
+              return super.execute(wrappedRequest, connection, context)
+            } catch {
+              case e: Exception =>
+                logInfo("Failed to downgrade the request to http", e)
+            }
+            super.execute(request, connection, context)
+          }
+        }
+        clientBuilder.setRequestExecutor(httpRequestDowngradeExecutor)
+      }
```
taiga-db, line 96: we allow the `neverUseHttps` case to go into the `noProxyHosts` route planner so that requests go through the path, 
```
 new HttpRoute(target, proxy)

```
where the default `secure` (see Apache docs [here](https://www.javadoc.io/doc/org.apache.httpcomponents/httpclient/4.3.3/org/apache/http/conn/routing/HttpRoute.html#HttpRoute(org.apache.http.HttpHost,%20java.net.InetAddress,%20boolean))) value is false. This prevents the tunneling (`CONNECT`) on https requests.

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
```diff
@@ -69,7 +70,31 @@ private[sharing] class DeltaSharingFileSystem extends FileSystem with Logging {
       val proxy = new HttpHost(proxyConfig.host, proxyConfig.port)
       clientBuilder.setProxy(proxy)
```
taiga-db, line 72: below is a `HttpRequestExecutor` that is conditionally (spark config) set to the client that effectively downgrades https requests to http by returning a  `wrappedRequest` with a modified uri. 
- new code path is only activated by spark config
- any failures during downgrade are swallowed and a log is emitted

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
```diff
@@ -69,7 +70,31 @@ private[sharing] class DeltaSharingFileSystem extends FileSystem with Logging {
...
 
-      if (proxyConfig.noProxyHosts.nonEmpty) {
+      val neverUseHttps = getConf.getBoolean("spark.delta.sharing.never.use.https", false)
```
wchau, line 73: So we are only doing https degrading if proxy is active? Looking at the other names of the DS confs, maybe `spark.delta.sharing.network.never.use.https`?

wchau (commented):
@linzhou-db For DS to use it, we need to release this in DS kernel, too, right?

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
taiga-db: yes, only downgrading if proxy is provided. If this is an unnecessary restriction, i can remove.

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
```diff
@@ -69,7 +70,31 @@ private[sharing] class DeltaSharingFileSystem extends FileSystem with Logging {
...
+
+              return super.execute(wrappedRequest, connection, context)
+            } catch {
+              case e: Exception =>
+                logInfo("Failed to downgrade the request to http", e)
+            }
+            super.execute(request, connection, context)
+          }
+        }
+        clientBuilder.setRequestExecutor(httpRequestDowngradeExecutor)
```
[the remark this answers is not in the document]
taiga-db: i dont think the order of which they're applied to the `clientBuilder` matters, but when the http client makes a request, the route planner first decides where it goes (to the proxy without tunneling), and then the request executor is applied where the downgrade happens  (sorry for using imprecise terms)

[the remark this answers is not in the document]
wchau: Specifically, I am wondering whether the HttpRequest URI is the HTTP proxy or the final endpoint. Hopefully the latter.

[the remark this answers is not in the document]
taiga-db: client accepts the https request, the executor is applied, and then the request (http) is sent to the proxy. 

And only the scheme is modified when the proxy receives the request. The (Storage) proxy will upgrade it before sending it to its final destination right?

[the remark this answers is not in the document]
wchau: Yes

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -89,6 +89,9 @@ object ConfUtils {
     "spark.delta.sharing.oauth.tokenRenewalThresholdInSeconds"
   val OAUTH_EXPIRATION_THRESHOLD_SECONDS_DEFAULT = 10 * 60 /* 10 mins */
 
+  val NEVER_USE_HTTPS = "spark.delta.sharing.never.use.https"
```
wchau, line 92: spark.delta.sharing.network.never.use.https?

wchau (approved):

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
taiga-db: ah okay, will update

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -89,6 +89,9 @@ object ConfUtils {
     "spark.delta.sharing.oauth.tokenRenewalThresholdInSeconds"
   val OAUTH_EXPIRATION_THRESHOLD_SECONDS_DEFAULT = 10 * 60 /* 10 mins */
 
+  val NEVER_USE_HTTPS = "spark.delta.sharing.network.never.use.https"
```
linzhou-db, line 92: nit: shall we we have a better name? or you prefer `never`?

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
```diff
@@ -69,7 +70,31 @@ private[sharing] class DeltaSharingFileSystem extends FileSystem with Logging {
       val proxy = new HttpHost(proxyConfig.host, proxyConfig.port)
       clientBuilder.setProxy(proxy)
 
-      if (proxyConfig.noProxyHosts.nonEmpty) {
+      val neverUseHttps = ConfUtils.getNeverUseHttps(getConf)
+      if (neverUseHttps) {
```
linzhou-db, line 74: Is this orthogonal with the proxy change? or could it only be applied when a proxy is set?

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
taiga-db: this naming mirrors existing settings (like [this one](https://src.dev.databricks.com/databricks/runtime@5268038c448df2bc3226fc11b80ce37319d3a9c1/-/blob/core/src/main/scala/com/databricks/io/HadoopConfUtils.scala?L61))

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
taiga-db: while it's possible to keep the logic separate from the proxy change, it's not realistic (at least from what I can tell) to  downgrade https requests to access storage buckets without one.

linzhou-db: so it could only be applied when a proxy is set?

taiga-db: yes. that is the plan

linzhou-db: could you guide me in which line is that enforced? 
And what's the code path when there's no proxy set (and it will ignore the `neverUseHttps` config)?

linzhou-db (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala
taiga-db: if proxy configs (host and port) are not defined, we do not enter [this block](https://github.com/delta-io/delta-sharing/blob/13abc8beb584a947890720aab42e5bd233741d2c/client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala#L68). This block contains all of the net-new code introduced in this PR. On top of that, the new code path for attaching the `HttpRequestExecutor` will never get executed if there's a proxy set but `neverUseHttps: false`

linzhou-db: got it, thanks!

PR #582, opened 2024-10-02
Make delta-sharing-client compatible with Spark 4.0

charlenelyu-db:
Currently delta-sharing-client is not compatible with Spark Master due to this commit
https://github.com/apache/spark/commit/a767f5cb1704075ee249169e8faf2ab3610b9dbc#diff-81eca9f7af2e9b23b13904131c3d32b0af9e9e1dcc7ddb5efba13201b00066c4 which changes the Column constructor. This PR changes `new Column(...)` to `Column(...)` to use the constructors in `object Column`, so delta-sharing-client can cross compile with both spark-3.5 and spark-4.0. 

Tested locally in sbt shell:
- **Compile with spark-3.x:** run `client/compile` without changing `build.sbt` > finished successfully
- **Compile with spark-4.0:** set `sparkVersion = "4.0.0-preview1"` in `build.sbt` > run `reload` > run `++ 2.13.11` to switch to scala 2.13 (spark-4.0 only runs on scala 2.13) > run `client/compile` > finished successfully

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
@@ -24,20 +24,33 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkException
 import org.apache.spark.delta.sharing.{CachedTableManager, TableRefreshResult}
 import org.apache.spark.internal.Logging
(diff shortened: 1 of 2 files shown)


linzhou-db (approved):

linzhou-db:
are we using delta-sharing-client 1.2 in delta, so we need to backport this to branch-1.2?

PR #591, opened 2024-10-07
Only Check includeEndStreamAction in response header when needed

linzhou-db:
Only Check includeEndStreamAction in response header when needed: 

- for sync queryTable and queryTableChanges, 
- not for async queries and getTableVersion and getTableMetadata

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -1180,8 +1180,8 @@ class DeltaSharingRestClient(
           // Only show the last 100 lines in the error to keep it contained.
           val responseToShow = lines.drop(lines.size - 100).mkString("\n")
           throw new UnexpectedHttpStatus(
-            s"HTTP request failed with status$getDsQueryIdForLogging: $status." +
-              s" $additionalErrorInfo $responseToShow",
+            s"HTTP request failed with status: $status" +
+              Seq(getDsQueryIdForLogging, additionalErrorInfo, responseToShow).mkString(" "),
             statusCode)
         }
         val capabilities = Option(
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -559,7 +559,7 @@ class DeltaSharingRestClient(
     var (filteredLines, endStreamAction) = maybeExtractEndStreamAction(lines)
     if (endStreamAction.isEmpty) {
       logWarning(
-        s"EndStreamAction is not returned in the response" + getDsQueryIdForLogging
+        s"EndStreamAction is not returned in the paginated response" + getDsQueryIdForLogging
       )
     }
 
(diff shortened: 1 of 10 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -336,7 +336,7 @@ class DeltaSharingRestClient(
     val target = getTargetUrl(
       s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/metadata" +
         s"$encodedParams")
-    val response = getNDJson(target)
+    val response = getNDJson(target, requireVersion = true, checkEndStreamActionHeader = false)
```
chakankardb, line 339: Question: do we set includeendstreamaction for metadata queries?
If do, would it be a problem if some DS server returns endStreamAction for it?

linzhou-db: makes sense, updated to not set includeendstreamaction for table version and metadata queries.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -495,7 +500,9 @@ class DeltaSharingRestClient(
       val (version, respondedFormat, lines, _) = getFilesByPage(table, target, request)
       (version, respondedFormat, lines)
     } else {
-      val response = getNDJson(target, request)
+      val response = getNDJsonPost(
+        target = target, data = request, setIncludeEndStreamAction = true
```
chakankardb, line 504: (optional, for my readability)
consider renaming: includeEndStreamAction -> endStreamActionEnabled
We can then setIncludeEndStreamAction = endStreamActionEnabled where we are setting it to true, and then at other places only test for setIncludeEndStreamAction.

linzhou-db: do you mean the renaming of the header or variable name?

linzhou-db: @chakankardb

chakankardb: The variable name @linzhou-db

linzhou-db: updated

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -871,14 +871,10 @@ class DeltaSharingRestClient(
           0L
         }
       },
-      respondedFormat = respondedFormat,
-      includeEndStreamActionHeader = includeEndStreamActionHeader,
+      respondedFormat = getRespondedFormat(capabilities),
```
chakankardb, line 874: Optional optimization:
Can getResponse itself create the capabilitesMap to avoid duplicate parsing and map creation from capabilities, since it seems the capabilities map is what is used everywhere?

chakankardb (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: updated.

PR #607, opened 2024-11-27
Upgrade delta-kernel-rs to 0.5

PatrickJin-db:
Existing python tests should be sufficient for correctness. Is any performance testing necessary?

### DIFF
--- a/python/delta-kernel-rust-sharing-wrapper/src/lib.rs
+++ b/python/delta-kernel-rust-sharing-wrapper/src/lib.rs
@@ -92,11 +92,11 @@ impl Scan {
                 delta_kernel::Error::Generic(format!("Could not get result schema: {e}"))
             })?);
         let results = self.0.execute(engine_interface.0.as_ref())?;
-        let record_batch_iter = RecordBatchIterator::new(
-            results.into_iter().map(|res| {
-                let data = res
-                    .raw_data
-                    .map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
+        let record_batches: Vec<_> = results
+            .map(|res| {
+                let scan_res = res.and_then(|res| Ok((res.full_mask(), res.raw_data?)));
+                let (mask, data) =
+                    scan_res.map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
                 let record_batch: RecordBatch = data
                     .into_any()
                     .downcast::<ArrowEngineData>()
@@ -104,20 +104,15 @@ impl Scan {
                         ArrowError::CastError("Couldn't cast to ArrowEngineData".to_string())
                     })?
                     .into();
-                if let Some(mut mask) = res.mask {
-                    let extra_rows = record_batch.num_rows() - mask.len();
-                    if extra_rows > 0 {
-                        // we need to extend the mask here in case it's too short
-                        mask.extend(std::iter::repeat(true).take(extra_rows));
-                    }
+                if let Some(mask) = mask {
                     let filtered_batch = filter_record_batch(&record_batch, &mask.into())?;
                     Ok(filtered_batch)
                 } else {
                     Ok(record_batch)
                 }
-            }),
-            result_schema,
-        );
+            })
+            .collect();
+        let record_batch_iter = RecordBatchIterator::new(record_batches, result_schema);
         Ok(PyArrowType(Box::new(record_batch_iter)))
     }
 }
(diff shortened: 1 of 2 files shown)


python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -93,23 +93,17 @@ impl Scan {
             })?);
         let results = self.0.execute(engine_interface.0.as_ref())?;
         let record_batch_iter = RecordBatchIterator::new(
-            results.into_iter().map(|res| {
-                let data = res
-                    .raw_data
-                    .map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
+            results.collect::<Vec<_>>().into_iter().map(|res| {
```
PatrickJin-db, line 96: Tried to avoid doing this but ran into lifetime issues without the `collect().into_iter()`. Though I don't think this should be a performance regression either since the `execute` returned a Vec in the previous delta-kernel-rs version so it was already copying.

zachschuermann: what's the error with just `results.map(...)`?

OussamaSaoudi-db: So the way we currently do `scan::execute` is it takes a reference to the engine, and returns an iterator with the same lifetime as the engine. They share lifetimes because you need a reference to the engine for the duration of iteration for all the processing.

One solution is to map, then collect at the very end: 
```rust
        let result_iter = self.0.execute(engine_interface.0.as_ref())?;
        let record_batches: Vec<_> = result_iter
            .map(|res| {
                let scan_res = res.and_then(|res| Ok((res.full_mask(), res.raw_data?)));
                let (mask, data) =
                    scan_res.map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
                    .into_any()
                    .downcast::<ArrowEngineData>()
                    .map_err(|_| {
                        ArrowError::CastError("Couldn't cast to ArrowEngineData".to_string())
                    })?
                    .into();
                if let Some(mask) = mask {
                    Ok(filtered_batch)
                } else {
                    Ok(record_batch)
                }
            })
            .collect();
        let record_batch_iter = RecordBatchIterator::new(record_batches, result_schema);

```

OussamaSaoudi-db: I'd encourage you to submit an issue with [delta-kernel-rs](https://github.com/delta-io/delta-kernel-rs) on the lifetime issue. I think it'll important to address eventually :)

PatrickJin-db: Thanks for the explanation and suggestion. The exact error was 
```
error: lifetime may not live long enough
   --> src/lib.rs:115:9
    |
87  |         &self,
    |         - let's call the lifetime of this reference `'1`
...
115 |         Ok(PyArrowType(Box::new(record_batch_iter)))
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'static`
    |
help: to declare that the trait object captures data from argument `self`, you can add an explicit `'_` lifetime bound
    |
89  |     ) -> DeltaPyResult<PyArrowType<Box<dyn RecordBatchReader + Send + '_>>> {
    |                                                                     ++++

error: lifetime may not live long enough
   --> src/lib.rs:115:9
    |
88  |         engine_interface: &PythonInterface,
    |                           - let's call the lifetime of this reference `'2`
...
115 |         Ok(PyArrowType(Box::new(record_batch_iter)))
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'2` must outlive `'static`
    |
help: to declare that the trait object captures data from argument `engine_interface`, you can add an explicit `'_` lifetime bound
    |
89  |     ) -> DeltaPyResult<PyArrowType<Box<dyn RecordBatchReader + Send + '_>>> {
    |                                                                     ++++

error: could not compile `delta-kernel-rust-sharing-wrapper` (lib) due to 2 previous errors
```

zachschuermann (approved):
LGTM with a follow-up to optimize that `collect` then `into_iter()`!

zachschuermann (approved):
re-stamping the last didn't seem to say stamped?

LGTM with a follow-up to optimize that `collect` then `into_iter()`!

PR #609, opened 2024-12-06
[DO NOT MERGE] Python connector kernel integration for CDF reads prototype

PatrickJin-db:
Known issues:
Timestamps may be incorrect, this is an issue with the cdf_prototype branch of delta-kernel-rs.

Tested via the following:
1. `cargo update && maturin develop` in `python/delta-kernel-rust-sharing-wrapper`
2. `pip install .` in `python`
3. Ran some `load_table_changes_as_pandas` calls, using both versions and timestamps. For more details, see the script on slack.

Added some unit and integration tests too


python/delta_sharing/reader.py
```diff
@@ -234,7 +203,146 @@ def to_pandas(self) -> pd.DataFrame:
...
+
+        min_version = start_version if start_version is not None else (10**20 - 1)
+        max_version = 0
+        version_to_actions = defaultdict(list)
+        version_to_timestamp = {}
+        
+        # Construct map from version to actions that took place in that version
+        for line in lines:
+            line_json = loads(line)
+            file = line_json["file"]
```
linzhou-db, line 264: there could be metadata in lines, that needs to be handled.
https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#metadata-in-delta-format

linzhou-db (commented):
overall looks good! 
The next steps:
1. what's the plan for unit test? is there a way to unit test without support delta format sharing in the oss server cdf?
2. what's a list of tables we want to test?
3. what's the largest table we want to test on? 
4. what's the release plan?

PR #612, opened 2024-12-11
Rust bindings for CDF reads

PatrickJin-db:
To be released as delta-kernel-rust-sharing-wrapper 0.2.0

### DIFF
--- a/python/delta-kernel-rust-sharing-wrapper/.gitignore
+++ b/python/delta-kernel-rust-sharing-wrapper/.gitignore
@@ -1,0 +1,1 @@
+Cargo.lock
--- a/python/delta-kernel-rust-sharing-wrapper/Cargo.toml
+++ b/python/delta-kernel-rust-sharing-wrapper/Cargo.toml
@@ -2,7 +2,7 @@
 name = "delta-kernel-rust-sharing-wrapper"
 edition = "2021"
 license = "Apache-2.0"
-version = "0.1.0"
+version = "0.2.0"
 
 [lib]
 name = "delta_kernel_rust_sharing_wrapper"
@@ -11,7 +11,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 arrow = { version = "53.3.0", features = ["pyarrow"] }
-delta_kernel = {version = "0.5", features = ["cloud", "default", "default-engine"]}
+delta_kernel = { version = "0.6.0", features = ["cloud", "default-engine"]}
 openssl = { version = "0.10", features = ["vendored"] }
 url = "2"
 
--- a/python/delta-kernel-rust-sharing-wrapper/src/lib.rs
+++ b/python/delta-kernel-rust-sharing-wrapper/src/lib.rs
@@ -1,37 +1,44 @@
 use std::sync::Arc;
 
 use arrow::compute::filter_record_batch;
-use arrow::datatypes::SchemaRef;
+use arrow::datatypes::SchemaRef as ArrowSchemaRef;
 use arrow::error::ArrowError;
 use arrow::pyarrow::PyArrowType;
-use delta_kernel::engine::arrow_data::ArrowEngineData;
+use arrow::record_batch::{RecordBatch, RecordBatchIterator, RecordBatchReader};
+
 use delta_kernel::engine::default::executor::tokio::TokioBackgroundExecutor;
 use delta_kernel::engine::default::DefaultEngine;
+use delta_kernel::scan::ScanResult;
+use delta_kernel::table_changes::scan::{
+    TableChangesScan as KernelTableChangesScan,
+    TableChangesScanBuilder as KernelTableChangesScanBuilder,
+};
+use delta_kernel::Error as KernelError;
+use delta_kernel::{engine::arrow_data::ArrowEngineData, schema::StructType};
+use delta_kernel::{DeltaResult, Engine};
+
 use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 
 use url::Url;
 
-use arrow::record_batch::{RecordBatch, RecordBatchIterator, RecordBatchReader};
-use delta_kernel::Engine;
-
 use std::collections::HashMap;
 
-struct KernelError(delta_kernel::Error);
+struct PyKernelError(KernelError);
 
-impl From<KernelError> for PyErr {
-    fn from(error: KernelError) -> Self {
+impl From<PyKernelError> for PyErr {
+    fn from(error: PyKernelError) -> Self {
         PyValueError::new_err(format!("Kernel error: {}", error.0))
     }
 }
 
-impl From<delta_kernel::Error> for KernelError {
-    fn from(delta_kernel_error: delta_kernel::Error) -> Self {
+impl From<KernelError> for PyKernelError {
+    fn from(delta_kernel_error: KernelError) -> Self {
         Self(delta_kernel_error)
     }
 }
 
-type DeltaPyResult<T> = std::result::Result<T, KernelError>;
+type DeltaPyResult<T> = std::result::Result<T, PyKernelError>;
 
 #[pyclass]
 struct Table(delta_kernel::Table);
@@ -40,7 +47,7 @@ struct Table(delta_kernel::Table);
 impl Table {
     #[new]
     fn new(location: &str) -> DeltaPyResult<Self> {
-        let location = Url::parse(location).map_err(delta_kernel::Error::InvalidUrl)?;
+        let location = Url::parse(location).map_err(KernelError::InvalidUrl)?;
         let table = delta_kernel::Table::new(location);
         Ok(Table(table))
     }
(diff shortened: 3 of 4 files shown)


python/delta-kernel-rust-sharing-wrapper/.gitignore
```diff
@@ -0,0 +1 @@
+Cargo.lock
```
linzhou-db, line 1: What's this for?

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -78,6 +80,38 @@ impl ScanBuilder {
     }
 }
 
+fn try_get_schema(schema: &Arc<StructType>) -> Result<SchemaRef, delta_kernel::Error> {
+    Ok(Arc::new(schema.as_ref().try_into().map_err(|e| {
+        delta_kernel::Error::Generic(format!("Could not get result schema: {e}"))
+    })?))
+}
+
+// TODO(patrick.jin): Change return type to RecordBatchIterator<impl Iterator<...>>
```
linzhou-db, line 89: should we finish this TODO before merging?

linzhou-db (commented):
I'll leave the main review to the kernel rust team

python/delta-kernel-rust-sharing-wrapper/.gitignore
PatrickJin-db: When you run `cargo` commands or `maturin develop`, a `Cargo.lock` file may be generated. Adding it to the gitignore makes sure you don't accidentally add it to git.

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
PatrickJin-db: I tried to but it seems non-trivial. But I will try to address it before the final release.

More specifically, changing the Vec to an Iterator requires the return lifetime of `Scan::execute` and `TableChangesScan::execute` to be `'_`, and that requires the return type of those functions to be `...<Arc<...>>` insteadof `...<Box<...>>`. And upon making that change, I get a `cannot be converted to a python object` error. Will see if @zachschuermann or @OussamaSaoudi-db knows more about this error.

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -78,6 +80,38 @@ impl ScanBuilder {
...
+                let filtered_batch = filter_record_batch(&record_batch, &mask.into())?;
+                Ok(filtered_batch)
+            } else {
+                Ok(record_batch)
+            }
+        })
+        .collect();
+    RecordBatchIterator::new(record_batches, result_schema)
+}
+
```
OussamaSaoudi-db, line 114:
```suggestion
fn try_create_record_batch_iter(
    results: impl Iterator<Item = DeltaResult<ScanResult>>,
    result_schema: ArrowSchemaRef,
) -> RecordBatchIterator<impl Iterator<Item = Result<RecordBatch, ArrowError>>> {
    let record_batches = results.map(|res| {
        let scan_res = res.and_then(|res| Ok((res.full_mask(), res.raw_data?)));
        let (mask, data) = scan_res.map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
        let record_batch: RecordBatch = data
            .into_any()
            .downcast::<ArrowEngineData>()
            .map_err(|_| ArrowError::CastError("Couldn't cast to ArrowEngineData".to_string()))?
            .into();
        if let Some(mask) = mask {
            let filtered_batch = filter_record_batch(&record_batch, &mask.into())?;
            Ok(filtered_batch)
        } else {
            Ok(record_batch)
        }
    });
    RecordBatchIterator::new(record_batches, result_schema)
}
```

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
OussamaSaoudi-db: Posted a fix this morning, so this is now fixed :)

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -87,38 +121,61 @@ impl Scan {
...
+            table
+                .0
+                .table_changes(engine_interface.0.as_ref(), start_version, end_version)?;
+        Ok(TableChangesScanBuilder(Some(
+            table_changes.into_scan_builder(),
+        )))
+    }
+
+    fn build(&mut self) -> DeltaPyResult<TableChangesScan> {
+        let scan = self.0.take().unwrap().build()?;
```
OussamaSaoudi-db, line 156: This will panic on `unwrap` if build is called twice.  Perhaps something like 
```rust
        let scan = self
            .0
            .take()
            .ok_or_else(|| {
                delta_kernel::Error::generic("Can only call build once on TableChangesScanBuilder")
            })?
            .build()?;
```
And you can put in a suitable error message.

OussamaSaoudi-db: Same issue with `ScanBuilder::build`

python/delta-kernel-rust-sharing-wrapper/.gitignore
zachschuermann: for further context, generally for libraries (like this) you don't want to commit whereas binaries you normally do

python/delta-kernel-rust-sharing-wrapper/Cargo.toml
```diff
@@ -11,7 +11,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 arrow = { version = "53.3.0", features = ["pyarrow"] }
-delta_kernel = {version = "0.5", features = ["cloud", "default", "default-engine"]}
+delta_kernel = { git = "https://github.com/delta-io/delta-kernel-rs.git", rev = "be1453f", features = ["cloud", "default", "default-engine"]}
```
zachschuermann, line 14: default features are empty 
```suggestion
delta_kernel = { git = "https://github.com/delta-io/delta-kernel-rs.git", rev = "be1453f", features = ["cloud", "default-engine"]}
```

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
zachschuermann: @OussamaSaoudi-db since this is a pretty common pattern I wonder if somewhere in default enging/arrow implementation we should basically provide this

zachschuermann (commented):
flushing comments

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
nicklan: Yeah, let's create an issue for it. we have it in like 3 tests and our examples, and here, so clearly it would be useful to have a canonical implementation

zachschuermann: https://github.com/delta-io/delta-kernel-rs/issues/592

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -73,11 +75,45 @@ impl ScanBuilder {
     }
 
     fn build(&mut self) -> DeltaPyResult<Scan> {
-        let scan = self.0.take().unwrap().build()?;
+        let scan = self
+            .0
+            .take()
+            .ok_or_else(|| {
+                delta_kernel::Error::generic("Can only call build() once on ScanBuilder")
```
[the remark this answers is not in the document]
OussamaSaoudi-db: similar for other `delta_kernel::.*`

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -87,38 +123,67 @@ impl Scan {
...
+            .take()
+            .ok_or_else(|| {
+                delta_kernel::Error::generic(
+                    "Can only call build() once on TableChangesScanBuilder",
+                )
+            })?
+            .build()?;
+        Ok(TableChangesScan(scan))
+    }
+}
```
OussamaSaoudi-db, line 167: Currently we fail on any subsequent call to build. If we want to avoid the potential error, you can do something like this:
```suggestion
#[pyclass]
struct TableChangesScanBuilder {
    table_changes: Arc<TableChanges>,
}

#[pymethods]
impl TableChangesScanBuilder {
    #[new]
    #[pyo3(signature = (table, engine_interface, start_version, end_version=None))]
    fn new(
        table: &Table,
        engine_interface: &PythonInterface,
        start_version: u64,
        end_version: Option<u64>,
    ) -> DeltaPyResult<TableChangesScanBuilder> {
        let table_changes = table
            .0
            .table_changes(engine_interface.0.as_ref(), start_version, end_version)?
            .into();
        Ok(TableChangesScanBuilder { table_changes })
    }

    fn build(&mut self) -> DeltaPyResult<TableChangesScan> {
        // Note: cheap Arc clone
        let scan_builder = self.table_changes.clone().scan_builder();
        let scan = scan_builder.build()?;
        Ok(TableChangesScan(scan))
    }
}
```

OussamaSaoudi-db: You can do the same thing in Snapsot/ScanBulider. @zachschuermann  would like to see what you think.

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -87,38 +123,67 @@ impl Scan {
...
+#[pyclass]
+struct TableChangesScan(delta_kernel::table_changes::scan::TableChangesScan);
+
+#[pymethods]
+impl TableChangesScan {
+    fn execute(
+        &self,
+        engine_interface: &PythonInterface,
+    ) -> DeltaPyResult<PyArrowType<Box<dyn RecordBatchReader + Send>>> {
+        let result_schema: ArrowSchemaRef = try_get_schema(self.0.schema())?;
```
OussamaSaoudi-db, line 178: This will be the same for every call to execute, so you may want to store this in `TableChangesScan`.

OussamaSaoudi-db (approved):
I've just got nits on how you may want to refactor/change the code. At a high level, this LGTM

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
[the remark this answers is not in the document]
PatrickJin-db: Done for some (but not all) `delta_kernel::*`. I think refactoring to adopt this pattern everywhere can be a followup

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
PatrickJin-db: I think if we change this behavior, we should change it for both TableChangesScanBuilder and ScanBuilder to be consistent.

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -87,50 +126,81 @@ impl Scan {
...
+        let result_schema = self.schema.clone();
+        let results = self.scan.execute(engine_interface.0.clone())?;
+        let record_batch_iter = try_create_record_batch_iter(results, result_schema);
         Ok(PyArrowType(Box::new(record_batch_iter)))
     }
 }
 
 #[pyclass]
-struct PythonInterface(Box<dyn Engine + Send>);
+struct PythonInterface(Arc<dyn Engine + Send>);
```
nicklan, line 191: what was the motivation to change from box to arc?

PatrickJin-db: It is related to [this](https://github.com/delta-io/delta-kernel-rs/pull/553) kernel change. Now we have to pass an `Arc<dyn Engine>` into `Scan::execute`/`TableChangesScan::execute`.

PR #615, opened 2024-12-18
Update Python connector version to 1.3.0

PatrickJin-db:


python/delta_sharing/version.py
```diff
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-__version__ = "1.1.0"
+__version__ = "1.3.0"
```
chakankardb, line 17: Question: why do we go from 1.1.0 to 1.3.0? why not 1.2.0?

chakankardb (approved):

python/delta_sharing/version.py
PatrickJin-db: It seems like we aren't updating the version in the main branch anymore, and only in the `python-branch-x` branches. For example, the version is 1.2.0 in python-branch-1.2.

PatrickJin-db:
Moved to #617 and #618

PR #621, opened 2024-12-19
Improve strict schema string match with smarter matching

andyl-db:
This PR improves the strict schema string match with smarter matching i.e. subset matching

Added unit test

Tested against DBR:
image

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
@@ -227,12 +227,52 @@ class RemoteSnapshot(
   }
 
   private def checkSchemaNotChange(newMetadata: Metadata): Unit = {
-    if (newMetadata.schemaString != metadata.schemaString ||
-      newMetadata.partitionColumns != metadata.partitionColumns) {
-      throw new SparkException(
-        s"""The schema or partition columns of your Delta table has changed since your
-           |DataFrame was created. Please redefine your DataFrame""")
+    val newSchema = DataType.fromJson(newMetadata.schemaString).asInstanceOf[StructType]
+    val currentSchema = DataType.fromJson(metadata.schemaString).asInstanceOf[StructType]
+
+    val newSchemaFields = newSchema.fields.map(field => {
+      field.name -> field.dataType
+    }).toMap
+
+    val currentSchemaFields = currentSchema.fields.map(field => {
+      field.name -> field.dataType
+    }).toMap
+
+    val schemaChangedException = new SparkException(
+      s"""The schema or partition columns of your Delta table has changed since your
+         |DataFrame was created. Please redefine your DataFrame""".stripMargin)
+
+    // Verify partition columns are the same
+    if (metadata.partitionColumns != newMetadata.partitionColumns) {
+      throw schemaChangedException
     }
+
+    val newSchemaFieldNames = newSchemaFields.keySet
+    val currentSchemaFieldNames = currentSchemaFields.keySet
+
+    // Ensure that all the current schema field names are a subset of new schema field names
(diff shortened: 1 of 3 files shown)


client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -227,12 +227,49 @@ class RemoteSnapshot(
   }
 
   private def checkSchemaNotChange(newMetadata: Metadata): Unit = {
-    if (newMetadata.schemaString != metadata.schemaString ||
-      newMetadata.partitionColumns != metadata.partitionColumns) {
```
charlenelyu-db, line 231: we should check the partition columns too?

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -227,12 +227,49 @@ class RemoteSnapshot(
...
+    }).toMap
+
+    val schemaChangedException = new SparkException(
+      s"""The schema or partition columns of your Delta table has changed since your
+         |DataFrame was created. Please redefine your DataFrame""".stripMargin)
+
+    val newSchemaFieldNames = newSchemaFields.keySet
+    val currentSchemaFieldNames = currentSchemaFields.keySet
+
+    // Ensure that all the new schema field names are a subset of current schema field names
```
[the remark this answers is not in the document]
charlenelyu-db: I feel we could follow this [Delta check](https://github.com/delta-io/delta/blob/55801a2bb5b5f7a555db17757451e885be5e63ec/spark/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala#L308-L308) in general. Right now I see few differences:
1. For the Delta check, they ensure current schema fields are a subset new schema fields, but we're doing the opposite.
2. They have checks on the nullability too.

I'm not familiar with the rationale behind the check either. Might worth to ask Ryan or Delta folks to make sure the change is safe.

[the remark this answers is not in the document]
andyl-db: 1. Reversed the check
2. Nullability is checked within DataType.equalsStructurally

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -227,11 +228,23 @@ class RemoteSnapshot(
...
+      s"""The schema or partition columns of your Delta table has changed since your
+         |DataFrame was created. Please redefine your DataFrame""".stripMargin)
+
+    if (ConfUtils.structuralSchemaMatchingEnabled(spark.sessionState.conf)) {
+      val newSchema = DataType.fromJson(newMetadata.schemaString).asInstanceOf[StructType]
+      val currentSchema = DataType.fromJson(metadata.schemaString).asInstanceOf[StructType]
+
+      if (
+        metadata.partitionColumns != newMetadata.partitionColumns ||
+          !SchemaUtils.isReadCompatible(currentSchema, newSchema)
```
linzhou-db, line 241: Though, reading the comment of the function and also the usage in DeltaSharingSource, it seems used with (newSchema, currentSchema)?

andyl-db: I only reversed it because I was looking at the implementation but I misread the subsetting. Changed.

andyl-db: Actually my tests started failing after making that change...I think there is an error in the current implementation.

linzhou-db (approved):

andyl-db:
Re-tested the latest changes work: 
image

PR #653, opened 2025-02-28
[DTS-4838] Optimize Retry Logic: Reduce Attempts and Adaptive Sleep Intervals

littlegrasscao:
This PR introduces the following changes to the retry logic:

1. Reduced Retry Attempts: The number of retries is reduced from 10 to 3. If the Delta Sharing server returns an error, the client will either retry 3 times or retry for a maximum of 10 minutes, whichever occurs first. This adjustment allows the recipient to fail faster.
2. Introduced retry sleep interval: A sleep interval between retries has been introduced, with a default duration of 1000 milliseconds. The sleep time would double after each retry.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
+++ b/client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
@@ -67,7 +67,7 @@ private[sharing] object RetryUtils extends Logging {
       case e: UnexpectedHttpStatus =>
         if (e.statusCode == 429) { // Too Many Requests
           true
-        } else if (e.statusCode >= 500 && e.statusCode < 600) { // Internal Error
+        } else if (e.statusCode == 503) { // Service Unavailable
           true
         } else {
           false
--- a/client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
+++ b/client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
@@ -31,9 +31,10 @@ private[sharing] object RetryUtils extends Logging {
 
   def runWithExponentialBackoff[T](
       numRetries: Int,
-      maxDurationMillis: Long = Long.MaxValue)(func: => T): T = {
+      maxDurationMillis: Long = Long.MaxValue,
+      retrySleepInterval: Long = 1000)(func: => T): T = {
     var times = 0
-    var sleepMs = 100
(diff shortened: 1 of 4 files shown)


client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
```diff
@@ -67,7 +67,7 @@ private[sharing] object RetryUtils extends Logging {
       case e: UnexpectedHttpStatus =>
         if (e.statusCode == 429) { // Too Many Requests
           true
-        } else if (e.statusCode >= 500 && e.statusCode < 600) { // Internal Error
+        } else if (e.statusCode == 503) { // Service Unavailable
```
linzhou-db, line 70: I'm not sure about this. I wonder we want to keep it retrying on 500 as we've reduced the # of retries to 3.

littlegrasscao: We could keep it as 5XX.

But my argument to change it to 503 is that:

1. Other 5XX doesn't seem to be recoverable from retry.
2. The retry logic we have in universe is more accurate based on monitoring. Since open recipients and databricks recipients get the same error sets from servers, it would be good to also use same retry logic here.

linzhou-db: the only thing is that some server side 503 could be translated as 500 -- it's possible right?

chakankardb: https://stackoverflow.com/questions/51770071/what-are-the-http-codes-to-automatically-retry-the-request

I'd say let's start with reducing the number of retries.
If we reduce the retries, perhaps we should backoff more aggressively; maybe start with 1 sec to give server a better chance to recover.
Also, do we want to limit the number retries for 429 as well?

littlegrasscao: Updated the PR to introduce a sleep interval setting, default to 1000ms. 
Kept the error codes as 5XX for now. 

How do we want to differentiate number of retries between 429 and 5XX? Right now, the retry number is a parameter passed in through config. Should we keep it consistent so that users don't get confused?

linzhou-db (approved):

chakankardb (approved):

PR #654, opened 2025-03-04
Python connector TIMESTAMP_NTZ support

PatrickJin-db:
Adds timestampntz to the capabilities with an integration test. See [this notebook](https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/2154539344732364?o=6051921418418893#command/7952259781755525) for how the table was created.

### DIFF
--- a/python/delta_sharing/rest_client.py
+++ b/python/delta_sharing/rest_client.py
@@ -143,7 +143,8 @@ class DataSharingRestClient:
     USER_AGENT: ClassVar[str] = _client_user_agent()
     DELTA_RESPONSE_FORMAT = "responseformat=delta"
     DELTA_AND_PARQUET_RESPONSE_FORMAT = "responseformat=delta,parquet"
-    DELTA_READER_FEATURES = "readerfeatures=deletionvectors,columnmapping"
+    DELTA_SNAPSHOT_READER_FEATURES = "readerfeatures=deletionvectors,columnmapping,timestampntz"
+    DELTA_CDF_READER_FEATURES = "readerfeatures=deletionvectors,columnmapping"
     CAPABILITIES_HEADER = "delta-sharing-capabilities"
     DELTA_TABLE_VERSION_HEADER = "delta-table-version"
     DELTA_FORMAT = "delta"
@@ -168,10 +169,16 @@ class DataSharingRestClient:
         if urlparse(profile.endpoint).hostname == "localhost":
             self._session.verify = False
 
+    def __get_reader_features(self, for_cdf):
+        if for_cdf:
+            return DataSharingRestClient.DELTA_CDF_READER_FEATURES
+        else:
+            return DataSharingRestClient.DELTA_SNAPSHOT_READER_FEATURES
+
     def set_sharing_capabilities_header(self):
         delta_sharing_capabilities = (
             DataSharingRestClient.DELTA_AND_PARQUET_RESPONSE_FORMAT + ';' +
-            DataSharingRestClient.DELTA_READER_FEATURES
+            self.__get_reader_features(False)
         )
         self._session.headers.update(
             {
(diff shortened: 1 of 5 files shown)


python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -113,7 +114,7 @@ def _verify_all_tables_result(tables: Sequence[Table]):
         Table(name="table_with_cm_name", share="share8", schema="default"),
         Table(name="table_with_cm_id", share="share8", schema="default"),
         Table(name="deletion_vectors_with_dvs_dv_property_on", share="share8", schema="default"),
-        Table(name="dv_and_cm_table", share="share8", schema="default")
+        Table(name="dv_and_cm_table", share="share8", schema="default"),
```
linzhou-db, line 117: extra , at the end?

python/delta_sharing/rest_client.py
```diff
@@ -143,7 +143,7 @@ class DataSharingRestClient:
     USER_AGENT: ClassVar[str] = _client_user_agent()
     DELTA_RESPONSE_FORMAT = "responseformat=delta"
     DELTA_AND_PARQUET_RESPONSE_FORMAT = "responseformat=delta,parquet"
-    DELTA_READER_FEATURES = "readerfeatures=deletionvectors,columnmapping"
+    DELTA_READER_FEATURES = "readerfeatures=deletionvectors,columnmapping,timestampntz"
```
linzhou-db, line 146: does this cover both snapshot and cdf queries? do we need to check whether delta kernel rust support it? or the test indicates that it's supported?

PatrickJin-db: looks like delta kernel rust does not support timestampntz + CDF. is there any way to indicate this limitation in the capabilities header?

python/delta_sharing/tests/test_delta_sharing.py
PatrickJin-db: shouldn't be an issue

python/delta_sharing/rest_client.py
PatrickJin-db: done

python/delta_sharing/rest_client.py
```diff
@@ -23,6 +23,7 @@
 import logging
 import pprint
 
+from numpy import e
```
linzhou-db, line 26: remove this? unused and causing pre-merge test failure.

python/delta_sharing/rest_client.py
```diff
@@ -168,10 +170,16 @@ def __auth_session(self, profile):
...
             self._session.verify = False
 
-    def set_sharing_capabilities_header(self):
+    def __get_reader_features(self, for_cdf):
+        if for_cdf:
+            return DataSharingRestClient.DELTA_CDF_READER_FEATURES
+        else:
+            return DataSharingRestClient.DELTA_SNAPSHOT_READER_FEATURES
+
+    def set_sharing_capabilities_header(self, for_cdf=False):
```
linzhou-db, line 179: for_cdf is not needed for set_sharing_capabilities_header? set_sharing_capabilities_header seems only called in `__get_table_metadata` in delta_sharing.py 
We can just set it to false at L182

python/delta_sharing/rest_client.py
PatrickJin-db: oops, removed

python/delta_sharing/rest_client.py
PatrickJin-db: done. was thinking of adding it in case we need it in the future but we can always make that change later when it's needed.

linzhou-db (approved):

linzhou-db:
@littlegrasscao The pre-merge test is failing because the new parameter `retrySleepInterval` added in https://sourcegraph.com/github.com/delta-io/delta-sharing/-/commit/e1c34ede0c48dbcbaeb342ebcd90d1c4f2c97e2d is not added to this class constructor: https://sourcegraph.com/github.com/delta-io/delta-sharing/-/blob/spark/src/test/scala/io/delta/sharing/spark/TestDeltaSharingClient.scala?L39

PR #656, opened 2025-03-04
Add more logging for client and cache

linzhou-db:
Add more logging for client and cache for better debug ability.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -469,6 +469,7 @@ class DeltaSharingRestClient(
       startingVersion: Long,
       endingVersion: Option[Long]
   ): DeltaTableFiles = {
+    val start = System.currentTimeMillis()
     val encodedShareName = URLEncoder.encode(table.share, "UTF-8")
     val encodedSchemaName = URLEncoder.encode(table.schema, "UTF-8")
     val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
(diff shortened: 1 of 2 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -502,6 +503,8 @@ class DeltaSharingRestClient(
         target = target, data = request, setIncludeEndStreamAction = endStreamActionEnabled
       )
       val (filteredLines, _) = maybeExtractEndStreamAction(response.lines)
+      logInfo(s"Took ${System.currentTimeMillis() - start} ms to query ${filteredLines.size} " +
```
yinouya-db, line 506: If the goal is "better debug ability" rather than normal event logging, should these be on debug level?

yinouya-db (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: @yinouya-db  I'm not sure how to enable debug level logs in DBR, and I wonder we are only able to obtain logs after the enablement so that the logs for the query that's hitting an issue is not available?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -103,7 +103,7 @@ class CachedTableManager(
         cache.remove(tablePath, cachedTable)
       } else if (cachedTable.expiration - System.currentTimeMillis() < refreshThresholdMs) {
         logInfo(s"Updating pre signed urls for $tablePath (expiration time: " +
-          s"${new java.util.Date(cachedTable.expiration)})")
+          s"${new java.util.Date(cachedTable.expiration)}), token:${cachedTable.refreshToken}")
```
littlegrasscao, line 106: Is refreshToken sensitive or it is known to recipient? Since you can print it out here, I guess it already known to recipient.

linzhou-db: no, it's an encrypted token and yes it's known to the recipient

PR #657, opened 2025-03-05
Support delta format for table protocol and metadata

PatrickJin-db:
This PR sets the capabilities header when querying table metadata. The delta protocol and metadata are parsed in a backwards-compatible fashion. Users can also explicitly set `use_delta_format` to False in case there are any old behaviors they need to keep using.


linzhou-db (approved):

PatrickJin-db:
@sameerawi There will be a new release with this fix by early next week.

PR #661, opened 2025-03-09
Update python branch 1.3

PatrickJin-db:
Cherry-picked python changes from main

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Cache Scala, SBT
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: |
             ~/.sbt
@@ -52,7 +52,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Cache Scala, SBT
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: |
             ~/.sbt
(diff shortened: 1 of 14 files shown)


linzhou-db:
@PatrickJin-db  could you also fix the test failure on listAllTables because we added one more table for test, and also patch it together with the fix on the TestDeltaSharingClient class to this branch.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1129,7 +1130,7 @@ class DeltaSharingRestClient(
   ): (Option[Long], Map[String, String], Seq[String]) = {
     // Reset dsQueryId before calling RetryUtils, and before prepareHeaders.
     dsQueryId = Some(UUID.randomUUID().toString().split('-').head)
-    RetryUtils.runWithExponentialBackoff(numRetries, maxRetryDuration) {
+    RetryUtils.runWithExponentialBackoff(numRetries, maxRetryDuration, retrySleepInterval) {
```
moderakh, line 1133: Scala changes are not relevant to the Python 1.3 release. To keep the scope limited, we should cherry-pick only the changes related to Python 1.3 rather than merging the entire branch.

I’m curious—why are we merging the branch instead of cherry-picking the relevant changes?

moderakh (changes requested):
To keep the scope limited, we should cherry-pick only the changes related to Python 1.3 rather than merging the entire branch.

I’m curious—why are we merging the branch instead of cherry-picking the relevant changes?

linzhou-db (approved):

linzhou-db:
wait for @moderakh  to approve as well.

moderakh (approved):

PR #679, opened 2025-04-09
Adding queryParamsHashId into tablePath for Snapshot/CDF/Streaming queries

littlegrasscao:
This change updates the keying strategy of the existing pre-signed URL cache (which stores fileId -> fileUrl mappings) to account for different query shapes on the same table. Previously, the cache key only included the table path, so a new query with different parameters could overwrite existing cache entries. This update ensures correctness by incorporating the full query context into the cache key.

**Query Parameters by Query Type:**

**Snapshot Queries:**
- Filters
- Limit
- Table version

**Change Data Feed (CDF) Queries:**

- A Map[String, String] of CDF options, including:
  - Starting version
  - Ending version

Filters are excluded from the key, as files are fetched for the entire version range and filtered in memory. CDF queries also do not support limits.

**Streaming Queries:**

Streaming parquet files
- Starting version
- Ending version (If applicable)

Streaming CDF files
- Same as CDF queries

Similar to CDF, filters and limits are excluded from the key, since filtering is applied in memory after fetching all relevant files.

This update prevents cache collisions between distinct query shapes on the same table and ensures accurate file-to-URL resolution for all supported query types.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -171,6 +171,10 @@ case class DeltaSharingSource(
     TableRefreshResult(Map.empty[String, String], None, None)
   }
 
+  // Ensure different query shapes against the same table have distinct entries
+  // in the pre-signed URL cache.
+  private var queryParamsHashId: String = _
+
   private lazy val getTableInfoForLogging: String =
     s" for table(id:$tableId, name:${deltaLog.table.toString}, source:$sourceId)"
 
@@ -436,6 +440,7 @@ case class DeltaSharingSource(
         jsonPredicateHints = None,
         refreshToken = None
       )
+      queryParamsHashId = QueryUtils.getQueryParamsHashId(Nil, None, None, fromVersion)
       latestRefreshFunc = _ => {
         val queryTimestamp = System.currentTimeMillis()
         val files = deltaLog.client.getFiles(
@@ -500,6 +505,7 @@ case class DeltaSharingSource(
       val tableFiles = deltaLog.client.getFiles(
         deltaLog.table, fromVersion, Some(endingVersionForQuery)
       )
+      queryParamsHashId = QueryUtils.getQueryParamsHashId(fromVersion, endingVersionForQuery)
       latestRefreshFunc = _ => {
         val queryTimestamp = System.currentTimeMillis()
         val addFiles = deltaLog.client.getFiles(
@@ -570,22 +576,21 @@ case class DeltaSharingSource(
     logInfo(s"Fetching CDF files with fromVersion($fromVersion), fromIndex($fromIndex), " +
       s"endingVersionForQuery($endingVersionForQuery)," + getTableInfoForLogging)
     resetGlobalTimestamp()
+    val cdfOptions = Map(
+      DeltaSharingOptions.CDF_START_VERSION -> fromVersion.toString,
+      DeltaSharingOptions.CDF_END_VERSION -> endingVersionForQuery.toString
+    )
     val tableFiles = deltaLog.client.getCDFFiles(
       deltaLog.table,
-      Map(
-        DeltaSharingOptions.CDF_START_VERSION -> fromVersion.toString,
-        DeltaSharingOptions.CDF_END_VERSION -> endingVersionForQuery.toString
-      ),
+      cdfOptions,
       true
     )
+    queryParamsHashId = QueryUtils.getQueryParamsHashId(cdfOptions)
     latestRefreshFunc = _ => {
       val queryTimestamp = System.currentTimeMillis()
       val d = deltaLog.client.getCDFFiles(
         deltaLog.table,
-        Map(
-          DeltaSharingOptions.CDF_START_VERSION -> fromVersion.toString,
-          DeltaSharingOptions.CDF_END_VERSION -> endingVersionForQuery.toString
-        ),
+        cdfOptions,
         true
       )
 
@@ -764,21 +769,15 @@ case class DeltaSharingSource(
     // version.
     val filteredActions = fileActions.filter{ indexedFile => indexedFile.getFileAction != null }
 
-    // For streaming queries, build the query parameters hash using the start/end version.
-    // All the files between start and end version are cached for a tablePath.
-    val queryParamsHashId = QueryUtils.getQueryParamsHashId(startVersion, endOffset)
-
     if (options.readChangeFeed) {
-      // Streaming CDF
       return createCDFDataFrame(
         filteredActions,
         lastQueryTimestamp,
-        urlExpirationTimestamp,
-        queryParamsHashId
+        urlExpirationTimestamp
       )
     }
 
-    createDataFrame(filteredActions, lastQueryTimestamp, urlExpirationTimestamp, queryParamsHashId)
+    createDataFrame(filteredActions, lastQueryTimestamp, urlExpirationTimestamp)
   }
 
   /**
@@ -789,8 +788,7 @@ case class DeltaSharingSource(
   private def createDataFrame(
       indexedFiles: Seq[IndexedFile],
       lastQueryTimestamp: Long,
-      urlExpirationTimestamp: Option[Long],
-      queryParamsHashId: String): DataFrame = {
+      urlExpirationTimestamp: Option[Long]): DataFrame = {
     val addFilesList = indexedFiles.map { indexedFile =>
       // add won't be null at this step as addFile is the only interested file when
       // options.readChangeFeed is false, which is when this function is called.
@@ -853,8 +851,7 @@ case class DeltaSharingSource(
   private def createCDFDataFrame(
       indexedFiles: Seq[IndexedFile],
       lastQueryTimestamp: Long,
-      urlExpirationTimestamp: Option[Long],
-      queryParamsHashId: String): DataFrame = {
+      urlExpirationTimestamp: Option[Long]): DataFrame = {
     val addFiles = ArrayBuffer[AddFileForCDF]()
     val cdfFiles = ArrayBuffer[AddCDCFile]()
     val removeFiles = ArrayBuffer[RemoveFile]()
--- a/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -770,11 +770,7 @@ case class DeltaSharingSource(
     val filteredActions = fileActions.filter{ indexedFile => indexedFile.getFileAction != null }
 
     if (options.readChangeFeed) {
-      return createCDFDataFrame(
-        filteredActions,
-        lastQueryTimestamp,
-        urlExpirationTimestamp
-      )
+      return createCDFDataFrame(filteredActions, lastQueryTimestamp, urlExpirationTimestamp)
     }
 
     createDataFrame(filteredActions, lastQueryTimestamp, urlExpirationTimestamp)
--- a/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -764,8 +764,12 @@ case class DeltaSharingSource(
     // version.
     val filteredActions = fileActions.filter{ indexedFile => indexedFile.getFileAction != null }
 
+    // For streaming queries, build the query parameters hash using the start/end version.
+    // All the files between start and end version are cached for a tablePath.
     val queryParamsHashId = QueryUtils.getQueryParamsHashId(startVersion, startIndex, endOffset)
+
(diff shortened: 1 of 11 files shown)


client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -163,21 +190,39 @@ private[sharing] case class RemoteDeltaSnapshotFileIndex(
     override val params: RemoteDeltaFileIndexParams,
     limitHint: Option[Long]) extends RemoteDeltaFileIndexBase(params) {
 
+  // Retrieves and caches all files for a version of the table.
   override def inputFiles: Array[String] = {
-    params.snapshotAtAnalysis.filesForScan(Nil, None, None, this)
-      .map(f => toDeltaSharingPath(f).toString)
+    val queryParamsHashId = QueryUtils.getQueryParamsHashId(
```
linzhou-db, line 195: why is it not guarded by flag at this place?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -73,6 +73,11 @@ class CachedTableManager(
     thread
   }
 
+  // Returning how many entries are in the cache.
```
linzhou-db, line 76: is this for testing purpose?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -577,3 +577,10 @@ class DeltaSharingSourceSuite extends QueryTest
     }
   }
 }
+
+class DeltaSharingSourceWithCacheSuite extends DeltaSharingSourceSuite {
```
linzhou-db, line 581: nit: The name of this Suite and also the flag, doesn't seem obvious to me.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
littlegrasscao: The flag check happens in `filesForScan` and `toDeltaSharingPath`, so we don't need to check the flag here.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: yes, this is for testing.

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
littlegrasscao: updated the suite name

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -764,11 +764,21 @@ case class DeltaSharingSource(
     // version.
     val filteredActions = fileActions.filter{ indexedFile => indexedFile.getFileAction != null }
 
+    // For streaming queries, build the query parameters hash using the start/end version.
+    // All the files between start and end version are cached for a tablePath.
+    val queryParamsHashId = QueryUtils.getQueryParamsHashId(startVersion, startIndex, endOffset)
```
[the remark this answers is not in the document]
littlegrasscao: I gave this a second thought. startIndex may not be needed, as the streaming query calls getFiles on the server and only provides the start and end versions for both CDF and Parquet streaming. All file actions are cached between the start and end versions, so the index doesn't really matter.

Since all file actions between the start and end versions are cached, even if startIndex changes, we can still serve the second query as long as the start and end versions remain the same—in cases where a single table version contains a large number of file actions.

Do you agree with this?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
linzhou-db: let's add that in the comment?

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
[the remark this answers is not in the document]
linzhou-db: yeah, I feel so.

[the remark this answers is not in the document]
littlegrasscao: removed startIndex from hash

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: added a comment

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -163,21 +190,39 @@ private[sharing] case class RemoteDeltaSnapshotFileIndex(
...
+      version = params.snapshotAtAnalysis.version
```
charlenelyu-db, line 196: Though the chance might be very low, it's still possible that between this `params.snapshotAtAnalysis.version` (triggers GetTableMetadata) and the `params.snapshotAtAnalysis.filesForScan` below (triggers QueryTable), the latest table version has changed?

We could use the version returned from `QueryTable` call to generate the hash id instead.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -163,21 +190,39 @@ private[sharing] case class RemoteDeltaSnapshotFileIndex(
...
-      convertToJsonPredicate(partitionFilters, dataFilters),
-      this
-    ))
+    val jsonPredicateHints = convertToJsonPredicate(partitionFilters, dataFilters)
+    val queryParamsHashId = QueryUtils.getQueryParamsHashId(
+      partitionFilters.map(_.sql).mkString(";"),
+      dataFilters.map(_.sql).mkString(";"),
+      jsonPredicateHints.getOrElse(""),
+      limitHint.map(_.toString).getOrElse(""),
+      params.snapshotAtAnalysis.version
```
charlenelyu-db, line 213: ditto for possible version discrepancies.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
```diff
@@ -163,21 +190,39 @@ private[sharing] case class RemoteDeltaSnapshotFileIndex(
...
-    makePartitionDirectories(params.snapshotAtAnalysis.filesForScan(
-      partitionFilters ++ dataFilters,
-      limitHint,
-      convertToJsonPredicate(partitionFilters, dataFilters),
-      this
-    ))
+    val jsonPredicateHints = convertToJsonPredicate(partitionFilters, dataFilters)
+    val queryParamsHashId = QueryUtils.getQueryParamsHashId(
+      partitionFilters.map(_.sql).mkString(";"),
+      dataFilters.map(_.sql).mkString(";"),
```
charlenelyu-db, line 210: nit: Similar to above, I think it's better use the real request params to generate the hash id. `partitionFilters` and `dataFilters` might be redundant since `jsonPredicateHints` is included.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -764,11 +764,21 @@ case class DeltaSharingSource(
     // version.
     val filteredActions = fileActions.filter{ indexedFile => indexedFile.getFileAction != null }
 
+    // For streaming queries, build the query parameters hash using the start/end version.
+    // All the files between start and end version are cached for a tablePath.
+    val queryParamsHashId = QueryUtils.getQueryParamsHashId(startVersion, endOffset)
```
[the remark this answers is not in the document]
littlegrasscao: endOffset already includes isStartingVersion, but readChangeFeed is missing. 

This is a good suggestion. I have added a global variable to capture query parameters. getFiles and getCDFfiles requires different inputs, so we should be safe this way.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
littlegrasscao: This is a good point. I changed the logic to build hash id inside filesForScan and return a hash id string to here.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
littlegrasscao: moved the logic into filesForScan

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala
littlegrasscao: I removed partitionFilters and dataFilters, and add a new predicates which seem to also be generated from partitionFilters and dataFilters.

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
```diff
@@ -31,31 +29,17 @@ import org.apache.spark.internal.Logging
...
 import org.apache.spark.sql.catalyst.encoders.ExpressionEncoder
-import org.apache.spark.sql.catalyst.expressions.{
-  And,
-  Attribute,
-  Cast,
-  Expression,
-  Literal,
-  SubqueryExpression
-}
+import org.apache.spark.sql.catalyst.expressions.{And, Attribute, Cast, Expression, Literal, SubqueryExpression}
```
charlenelyu-db, line 32: nit: format

charlenelyu-db (approved):
LGTM!

client/src/main/scala/io/delta/sharing/spark/RemoteDeltaLog.scala
littlegrasscao: fixed format

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -791,11 +796,25 @@ case class DeltaSharingSource(
...
-      spark, initSnapshot, deltaLog.client.getProfileProvider)
+      spark, initSnapshot, deltaLog.client.getProfileProvider, Some(queryParamsHashId))
     val fileIndex = new RemoteDeltaBatchFileIndex(params, addFilesList)
+
+    val tablePathWithParams =
+      if (ConfUtils.sparkParquetIOCacheEnabled(spark.sessionState.conf)) {
+        // Ensure different query shapes against the same table have distinct entries
+        // in the pre-signed URL cache.
+        QueryUtils.getTablePathWithIdSuffix(
+          params.path.toString, params.queryParamsHashId.get
```
linzhou-db, line 810: nit: you can directly use queryParamsHashId, instead of params.queryParamsHashId.get

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -791,11 +796,25 @@ case class DeltaSharingSource(
       add.id -> add.url
     }.toMap
 
+    // For streaming queries, we return a DataFrame.
+    // The FileIndex here is a one-time use object to list files.
     val params = new RemoteDeltaFileIndexParams(
-      spark, initSnapshot, deltaLog.client.getProfileProvider)
+      spark, initSnapshot, deltaLog.client.getProfileProvider, Some(queryParamsHashId))
```
linzhou-db, line 802: This actually have a strong assumption (which should be true), but not enforced in code: `queryParamsHashId` will NOT change after the rpc is made, and before the files are listed.

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -577,3 +577,10 @@ class DeltaSharingSourceSuite extends QueryTest
     }
   }
 }
+
+class DeltaSharingSourceWithParquetIOCacheEnabledSuite extends DeltaSharingSourceSuite {
```
linzhou-db, line 581: Have you run this Suite locally?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSuite.scala
```diff
@@ -566,3 +566,10 @@ class DeltaSharingSuite extends QueryTest with SharedSparkSession with DeltaShar
     }
   }
 }
+
+class DeltaSharingWithParquetIOCacheEnabledSuite extends DeltaSharingSuite {
```
linzhou-db, line 570: And this one?

PR #684, opened 2025-04-16
Fix logging when refreshRes.expirationTimestamp is None

littlegrasscao:
It is possible that refreshRes.expirationTimestamp is None. When it is None, None.get throws an exception and causing register to fail.


linzhou-db (approved):
how did you find this?

littlegrasscao:
> how did you find this?

One of the unit test is not setting this value. It should be always Some() in prod.

PR #688, opened 2025-04-22
Backport to branch 1.2: Fix error message in checkEndStreamAction 

linzhou-db:
Fix error message in checkEndStreamAction

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -1009,13 +1009,13 @@ class DeltaSharingRestClient(
     val includeEndStreamActionHeader = getRespondedIncludeEndStreamActionHeader(capabilitiesMap)
     includeEndStreamActionHeader match {
       case Some(true) =>
-        val lastLineAction = JsonUtils.fromJson[SingleAction](lines.last)
+        val lastLine = lines.last
+        val lastLineAction = JsonUtils.fromJson[SingleAction](lastLine)
         if (lastLineAction.endStreamAction == null) {
-          throw new MissingEndStreamActionException(s"Client sets " +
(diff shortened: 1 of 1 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1009,13 +1009,13 @@ class DeltaSharingRestClient(
     val includeEndStreamActionHeader = getRespondedIncludeEndStreamActionHeader(capabilitiesMap)
     includeEndStreamActionHeader match {
       case Some(true) =>
-        val lastLineAction = JsonUtils.fromJson[SingleAction](lines.last)
+        val lastLine = lines.last
+        val lastLineAction = JsonUtils.fromJson[SingleAction](lastLine)
```
littlegrasscao, line 1013: If lastLine is malformed, can JsonUtils.fromJson go wrong and you won't see the below exception?

littlegrasscao (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: true, I've been thinking about that. 
it will be throw an json parsing exception, which will also be seen but won't have the exception below. 

shall we catch that?

littlegrasscao: It's up to you if we need details below for debugging.

charlenelyu-db (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: let's handle it.

linzhou-db: sent out https://github.com/delta-io/delta-sharing/pull/693

PR #687, opened 2025-04-22
Try to improve error message for server error

linzhou-db:
Try to improve error message for server error, currently it's always the error below, which is not meaningful for debugging: 
```
Request to delta sharing server failed due to org.apache.http.ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected.
```

PR #692, opened 2025-04-22
Backport to branch-1.2: Try to improve error message for server error 

linzhou-db:
The original goal of the PR is to backport https://github.com/delta-io/delta-sharing/pull/687, but it turns out that the cherry-pick cannot auto resolve as there are some previous changes in DeltaSharingClient not backported to branch-1.2. 

I want to include all them as there's no logical change and mostly about logging changes, it's good to keep the code in consistency for logging and debugging purpose.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -19,9 +19,6 @@ package io.delta.sharing.client
 import java.io.{BufferedReader, InputStreamReader}
 import java.net.{URL, URLEncoder}
 import java.nio.charset.StandardCharsets.UTF_8
-import java.sql.Timestamp
-import java.time.LocalDateTime
-import java.time.format.DateTimeFormatter.{ISO_DATE, ISO_DATE_TIME}
 import java.util.UUID
 
 import scala.collection.mutable.{ArrayBuffer, ListBuffer}
@@ -99,6 +96,21 @@ case class ParsedDeltaSharingTablePath(
     share: String,
     schema: String,
     table: String)
+
+/**
+ * A case class containing the parsed response of a delta sharing rpc.
+ *
+ * @param version the table version of the shared table.
+ * @param respondedFormat the sharing format (parquet or delta), used to parse the lines.
+ * @param lines all lines in the response.
+ * @param capabilitiesMap Map parsed from the value of delta-sharing-capabilities in the
+ *                        response header
+ */
+case class ParsedDeltaSharingResponse(
+    version: Long,
+    respondedFormat: String,
+    lines: Seq[String],
+    capabilitiesMap: Map[String, String])
 
 private[sharing] trait PaginationResponse {
   def nextPageToken: Option[String]
@@ -197,8 +209,6 @@ class DeltaSharingRestClient(
 
   // Convert the responseFormat to a Seq to be used later.
   private val responseFormatSet = responseFormat.split(",").toSet
-
-  private var queryId: Option[String] = None
 
   private lazy val client = {
     val clientBuilder: HttpClientBuilder = if (sslTrustAll) {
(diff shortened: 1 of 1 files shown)


linzhou-db:
@charlenelyu-db @littlegrasscao @chakankardb  please help spent 5 minutes to verify that there's no logical change backported. Thanks!

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -625,19 +638,22 @@ class DeltaSharingRestClient(
         expectedRespondedFormat = respondedFormat,
         expectedProtocol = protocol,
         expectedMetadata = metadata,
-        pageNumber = numPages
+        pageNumber = numPages,
+        // Do not set EndStreamAction for async queries yet, and set it for sync queries.
+        setIncludeEndStreamAction = !enableAsyncQuery
```
littlegrasscao, line 643: Should this line be backported?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -751,31 +767,33 @@ class DeltaSharingRestClient(
...
         requestBody = None,
-        expectedVersion = version,
-        expectedRespondedFormat = respondedFormat,
+        expectedVersion = response.version,
+        expectedRespondedFormat = response.respondedFormat,
         expectedProtocol = protocol,
         expectedMetadata = metadata,
-        pageNumber = numPages
+        pageNumber = numPages,
+        setIncludeEndStreamAction = endStreamActionEnabled
```
littlegrasscao, line 775: Should this line be backported?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -788,37 +806,42 @@ class DeltaSharingRestClient(
       expectedRespondedFormat: String,
       expectedProtocol: String,
       expectedMetadata: String,
-      pageNumber: Int): (Seq[String], Option[EndStreamAction]) = {
+      pageNumber: Int,
+      setIncludeEndStreamAction: Boolean): (Seq[String], Option[EndStreamAction]) = {
```
littlegrasscao, line 810: Should setIncludeEndStreamAction be backported?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: checking the logic I think it should be, as the purpose is to not set the header for async queries. 
We could enable it in a follow up launch once we want to. 

Here is the original commit: https://github.com/delta-io/delta-sharing/pull/591/files#diff-365661c3736e3d25a7a5adce691358643b2f6c030990d52875bc314f324d0516
cc @andyl-db  @charlenelyu-db  thought?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: yes this should be needed based on the thread above

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: let's leverage the discussion above.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
charlenelyu-db: Since this PR is for backporting I'm ok with it. But what's the initial reason of not setting it for async queries?

charlenelyu-db (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: I honestly cannot remember exactly. So just guessing to be it not launched and I'm not sure the best behavior for it.

PR #704, opened 2025-04-24
Add DeltaSharingClient Proxy Configuration

zhu-tom:


elghali97:
Nice thanks for the PR @zhu-tom  !

moderakh (changes requested):
@zhu-tom The change looks good to me as a custom modification to support proxy usage when interacting with the server. However, please do not merge the PR.

Thank you

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -227,6 +232,31 @@ class DeltaSharingRestClient(
...
+        val routePlanner = new DefaultRoutePlanner(DefaultSchemePortResolver.INSTANCE) {
+          override def determineRoute(target: HttpHost,
+                                      request: HttpRequest,
+                                      context: HttpContext): HttpRoute = {
+            if (proxyConfig.noProxyHosts.contains(target.getHostName)) {
+              // Direct route (no proxy)
+              new HttpRoute(target)
+            } else {
+              // Route via proxy
+              new HttpRoute(target, proxy)
```
bob-hodgeman-kr, line 253: This constructor always constructs an insecure (http) route through the proxy. This does not work if one is using an **http** proxy to make an **https** call to the server. Instead, consider using an alternate constructor and setting the `secure` flag appropriately based on the target:

```suggestion
              val isSecure = target.getSchemeName == "https"
              new HttpRoute(target, null, proxy, isSecure)
```

bob-hodgeman-kr (changes requested):
Upon testing this PR, I found that I was unable to make an `https` call via an `http` proxy. I've tested the proposed fix and basic testing shows that it is working.

PR #708, opened 2025-04-28
Introducing register/refresh QuerySpecificCachedTable in Pre-signed url cache

littlegrasscao:


client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -174,6 +173,15 @@ trait DeltaSharingProfileProvider {
       refresher: Option[String] => TableRefreshResult): Option[String] => TableRefreshResult = {
     refresher
   }
+
+  // Returns an optional query identifier, which can be used to associate query-specific
+  // refreshers with a table. This is useful for managing refreshers in scenarios where
+  // multiple queries interact with the same table.
+  def getQueryId(): Option[String] = { None }
```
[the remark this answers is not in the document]
linzhou-db: let's rename it to align with other functions. getCustom...

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
```diff
@@ -174,6 +173,15 @@ trait DeltaSharingProfileProvider {
...
+
+  // Returns an optional refresher wrapper containing the server states required to execute
+  // a refresh function.
+  def getRefresherWrapper(): Option[QuerySpecificCachedTable.RefresherWrapper] = { None }
```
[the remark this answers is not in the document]
linzhou-db: let's rename to getCustom...

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -35,32 +37,95 @@ case class TableRefreshResult(
...
-    val refs: Seq[WeakReference[AnyRef]],
     @volatile var lastAccess: Long,
-    val refresher: Option[String] => TableRefreshResult,
-    val refreshToken: Option[String])
+    val refreshToken: Option[String]
+)
+
+/**
+ * Represents a cached table entry with a single refresher function.
+ * All the queries can share the same refresher to update file URLs.
```
linzhou-db, line 58: All the queries with the same XXX can share the same ...

right?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -35,32 +37,95 @@ case class TableRefreshResult(
...
+ * even if two queries are identical in structure, their refreshers can maintain independent states.
+ * The refreshFunction is responsible for refreshing the pre-signed URLs for the table, while the
+ * RefresherWrapper allows additional customization or state management during the refresh process.
+ * This is necessary because the server may require specific states in the refresher function to
+ * refresh the pre-signed URLs. The refresh results can be shared across the same query for the same
+ * table.
+ *
+ * Example:
+ * Query 1: SELECT * FROM table WHERE col1 = 'value1'
+ * Query 2: SELECT * FROM table WHERE col1 = 'value2'
```
linzhou-db, line 92: nit: are you trying to use the same query?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -93,68 +158,152 @@ class CachedTableManager(
...
-            }
-          }
-          val newTable = new CachedTable(
+      } else {
+        cachedTable match {
+          case table: CachedTable =>
+            handleCachedTableRefresh(tablePath, table)
+          case table: QuerySpecificCachedTable =>
+            handleQuerySpecificTableRefresh(tablePath, table)
+          case _ =>
```
linzhou-db, line 186: hmmm, what's our story here? should we throw error?

linzhou-db: let's comment on why logWarning is good enough.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -93,68 +158,152 @@ class CachedTableManager(
...
+            logWarning(s"Unknown table type for $tablePath, type ${cachedTable.getClass}.")
+        }
+      }
+    }
+  }
+
+  private def logDifferencesBetweenUrls(
```
linzhou-db, line 193: no change here right?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -93,68 +158,152 @@ class CachedTableManager(
...
+        logWarning(s"Keys only in refreshRes.idToUrl: ${onlyInRefresh.mkString(", ")}")
+      }
+      if (onlyInCached.nonEmpty) {
+        logWarning(s"Keys only in cachedTable.idToUrl: ${onlyInCached.mkString(", ")}")
+      }
+    }
+  }
+
+  /** Refreshes a `CachedTable` if necessary. */
+  private def handleCachedTableRefresh(tablePath: String, cachedTable: CachedTable): Unit = {
```
linzhou-db, line 215: no change here right?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -93,68 +158,152 @@ class CachedTableManager(
...
+      // If all the references are gone, we will remove the table from the cache.
+      logInfo(s"Removing table $tablePath as no valid query mappings remain.")
+      cache.remove(tablePath, querySpecificCachedTable)
+    } else if (
+      querySpecificCachedTable.expiration - System.currentTimeMillis() < refreshThresholdMs
+    ) {
+      // If the pre-signed URLs are going to expire, we will refresh them.
+      val (_, (_, refresherWrapper)) = validStates.head
+      try {
+        val refreshRes = refresherWrapper(
```
linzhou-db, line 279: will this actually run the refresh function?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -169,13 +318,114 @@ class CachedTableManager(
     }
     cachedTable.lastAccess = System.currentTimeMillis()
     val url = cachedTable.idToUrl.getOrElse(fileId, {
-      logInfo(s"${cachedTable.idToUrl.size} urls in cache " +
```
linzhou-db, line 172: why are we removing this?

linzhou-db (commented):
half way

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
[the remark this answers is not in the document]
littlegrasscao: this is 
UCSManager
      .capture()
      .getHashedStateId

client/src/main/scala/io/delta/sharing/client/DeltaSharingProfileProvider.scala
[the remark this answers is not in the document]
littlegrasscao: updated

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: with this CachedTable, it assumes all queries with the same table share the same refresh. Let me update.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: yes, updated

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: added a comment

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: right, this is only a refactoring.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: right, no change here. Only refactoring

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: yes, this runs refresh function with wrapper applied. added comments

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: this is a rebase mistake. let me add it back.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -35,32 +37,95 @@ case class TableRefreshResult(
...
- *             remove the cached table from our cache.
  * @param lastAccess When the table was accessed last time. We will remove old tables that are not
  *                   accessed after `expireAfterAccessMs` milliseconds.
- * @param refresher the function to generate a new file id to pre sign url map, with the new
- *                  expiration timestamp of the urls and the new refresh token.
  * @param refreshToken the optional refresh token that can be used by the refresher to retrieve
  *                     the same set of files with refreshed urls.
  */
-class CachedTable(
+abstract class BaseCachedTable(
```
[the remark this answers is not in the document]
littlegrasscao: yes that's possible to do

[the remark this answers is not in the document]
littlegrasscao: Do you want me separate this PR into two?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -176,6 +328,109 @@ class CachedTableManager(
...
+      // If the table is not in cache, we will create a new entry
+      val newCachedTable = new QuerySpecificCachedTable(
+        expiration = resolvedExpiration,
+        idToUrl = resolvedIdToUrl,
+        lastAccess = System.currentTimeMillis(),
+        refreshToken = resolvedRefreshToken,
+        refreshFunction = refresher,
+        queryStates = Map(queryId -> (refs, refresherWrapper))
+      )
+      cache.put(tablePath, newCachedTable)
```
charlenelyu-db, line 394: What if other thread insert an entry between `cache.get` above and `cache.put` here?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -176,6 +328,109 @@ class CachedTableManager(
...
+      // be shared across identical queries.
+      val updatedCachedTable = new QuerySpecificCachedTable(
+        expiration = resolvedExpiration,
+        idToUrl = resolvedIdToUrl,
+        lastAccess = System.currentTimeMillis(),
+        refreshToken = resolvedRefreshToken,
+        refreshFunction = refresher,
+        queryStates = newQueryStates
+      )
+      cache.put(tablePath, updatedCachedTable)
```
charlenelyu-db, line 426: ditto. I think this entire function needs to be atomic if you written it this way. Consider using `compute` etc instead, or use a while loop similar to the old `register`.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -205,6 +460,26 @@ class CachedTableManager(
     val customTablePath = profileProvider.getCustomTablePath(tablePath)
     val customRefresher = profileProvider.getCustomRefresher(refresher)
```
charlenelyu-db, line 461: Is it true that both `getCustomTablePath` and `getCustomRefresher` are actually redundant now after we introduced the `queryStates` mapping? These custom methods were introduced to capture the query state to begin with. Shall we dedupe somehow? Like using `customTablePath` and `customRefresher` for the old cache only?

littlegrasscao: It makes sense. I moved getCustomRefresher to only apply for the old logic.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: We need to add a new global ConcurrentHashMap to store `tablePath -> lock`. Both the refresh and register operations will retrieve a lock from this map and synchronize on the section where read and write are combined. 

The refresh logic also needs to delete the lock if the table cache is removed.

This change requires some code refactoring. Given the complexity of this PR, I will add the lock in a follow-up PR.

littlegrasscao: I take that back. I think your suggestion of using compute and computeIfPresent might be better.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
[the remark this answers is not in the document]
linzhou-db: let's prefer small PRs in the future.

linzhou-db (approved):
@charlenelyu-db  please also take a look. 

@littlegrasscao  thanks for this big change! let's test it, especially the execution of the refreshing part. 
also figure out the flag in a separate PR.

PR #709, opened 2025-04-29
[Backport-1.2]: Parse Error Message in Client in EndStreamAction

JialeTomTian:
### Purpose:
Add functionality to parse error endStreamAction within the open source client.

### Tests:
Add unit tests to ensure that error within EndStreamAction is properly parsed.

End to End Testing on 16.3 with branch 1.2
Screenshot 2025-05-06 at 3 32 56 PM

PR #711, opened 2025-05-02
Add support for error message in EndStreamAction

JialeTomTian:
### Purpose:
Add functionality to parse error endStreamAction within the open source client. For the naming, Lin [suggested ](https://github.com/delta-io/delta-sharing/pull/709#discussion_r2070932086) `DeltaSharingRPCExceptionInServer`. I wasn't a fun of the word "in" (lol) and thought `Exception` might be better to put at the end so I went with `DeltaSharingRPCServerException`. Happy to discuss this further to get more opinions.

### Tests:
Add unit tests to ensure that error within EndStreamAction is properly parsed.

End to End Testing: Here are the before and after results within using the original approach compared to the new approach
Before:
Before
After (The error class is a little different, we can discuss on how to name it):
After

### Follow-up:
- Will also backport to 0.7, 1,1 and 1.2. Since the content is very similar, I will cut the backport after this master PR is approved.
- I have added the same field within the Server implementation. However due to the difference in [the way](https://github.com/delta-io/delta-sharing/blob/branch-1.2/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala#L613) the server process streaming requests, we need to figure whether we want to implement this for the open source server and the best way to do this. (Note, this implementation is backwards compatible, meaning we don't have to implement this for the open source server and the behavior with open server would stay the same)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -41,7 +41,7 @@ import org.apache.spark.sql.SparkSession
 import io.delta.sharing.client.auth.{AuthConfig, AuthCredentialProviderFactory}
 import io.delta.sharing.client.model._
 import io.delta.sharing.client.util.{ConfUtils, JsonUtils, RetryUtils, UnexpectedHttpStatus}
-import io.delta.sharing.spark.MissingEndStreamActionException
+import io.delta.sharing.spark.{DeltaSharingRPCServerException, MissingEndStreamActionException}
```
linzhou-db, line 44: What about just `DeltaSharignServerException`?

linzhou-db (approved):
let's finalize on naming,  otherwise looking good.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
JialeTomTian: Updated. If there are no further objections on the name. I will merge this PR and put out all the backport PRs as well. cc @charlenelyu-db @chakankardb @littlegrasscao.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -41,7 +41,7 @@ import org.apache.spark.sql.SparkSession
 import io.delta.sharing.client.auth.{AuthConfig, AuthCredentialProviderFactory}
 import io.delta.sharing.client.model._
 import io.delta.sharing.client.util.{ConfUtils, JsonUtils, RetryUtils, UnexpectedHttpStatus}
-import io.delta.sharing.spark.MissingEndStreamActionException
+import io.delta.sharing.spark.{DeltaSharingServerException, MissingEndStreamActionException}
```
chakankardb, line 44: DeltaSharingServerException sgtm.

server/src/main/scala/io/delta/sharing/server/model.scala
```diff
@@ -157,7 +157,8 @@ case class RemoveFile(
 case class EndStreamAction(
     refreshToken: String,
     nextPageToken: String,
-    minUrlExpirationTimestamp: java.lang.Long
+    minUrlExpirationTimestamp: java.lang.Long,
+    endStreamAction: String = null
```
littlegrasscao, line 161: what is this new endStreamAction used for?

JialeTomTian: Good catch! This should be `errorMessage`. This is for the `EndStreamAction` within the server. I have added this field in the Server model to maintain consistency. This code path is not currently hit and there are no tests for this model change. More context in this comment from my PR here:

> I have added the same field within the Server implementation. However due to the difference in [the way](https://github.com/delta-io/delta-sharing/blob/branch-1.2/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala#L613) the server process streaming requests, we need to figure whether we want to implement this for the open source server and the best way to do this. (Note, this implementation is backwards compatible, meaning we don't have to implement this for the open source server and the behavior with open server would stay the same)

Let me make a fast-follow PR to fix this inconsistency.

PR #712, opened 2025-05-02
Don't use JValue in validateSourceVersion

littlegrasscao:
DBR 17.x uses an incompatible JSON version. This causes failures at execution time.

Solution: Rewrite the code to remove dependency on JValue in the repo.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/DeltaSharingErrors.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/DeltaSharingErrors.scala
@@ -40,8 +40,8 @@ object DeltaSharingErrors {
   }
 
   def unsupportedTableReaderVersion(supportedVersion: Long, tableVersion: Long): Throwable = {
-    new IllegalStateException(s"The table reader version ${tableVersion} is larger than " +
-      s"supported reader version $supportedVersion. Please upgrade to a new release."
+    new IllegalStateException(s"The table reader version ${tableVersion} is not equal to " +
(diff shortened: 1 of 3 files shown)


littlegrasscao:
After this fix, DBR 17.X is still blocked due to Spark version incompatible.

Caused by: NoSuchMethodError: 'void org.apache.spark.sql.Column.<init>(org.apache.spark.sql.catalyst.expressions.Expression)'
[https://github.com/delta-io/delta-sharing/blob/a58a771d7c5e68c3523cfe1425c413401ce[…]rc/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala][(https://github.com/delta-io/delta-sharing/blob/a58a771d7c5e68c3523cfe1425c413401ce9f8d9/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala#L130)](https://github.com/delta-io/delta-sharing/blob/main/client/src/main/scala/io/delta/sharing/spark/RemoteDeltaFileIndex.scala#L130)

linzhou-db:
could you check the premerge test failures?

charlenelyu-db (approved):

littlegrasscao:
> could you check the premerge test failures?

6 test failures caused by org.apache.http.conn.HttpHostConnectException: Connect to localhost:12345 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1] failed: Connection refused (Connection refused)

We might need to look into how to resolve this issue for test stability.

PR #719, opened 2025-05-08
Handle race condition between a refresh thread and register threads

littlegrasscao:
In the previous `CacheTable` implementation, each query was registered as a separate cache entry, which inherently avoided race conditions. However, the new `QuerySpecificCachedTable` groups identical queries under the same cache entry. This change requires us to ensure thread safety during both registration and refresh operations.

Key details:

- A single background refresh thread is responsible for updating the cache.

- There can be multiple concurrent registration threads, each triggered by queries on the same table path.

- To ensure safe concurrent access, we must guard against race conditions between register and refresh logic.

### Thread-Safety Mechanisms
`registerQuerySpecificCachedTable` uses `cache.compute(tablePath, (_, existingTable) -> ...)` to ensure atomic updates when multiple registration threads attempt to initialize or update the cache entry concurrently.

To prevent race conditions between the refresh thread and registration threads, we introduce `cache.computeIfPresent(tablePath, (_, cachedTable) -> ...)` inside `handleQuerySpecificCachedTableRefresh`. This ensures that refresh operations only apply to existing entries in a thread-safe manner.

### DIFF
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -195,7 +195,7 @@ class CachedTableManager(
           case table: CachedTable =>
             handleCachedTableRefresh(tablePath, table)
           case table: QuerySpecificCachedTable =>
-            handleQuerySpecificCachedTableRefresh(tablePath, table)
+            handleQuerySpecificCachedTableRefresh(tablePath)
           case _ =>
             // We should never have a table that is not CachedTable or QuerySpecificCachedTable
             // Also, refresh happens in a background thread, throw an exception doesn't help.
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -325,11 +325,27 @@ class CachedTableManager(
                 logInfo(s"Removing table $tablePath as the pre-signed URL has expired.")
                 null
               } else {
-                querySpecificCachedTable
+                // Update query states
+                new QuerySpecificCachedTable(
+                  querySpecificCachedTable.expiration,
+                  querySpecificCachedTable.idToUrl,
+                  querySpecificCachedTable.lastAccess,
+                  querySpecificCachedTable.refreshToken,
+                  querySpecificCachedTable.refresher,
+                  validStates
+                )
               }
           }
         } else {
-          querySpecificCachedTable
+          // Update query states
+          new QuerySpecificCachedTable(
+            querySpecificCachedTable.expiration,
+            querySpecificCachedTable.idToUrl,
+            querySpecificCachedTable.lastAccess,
+            querySpecificCachedTable.refreshToken,
+            querySpecificCachedTable.refresher,
+            validStates
+          )
         }
 
       case _ =>
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -144,6 +144,19 @@ class CachedTableManager(
     cache.size
   }
 
+  // Returns the number of query states in the cache for a given table path.
+  // This method is mainly for testing purpose.
+  def getQueryStateSize(tablePath: String): Int = {
+    val cachedTable = cache.get(tablePath)
+    if (cachedTable == null) {
+      throw new IllegalStateException(s"table $tablePath was removed")
+    }
+    cachedTable match {
+      case table: QuerySpecificCachedTable => table.queryStates.size
+      case _ => 0
+    }
+  }
+
   def isValidUrlExpirationTime(expiration: Option[Long]): Boolean = {
     // refreshThresholdMs is the buffer time for the refresh RPC.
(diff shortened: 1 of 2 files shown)


client/src/test/scala/org/apache/spark/delta/sharing/CachedTableManagerSuite.scala
```diff
@@ -673,4 +673,194 @@ class CachedTableManagerSuite extends SparkFunSuite with SharedSparkSession{
...
+    assert(manager.size == 1)
+    assert(manager.getQueryStateSize(tablePath) == 10)
+    val (retrievedUrl1, _) = manager.getPreSignedUrl(tablePath, fileId)
+    assert(retrievedUrl1 === initialUrl)
+
+    // We only refresh cache entry when urls are going to expire
+    refs.take(5).foreach(_.clear())
+    manager.refresh()
+    assert(manager.size == 1)
+    assert(manager.getQueryStateSize(tablePath) == 10) // still 10
```
linzhou-db, line 735: why this is still 10 after a refresh()?

littlegrasscao: We only update the entry if it is about to expire:
https://github.com/delta-io/delta-sharing/blob/main/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala#L277

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -261,56 +274,68 @@ class CachedTableManager(
...
-        // Run the refresh function with the wrapper to get the new pre-signed URLs
-        val refreshRes = refresherWrapper(
-          querySpecificCachedTable.refreshToken,
-          querySpecificCachedTable.refresher
-        )
-        logDifferencesBetweenUrls(refreshRes.idToUrl, querySpecificCachedTable.idToUrl)
+  private def handleQuerySpecificCachedTableRefresh(tablePath: String): Unit = {
+    // Use `computeIfPresent` to ensure atomicity of the operation and prevent interference
+    // with concurrent register threads.
+    cache.computeIfPresent(tablePath, (_, cachedTable) => cachedTable match {
```
linzhou-db, line 280: this seems the only/main change in the PR, all others are just indent changes?

linzhou-db (commented):
could you also fix the lint errors in pre-merge tests?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: Yes, others are indent and return type changes

littlegrasscao:
> could you also fix the lint errors in pre-merge tests?

I fixed the format issue in the config change PR:
https://github.com/delta-io/delta-sharing/pull/720

linzhou-db:
> > could you also fix the lint errors in pre-merge tests?
> 
> I fixed the format issue in the config change PR: #720

@littlegrasscao why not in the same PR?

linzhou-db (approved):

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -261,56 +274,68 @@ class CachedTableManager(
...
+              queryStates = validStates
+            )
+          } catch {
+            case NonFatal(e) =>
+              logError(s"Failed to refresh pre-signed URLs for table $tablePath", e)
+              if (querySpecificCachedTable.expiration < System.currentTimeMillis()) {
+                logInfo(s"Removing table $tablePath as the pre-signed URL has expired.")
+                null
+              } else {
+                querySpecificCachedTable
```
charlenelyu-db, line 328: We can still update the cache with the new `validStates` here?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -261,56 +274,68 @@ class CachedTableManager(
...
+              }
           }
-      }
-    }
+        } else {
+          querySpecificCachedTable
```
charlenelyu-db, line 332: ditto.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -261,56 +274,68 @@ class CachedTableManager(
...
+        }
+
+      case _ =>
+        logWarning(s"Unexpected table type for $tablePath. Expected QuerySpecificCachedTable.")
```
charlenelyu-db, line 336: Shall we fail with illegal state here?

charlenelyu-db (approved):

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: Sure, we can copy the object. let me update there.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: This is a background thread. throw an exception won't do anything. 

There shouldn't be a type mismatch as caller should already check the type.

PR #720, opened 2025-05-09
Switch SPARK_PARQUET_IO_CACHE_CONF to use driver level conf

littlegrasscao:
To prevent query registration failures when toggling the flag in DBR, we need to use a driver-level configuration instead of a session-level one. 

This is because, at the time the flag is flipped, the pre-signed URL cache is global, and there may still be existing CacheTable entries for the same table. Attempting to register a new QuerySpecificCachedTable in this state can fail due to class type mismatches. 

To ensure consistency, the flag should only take effect upon Spark cluster restart.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
+++ b/client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
@@ -803,7 +803,7 @@ case class DeltaSharingSource(
     val fileIndex = new RemoteDeltaBatchFileIndex(params, addFilesList)
 
     val tablePathWithParams =
-      if (ConfUtils.sparkParquetIOCacheEnabled(spark.sessionState.conf)) {
+      if (ConfUtils.sparkParquetIOCacheEnabled(spark.sparkContext.getConf)) {
         // Ensure different query shapes against the same table have distinct entries
         // in the pre-signed URL cache.
         QueryUtils.getTablePathWithIdSuffix(
--- a/spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
+++ b/spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
@@ -21,7 +21,7 @@ import java.nio.file.Files
 
 import org.apache.commons.io.FileUtils
 import org.apache.hadoop.fs.Path
-import org.apache.spark.{SparkException, SparkFunSuite}
+import org.apache.spark.{SparkConf, SparkException, SparkFunSuite}
 import org.apache.spark.delta.sharing.CachedTableManager
 import org.apache.spark.sql.SparkSession
 import org.apache.spark.sql.catalyst.expressions.{AttributeReference => SqlAttributeReference, EqualTo => SqlEqualTo, GreaterThan => SqlGreaterThan, Literal => SqlLiteral}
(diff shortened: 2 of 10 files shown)


spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
```diff
@@ -102,428 +98,162 @@ class RemoteDeltaLogSuite extends SparkFunSuite with SharedSparkSession {
     assert(TestDeltaSharingClient.jsonPredicateHints.size === 0)
   }
 
-  test("read file urls from pre-signed url cache with queryParamsHashId") {
+  test("jsonPredicateV2Hints test") {
```
littlegrasscao, line 101: Nothing changed in tests. I just moved config related tests to a new suite.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -803,7 +803,7 @@ case class DeltaSharingSource(
     val fileIndex = new RemoteDeltaBatchFileIndex(params, addFilesList)
 
     val tablePathWithParams =
-      if (ConfUtils.sparkParquetIOCacheEnabled(spark.sessionState.conf)) {
+      if (ConfUtils.sparkParquetIOCacheEnabled(spark.sparkContext.getConf)) {
```
charlenelyu-db, line 806: `SparkEnv.get.conf` might be slightly preferred than `sparkContext.getConf` because the latter one will make a copy of all configs each time called. 
image

charlenelyu-db (approved):
> Attempting to register a new QuerySpecificCachedTable in this state can fail due to class type mismatches.

Rethinking this - technically above should not happen because the cache key will be different for two types right? 

There are pros and cons to change it to a cluster scope conf. The main pro is there won't be mixed types in the cache. The main con is it'll be harder to dynamically ramp up the flag. You'll need some additional change to integrate with SAFE.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
littlegrasscao: I think safer only stores in sparkContext not sparkEnv:
image
image

littlegrasscao:
> > Attempting to register a new QuerySpecificCachedTable in this state can fail due to class type mismatches.
> 
> Rethinking this - technically above should not happen because the cache key will be different for two types right?
> 
> There are pros and cons to change it to a cluster scope conf. The main pro is there won't be mixed types in the cache. The main con is it'll be harder to dynamically ramp up the flag. You'll need some additional change to integrate with SAFE.

You're right that the keys are different. However, the config check happens in multiple places—one before cache registration, and another before returning the tablePath + fileId. Although it's very unlikely, there's a chance the flag could flip in the middle.

littlegrasscao:
> > > Attempting to register a new QuerySpecificCachedTable in this state can fail due to class type mismatches.
> > 
> > 
> > Rethinking this - technically above should not happen because the cache key will be different for two types right?
> > There are pros and cons to change it to a cluster scope conf. The main pro is there won't be mixed types in the cache. The main con is it'll be harder to dynamically ramp up the flag. You'll need some additional change to integrate with SAFE.
> 
> You're right that the keys are different. However, the config check happens in multiple places—one before cache registration, and another before returning the tablePath + fileId. Although it's very unlikely, there's a chance the flag could flip in the middle.

I take this back. SQLConf only gets applied when a new session starts. So it won't affect a running query. It is safe to use session level conf.

PR #733, opened 2025-05-22
OAuth Private Key support

paalvibe:
Private Key is a commonly used pattern in OAuth, and is used by national identity infrastructure in Norway (Skyporten).

### DIFF
--- a/python/delta_sharing/_internal_auth.py
+++ b/python/delta_sharing/_internal_auth.py
@@ -15,15 +15,17 @@
 #
 
 from abc import ABC, abstractmethod
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Optional
 import requests
 import base64
 import json
+from jwcrypto import jwk, jwt
 import threading
 import requests.sessions
 import time
 from typing import Dict
+import uuid
 
 from delta_sharing.protocol import (
     DeltaSharingProfile,
@@ -113,28 +115,10 @@ class OAuthClientCredentials:
         self.creation_timestamp = creation_timestamp
 
 
-class OAuthClient:
-    def __init__(
-        self, token_endpoint: str, client_id: str, client_secret: str, scope: Optional[str] = None
-    ):
-        self.token_endpoint = token_endpoint
-        self.client_id = client_id
-        self.client_secret = client_secret
-        self.scope = scope
-
+class OAuthClient(ABC):
+    @abstractmethod
     def client_credentials(self) -> OAuthClientCredentials:
-        credentials = base64.b64encode(
-            f"{self.client_id}:{self.client_secret}".encode("utf-8")
-        ).decode("utf-8")
-        headers = {
-            "accept": "application/json",
-            "authorization": f"Basic {credentials}",
-            "content-type": "application/x-www-form-urlencoded",
-        }
-        body = f"grant_type=client_credentials{f'&scope={self.scope}' if self.scope else ''}"
-        response = requests.post(self.token_endpoint, headers=headers, data=body)
-        response.raise_for_status()
-        return self.parse_oauth_token_response(response.text)
+        pass
 
     def parse_oauth_token_response(self, response: str) -> OAuthClientCredentials:
         if not response:
@@ -142,7 +126,9 @@ class OAuthClient:
         # Parsing the response per oauth spec
         # https://datatracker.ietf.org/doc/html/rfc6749#section-5.1
         json_node = json.loads(response)
-        if "access_token" not in json_node or not isinstance(json_node["access_token"], str):
+        if "access_token" not in json_node or not isinstance(
+            json_node["access_token"], str
+        ):
             raise RuntimeError("Missing 'access_token' field in OAuth token response")
         if "expires_in" not in json_node:
             raise RuntimeError("Missing 'expires_in' field in OAuth token response")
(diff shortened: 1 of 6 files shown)


python/delta_sharing/_internal_auth.py
```diff
@@ -169,6 +153,94 @@ def parse_oauth_token_response(self, response: str) -> OAuthClientCredentials:
...
+        if algorithm is None:
+            algorithm = "RS256"
+        self.algorithm = algorithm
+
+    def client_credentials(self) -> OAuthClientCredentials:
+        timestamp = int(datetime.now(timezone.utc).timestamp())
+        jwt_header = {"alg": self.algorithm, "kid": self.key_id}
+        jwt_claims = {
+            "aud": self.issuer,
+            "iss": self.client_id,
```
moderakh, line 212: @paalvibe how do you intend to use this client? Specifically, do you plan to use this change with [Delta Sharing over OIDC M2M](https://docs.databricks.com/aws/en/delta-sharing/sharing-over-oidc-m2m)?

From the PR, it sounds like you are sending a self-signed token to your configured token endpoint, which then returns a JWT. 

I am curious what the resultant token from your token endpoint looks like?

Could you share a sample of the JWT that your token endpoint returns (and that the client ultimately sends to the Delta Sharing server), along with the corresponding OIDC federation policy configuration the server uses to authenticate the request?

python/delta_sharing/_internal_auth.py
```diff
@@ -169,6 +153,94 @@ def parse_oauth_token_response(self, response: str) -> OAuthClientCredentials:
...
+            "content-type": "application/x-www-form-urlencoded",
+        }
+        response = requests.post(self.token_endpoint, headers=headers, data=body)
+        response.raise_for_status()
+        return self.parse_oauth_token_response(response.text)
+
+    def _signed_jwt(self, jwt_header, jwt_claims):
+        """Generate a signed JWT token using the private key"""
+        jwt_token = jwt.JWT(header=jwt_header, claims=jwt_claims)
+        with open(self.private_key, "rb") as key_file:
```
moderakh, line 237: This assumes the key_file is available on disk and that the client can read it directly. How would this work in a Spark cluster environment, where the key isn’t necessarily stored on local disk? Where is the key expected to be stored in that case?

moderakh (commented):
@paalvibe
Thank you for filing this feature request and raising the PR, it’s a great effort to extend the Delta Sharing client.

Apologies for the delay in reviewing. I do have a few questions, adding them on the PR.

python/delta_sharing/_internal_auth.py
paalvibe: Hi, Moe,

Yes, we are planning to use Delta Sharing over OIDC M2M.

Here is an example of the decoded JWT returned by the token endpoint:

```
{
  "aud": "https://acme.org",
  "sub": "0192: 315300649;acme:customerdata.gold",
  "scope": "acme:customerdata.gold",
  "iss": "https://sky.maskinporten.no",
  "client_amr": "private_key_jwt",
  "token_type": "Bearer",
  "exp": 1694222211,
  "iat": 1694333311,
  "client_id": "abcd1234-1234-abcd-abcd-12341234abcd",
  "jti": "lwlwlwlw4lwlwlwlwl4lwlw4-lw-lwl4lwl4lwl4lwl4",
  "consumer": {
    "authority": "iso6523-actorid-upis",
    "ID": "0192:315300649"
  }
}
```

Here are the policy details needed in Databricks OIDC server configuration:

```
Issuer URL: https://sky.maskinporten.no/
Subject claim: sub
Subject: 0192:315300649;acme:customerdata.gold
Audiences: https://acme.org/
```

`sky.maskinporten.no` is the national company identity provider in Norway: https://docs-digdir-no.translate.goog/docs/Maskinporten/maskinporten_skyporten.html?_x_tr_sl=no&_x_tr_tl=en&_x_tr_hl=en&_x_tr_pto=wapp

It is a machine to machine identity service, where access is granted in a maskinporten web interface or API according to organization number e.g. "0192:315300649" and scope e.g. "acme:customerdata.gold".

We have already implemented this for sharing cloud resources on AWS, Azure and GCP, as explained in the documentation above.

We have already tested it with our branch and it works. It would enable any data consumer identified with skyporten to read a Delta Share wherever they are, inside or outside Databricks, e.g. in a local notebook.

More info if of intersest:

At [Samferdselsdata.no](http://samferdselsdata.no/) (Public Transport Sector Data Sharing initiative) we have working with [Norwegian Digitalisation Agency (Digdir)](https://www.linkedin.com/company/digitaliseringsdirektoratet/) to implement Delta Shares directly with the Maskinporten National Orgnumber OAuth2 service for authentication, like we have achieved for IAM based cloud access with Skyporten (https://docs.digdir.no/docs/Maskinporten/maskinporten_skyporten). This would mean that one could simply declare which org number should have access, and avoid any credentials exchange at all. Hopefully, a similar pattern will also be possible to use across Europe soon. In practice this enables country code+orgnumber based delta sharing instead of Entra or email-based access.

python/delta_sharing/_internal_auth.py
paalvibe: We have not tested it inside a spark cluster environment, since we envisioned that data scientists would want to read the data in their development context. However, we are happy to get feedback from you on how to get it. In spark, one could get the key from a secrets service, and write the contents to a local path which is passed to DeltaSharingProfile(). However, an alternative or secondary option could be to pass the secret value into DeltaSharingProfile() as parameter.

python/delta_sharing/protocol.py
```diff
@@ -78,7 +83,24 @@ def from_json(json) -> "DeltaSharingProfile":
             )
         elif share_credentials_version == 2:
             type = json["type"]
-            if type == "oauth_client_credentials":
+            if type == "oauth_client_private_key":
```
moderakh: Can we update the share profile format for the new auth type to use the following structure?
```json
{
  "shareCredentialsVersion": 2,
  "endpoint": "https://sharing.delta.io/delta-sharing",
  "type": "oauth_jwt_bearer_private_key_jwt",
  "auth": {
    "tokenEndpoint": "https://test.sky.maskinporten.no/token",
    "clientId": "6ff6dc4a-6924-42a7-a692-309b2323bb5c",
    "issuer": "https://test.sky.maskinporten.no/",
    "audience": "https://sky.kartverket.no/",
    "scope": "entur:skyporten.demo",
    "privateKey" : {
      "keyId": "b7eb622a-58b5-4e3c-9358-45301c4ea584",
      "privateKeyFile": "/home/secrets/delta_share.pem",
      "algorithm": "RS256"
    }
  }
}
```

We should keep the existing supported format unchanged for backward compatibility.
This new structure would only apply to the new authentication method.

paalvibe: Yes. We will make the change.

Issue #735, opened 2025-05-24
Delta format query with version specified is broken in reference server


PatrickJin-db:
Details TBD.

PR #736, opened 2025-05-24
Improve python connector test coverage

PatrickJin-db:
Adds integration tests for adding columns + updating rows, with and without partitions, for both snapshot and CDF queries. These tests also caught the following bugs:
* CDF queries do not preserve column order (fixed in this PR too)
* #735 
* Reference server does not return writerFeatures in delta format CDF query response. I'm unsure if this is a server bug or a client bug, but it may cause CDF queries with kernel to fail, since kernel enforces that writerFeatures must be included if minWriterVersion >= 7.


python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -1687,3 +1690,420 @@ def test_load_table_changes_as_spark(
...
+@pytest.mark.parametrize(
+    "fragments,version,expected",
+    [
+        pytest.param(
+            "share8.default.add_columns_non_partitioned_cdf",
+            3,
+            pd.DataFrame(
+                {
+                    "c1": pd.Series([1, 2, 3], dtype="int32"),
+                    "c2": pd.Series([None, None, 4.0], dtype="object"),
```
PatrickJin-db, line 1705: I have noticed that when an integer column has null values, the type it gets converted to is inconsistent: sometimes the column gets cast to float, sometimes to object. Is this a bug too?

PatrickJin-db: I think it is a bug in our code that stems from pandas type conversion rules. A column that only contains `None` will have type `object`, and if you concatenate with a column of an integer type, the column type remains `object`. However, if you create a column with both integers and None, the column will instead have a float type. Which of these two cases gets hit would thus depend on if a column was missing or only contained null values in one of the source parquet files.

One question is how strict we should be about our mappings from delta types to pandas types. I'm afraid strictly enforcing type conversions will increase memory usage due to triggering pandas consolidation. Maybe we can introduce a flag for how to handle nullable int columns.

linzhou-db: - in L1705, 4.0 is a float but not an integer? 
- but yeah, introducing a flag/parameter sg.

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -1687,3 +1689,420 @@ def test_load_table_changes_as_spark(
             match="Unable to import pyspark. `load_table_changes_as_spark` requires" + " PySpark.",
         ):
             load_table_changes_as_spark("not-used")
+
```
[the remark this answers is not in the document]
PatrickJin-db: done

python/delta_sharing/tests/test_delta_sharing.py
PatrickJin-db: Good point, no need to use 4.0 in that test case since it's object type. I changed it to 4. As for the flag for strict type conversion, can we create an issue and fix it later? It's not urgent and I don't want to block the next release on the type conversion issue.

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -1687,3 +1689,436 @@ def test_load_table_changes_as_spark(
...
+
+@pytest.mark.skipif(not ENABLE_INTEGRATION, reason=SKIP_MESSAGE)
+@pytest.mark.parametrize(
+    "fragments,version,expected",
+    [
+        pytest.param(
+            "share8.default.add_columns_non_partitioned_cdf",
+            3,
+            # table initially contains c1 = [1, 2]
```
linzhou-db: version x contains...
version y added int column...
version z insert a row...

python/delta_sharing/reader.py
```diff
@@ -382,17 +382,24 @@ def table_changes_to_pandas(self, cdfOptions: CdfOptions) -> pd.DataFrame:
...
             pdfs.append(pdf)
 
-        return pd.concat(pdfs, axis=0, ignore_index=True, copy=False)
+        merged = pd.concat(pdfs, axis=0, ignore_index=True, copy=False)
+
+        col_map = {}
+        for col in merged.columns:
+            col_map[col.lower()] = col
+
+        return merged[[col_map[field["name"].lower()] for field in schema_with_cdf["fields"]]]
```
linzhou-db, line 419: this is to preserve the column order?

linzhou-db: and why does the column order matter?

linzhou-db (dismissed):

python/delta_sharing/tests/test_delta_sharing.py
PatrickJin-db: added

python/delta_sharing/reader.py
PatrickJin-db: for non-cdf query we preserve column order, so it only makes sense to do the same for cdf queries too. i'm also pretty sure kernel preserves column order.

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -1795,3 +1797,516 @@ def test_load_table_changes_as_pandas_delta_batch_convert(
...
+            # table initially contains c1 = [1, 2] at version 1
+            # then we add int column c2 at version 2
+            # then we insert a row with c1 = 3, c2 = 4
+            pd.DataFrame({"c1": pd.Series([1, 2, 3], dtype="int32"), "c2": [None, None, 4]}),
+            id="version 3",
+        ),
+        pytest.param(
+            "share8.default.add_columns_non_partitioned_cdf",
+            4,
+            # v4: change c1 to 5 where c1 = 1
```
chakankardb: Nit: c2 to 5 ?

PatrickJin-db: fixed

PR #743, opened 2025-06-03
Build both scala 2.12 and scala 2.13

littlegrasscao:
To align with the Delta OSS master branch, we now support building with both Scala 2.12 and Scala 2.13 in the master branch.

- Scala 2.12: Depends on Spark 3.5.3. A new scala-2.12 directory has been added to host version-specific code.
- Scala 2.13: Depends on Spark 4.0.0. A new scala-2.13 directory has been added to host version-specific code.

Additionally, a new CI pipeline has been added to run Scala 2.12 tests: `build-and-test-client-scala-2_12`.
The `push`-triggered CI pipeline has been removed.


.github/workflows/build-and-test.yml
```diff
@@ -1,5 +1,5 @@
 name: Build and Test
-on: [push, pull_request, workflow_dispatch]
+on: [pull_request, workflow_dispatch]
```
linzhou-db, line 2: what's the consequence of removing `push`?

.github/workflows/build-and-test.yml
```diff
@@ -55,6 +55,33 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Cache Scala, SBT
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.sbt
+            ~/.ivy2
+            ~/.cache/coursier
+          key: build-and-test-client
```
linzhou-db: this key is the same as L50, does it matter?

.github/workflows/build-and-test.yml
```diff
@@ -55,6 +55,33 @@ jobs:
           java-version: '17'
       - run: ./build/sbt client/test spark/test
 
+  build-and-test-client-scala-2_12:
```
linzhou-db, line 58: Do we want to change add -2_13 to L31?

linzhou-db:
@littlegrasscao  is this PR still needed?

.github/workflows/build-and-test.yml
littlegrasscao: CI pipeline would't be triggered if the branch doesn't create a PR.

Once a PR is created, every push still triggers a CI run. As of now, we always trigger 2 runs when pushing a commit to a PR. With this change, we will only see 1 trigger.

.github/workflows/build-and-test.yml
littlegrasscao: updated to add -scala-2_13

.github/workflows/build-and-test.yml
littlegrasscao: added suffix to the key

littlegrasscao:
> anything worth attention in the scala files? or is it just copy-paste?

Scala files under 2.13 directories are copies.
Scala files under 2.12 directories are copies with reverting spark upgrade fixes.

build.sbt
```diff
@@ -66,20 +74,24 @@ lazy val root = (project in file(".")).aggregate(client, spark, server)
...
-    "org.apache.httpcomponents" % "httpclient" % "4.5.14",
-    "org.apache.spark" %% "spark-sql" % sparkVersion % "provided",
-    "org.apache.spark" %% "spark-catalyst" % sparkVersion % "test" classifier "tests",
-    "org.apache.spark" %% "spark-core" % sparkVersion % "test" classifier "tests",
-    "org.apache.spark" %% "spark-sql" % sparkVersion % "test" classifier "tests",
-    "org.scalatest" %% "scalatest" % "3.2.3" % "test",
-    "org.scalatestplus" %% "mockito-4-11" % "3.2.18.0" % "test"
-  ),
+  libraryDependencies ++= {
+    val sv = scalaVersion.value
```
linzhou-db, line 83: question: scalaVersion is defind in L76, does L83 always result in `2.13.13`, or does it consider `crossScalaVersions`?

linzhou-db:
and curious how is the change in yml and sbt file tested?

build.sbt
littlegrasscao: scalaVersion at L76 is the default build scala version.

scalaVersion.value here is dynamic based on the scala version is currently compiling.

PR #749, opened 2025-06-16
Support HTTPS downgrade when not using the proxy. Add support for wildcard DNS support for no proxy hosts.

### LINKED ISSUE #747: spark.delta.sharing.network.never.use.https does not work if proxy is not used
There's a lot of internal inner workings of the apache library, but request.getRequestLine.getUri is actually empty for non-proxy calls. I will wrap an HttpClient that creates a proxy and a non-proxy client (if applicable) and then switch between the 2 depending on the URI so that we don't have to interfere with the inner workings of HttpClient.

wchau:
The previous code touched too much Apache HttpClient internals resulting in certain edge cases not being considered when HTTPS downgrade and proxy settings are being used. This creates a delegating HttpClient that wraps a proxy and a non-proxy HttpClient so that we can minimize overriding internal methods of how HttpClient works. Adding anything more within these internals to fix this issue would make it more difficult to maintain and may introduce other bugs.

This also adds wildcard DNS support for no proxy hosts.

Fixes #747

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
@@ -1,0 +1,81 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.sharing.client
+
+import java.net.URI
+
+import org.apache.http.{HttpHost, HttpRequest}
+import org.apache.http.client.methods.{CloseableHttpResponse, HttpRequestWrapper}
+import org.apache.http.client.utils.URIBuilder
+import org.apache.http.conn.ClientConnectionManager
+import org.apache.http.impl.client.CloseableHttpClient
+import org.apache.http.params.HttpParams
+import org.apache.http.protocol.HttpContext
+
+private[sharing] case class DeltaSharingFileSystemHttpClient(
+  noProxyHttpClient: CloseableHttpClient,
+  proxyHttpClient: CloseableHttpClient,
+  useProxy: Boolean,
+  noProxyHosts: Seq[String],
+  disableHttps: Boolean) extends CloseableHttpClient {
+
+  private def hasNoProxyHostsMatch(host: String): Boolean = {
+    noProxyHosts.exists(record => {
+      // Wildcard DNS records support
+      if (record.startsWith("*.")) {
+        host.endsWith(record.drop(2))
+      } else {
+        host == record
+      }
+    })
+  }
+
+  override protected def doExecute(
+    target: HttpHost,
+    request: HttpRequest,
+    context: HttpContext
+  ): CloseableHttpResponse = {
+    val (updatedTarget, updatedRequest) = if (disableHttps && target.getSchemeName == "https") {
+      val modifiedUri: URI = new URIBuilder(request.getRequestLine.getUri).setScheme("http").build()
+      val wrappedRequest = HttpRequestWrapper.wrap(request)
+      wrappedRequest.setURI(modifiedUri)
+      (new HttpHost(target.getHostName, target.getPort, "http"), wrappedRequest)
+    } else {
+      (target, request)
+    }
+    if (useProxy && !hasNoProxyHostsMatch(target.getHostName)) {
+      proxyHttpClient.execute(updatedTarget, updatedRequest, context)
+    } else {
+      noProxyHttpClient.execute(updatedTarget, updatedRequest, context)
+    }
+  }
+
+  override def close(): Unit = {
+    noProxyHttpClient.close()
+    proxyHttpClient.close()
+  }
+
+  // This is deprecated since 4.3, so overriding with a dummy implementation
+  override def getParams(): HttpParams = {
+    throw new UnsupportedOperationException("getParams")
+  }
+
+  // This is deprecated since 4.3, so overriding with a dummy implementation
+  override def getConnectionManager(): ClientConnectionManager = {
+    throw new UnsupportedOperationException("getConnectionManager")
+  }
+}
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
@@ -33,11 +33,11 @@ private[sharing] case class DeltaSharingFileSystemHttpClient(
   noProxyHosts: Seq[String],
   disableHttps: Boolean) extends CloseableHttpClient {
 
-  private def hasNoProxyHostsMatch(host: String): Boolean = {
+  private[sharing] def hasNoProxyHostsMatch(host: String): Boolean = {
     noProxyHosts.exists(record => {
       // Wildcard DNS records support
       if (record.startsWith("*.")) {
-        host.endsWith(record.drop(2))
+        host.endsWith(record.drop(1))
       } else {
         host == record
       }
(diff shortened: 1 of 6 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClientBuilder.scala
```diff
@@ -0,0 +1,85 @@
...
+
+package io.delta.sharing.client
+
+import org.apache.http.HttpHost
+import org.apache.http.client.config.RequestConfig
+import org.apache.http.impl.client.HttpClientBuilder
+
+private[sharing] class DeltaSharingFileSystemHttpClientBuilder {
+  private var noProxyHttpClientClientBuilder: HttpClientBuilder = HttpClientBuilder.create()
+  private var proxyHttpClientClientBuilder: HttpClientBuilder = HttpClientBuilder.create()
```
littlegrasscao, line 25: Does it mean we keep two HTTP client Builders for every request? Are these cheap to maintain?
Since useHTTPS is determined before entering here, the request would only need one of them?

linzhou-db: good point.

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
```diff
@@ -0,0 +1,81 @@
...
+  noProxyHttpClient: CloseableHttpClient,
+  proxyHttpClient: CloseableHttpClient,
+  useProxy: Boolean,
+  noProxyHosts: Seq[String],
+  disableHttps: Boolean) extends CloseableHttpClient {
+
+  private[sharing] def hasNoProxyHostsMatch(host: String): Boolean = {
+    noProxyHosts.exists(record => {
+      // Wildcard DNS records support
+      if (record.startsWith("*.")) {
```
linzhou-db, line 39: Is there a test case for this?

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileSystemSuite.scala
```diff
@@ -235,7 +235,59 @@ class DeltaSharingFileSystemSuite extends SparkFunSuite {
...
+  test("https traffic is downgraded to http") {
+    val server = new Server(0)
+    val handler = new ServletHandler()
+    server.setHandler(handler)
+    handler.addServletWithMapping(new ServletHolder(new HttpServlet {
+      override def doGet(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+        resp.setContentType("text/plain")
+        resp.setStatus(HttpServletResponse.SC_OK)
+
+        // scalastyle:off println
```
linzhou-db, line 247: do we still need this in the test?

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
```diff
@@ -0,0 +1,81 @@
...
+  ): CloseableHttpResponse = {
+    val (updatedTarget, updatedRequest) = if (disableHttps && target.getSchemeName == "https") {
+      val modifiedUri: URI = new URIBuilder(request.getRequestLine.getUri).setScheme("http").build()
+      val wrappedRequest = HttpRequestWrapper.wrap(request)
+      wrappedRequest.setURI(modifiedUri)
+      (new HttpHost(target.getHostName, target.getPort, "http"), wrappedRequest)
+    } else {
+      (target, request)
+    }
+    if (useProxy && !hasNoProxyHostsMatch(target.getHostName)) {
```
linzhou-db, line 60: according to this, the client to be used is actually decided during execution time?


cc @littlegrasscao

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
wchau: Whoops. Missed the push for this. Added tests for the function. Can't test in integration since we don't have any special domains for local server.

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileSystemSuite.scala
wchau: I am following what the other tests are doing. We still need the scalastyle override to make the scalastyle test pass.

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClient.scala
wchau: Yes, there's logic that needs to be considered during runtime, so it's not static.

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClientBuilder.scala
wchau: There's logic that needs to be considered during runtime, so it's not static, but we are not constructing a new http client for every request.

These should be cheap to maintain after initialization. I want to minimize mistakes from accidentally not constructing the client when we actually need it, since that logic should be abstracted outside of this client builder preferably.

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileSystemSuite.scala
linzhou-db: no the scalastyle, but the "hello world" line.

linzhou-db:
@littlegrasscao  could you take a look at the test failure related to "spark read limit"?

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClientBuilder.scala
moderakh: @wchau I think we should be cautious about instantiating this per request. If we repeatedly open and close a connection pool, we could easily exhaust the available ephemeral ports. Even after a connection is closed, the OS typically keeps the socket in a CLOSE_WAIT or TIME_WAIT state for up to 90 seconds, which can lead to port exhaustion under high load.

wchau: The difference between before and after is that we now instantiate 2 http clients instead one. We are not creating a http client for every URL being accessed.

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileSystemSuite.scala
wchau: Ah. Do you want me to remove this from all the tests?

linzhou-db: yeah, if it's not used in any test, then we can remove?

linzhou-db (commented):
question: what's the risk of this change affecting existing queries?

wchau:
> question: what's the risk of this change affecting existing queries?

@linzhou-db This affects access of delta sharing storage URLs, but the testing we have is pretty comprehensive since it brings up a local proxy and storage server.

There are no customers using both the never.use.https and noProxyHosts today, so the new parts where both http clients will be used will only with the upcoming DB URLs testing.

I am trying to also build a custom image and run it in serverless to make sure it all works, but there may be other issues (unrelated to this).

linzhou-db:
> > question: what's the risk of this change affecting existing queries?
> 
> @linzhou-db This affects access of delta sharing storage URLs, but the testing we have is pretty comprehensive since it brings up a local proxy and storage server.
> 
> There are no customers using both the never.use.https and noProxyHosts today, so the new parts where both http clients will be used will only with the upcoming DB URLs testing.
> 
> I am trying to also build a custom image and run it in serverless to make sure it all works, but there may be other issues (unrelated to this).

Approved given this will/should not affect existing prod queries.

littlegrasscao:
> @littlegrasscao could you take a look at the test failure related to "spark read limit"?

java.lang.NullPointerException: Cannot invoke "java.io.File.getCanonicalPath()" because the return value of "io.delta.sharing.spark.DeltaSharingSuite.testProfileFile()" is null

From the code, I don't know why this could be null. Let me edit the test case to force generate a profile file.

client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystemHttpClientBuilder.scala
moderakh: If I remember correctly `client/src/main/scala/io/delta/sharing/client/DeltaSharingFileSystem.scala` is instantiated once per spark session not per request. so likely we do not instantiate this per request but rather per spark session which should be fine.

moderakh (approved):

client/src/test/scala/io/delta/sharing/client/DeltaSharingFileSystemSuite.scala
wchau: Actually, we do need it. It's not console println, but writing out the contents as a response to verify in the test. The scalastyle just didn't realize it's not a console println.

PR #756, opened 2025-07-23
Add registerQueryStatesInQuerySpecificCachedTable to CachedTableManager for Query-Specific State Registration

littlegrasscao:
This PR introduces a new method, `registerQueryStatesInQuerySpecificCachedTable`, to the `CachedTableManager` class in `PreSignedUrlCache.scala`. The new function allows for the atomic and thread-safe registration of query-specific states (including weak references and refresher wrappers) to an existing `QuerySpecificCachedTable` entry in the cache.

### DIFF
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -372,6 +372,48 @@ class CachedTableManager(
   }
 
   /**
+   * Adds new query states to an existing QuerySpecificCachedTable entry in the cache.
+   *
+   * This method is used when a new query (with its own queryId and refresher wrapper)
+   * wants to share the same cached table entry as other queries, but only needs to add
+   * its own state (references and refresher wrapper) without modifying the rest of the
+   * table's cache entry. If the table is not a QuerySpecificCachedTable, or does not exist,
+   * this method is a no-op.
+   *
+   * The method is atomic and thread-safe, using computeIfPresent to avoid race conditions
+   * with concurrent register or refresh operations.
+   *
+   * @param tablePath The path of the table whose query states should be updated.
+   * @param refs A sequence of weak references for the new query, used to track its usage.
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -371,6 +371,48 @@ class CachedTableManager(
...
+    tablePath: String,
+    refs: Seq[WeakReference[AnyRef]],
+    profileProvider: DeltaSharingProfileProvider
+  ): Unit = {
+    val queryId = profileProvider.getCustomQueryId()
+      .getOrElse(throw new IllegalStateException("Query ID is not defined."))
+    val refresherWrapper = profileProvider.getCustomRefresherWrapper().getOrElse(
+      throw new IllegalStateException("refresherWrapper is not defined.")
+    )
+    cache.computeIfPresent(tablePath, (_, existingTable) => existingTable match {
```
charlenelyu-db: we should still use customTablePath as the cache key to keep things consistent? Though it could be a no-op.

charlenelyu-db (approved):
So the idea is caller will check cache existence first. If exists, it'll use `registerQueryStatesInQuerySpecificCachedTable` instead of `register`, so we could avoid fetching the urls again?

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: Yes, updated to use custom table path.

Right, it's no-op since we removed queryId.

littlegrasscao:
> So the idea is caller will check cache existence first. If exists, it'll use `registerQueryStatesInQuerySpecificCachedTable` instead of `register`, so we could avoid fetching the urls again?

This is correct, but I only plan to do it for the Delta format. It also saves writes to the block manager.
I don't want to cache all the file paths in memory for the Parquet format.

PR #757, opened 2025-07-27
Add keepUrlsAfterRefsGone parameter to QuerySpecificCachedTable for flexible URL retention

littlegrasscao:
This PR introduces a new `keepUrlsAfterRefsGone` boolean parameter to the `QuerySpecificCachedTable` class and related methods in `PreSignedUrlCache.scala`. This enhancement provides more flexible control over URL retention behavior when all FileIndex references are garbage collected. New queries don't need to be concurrent to reuse the existing urls.

### DIFF
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -98,6 +98,8 @@ class CachedTable(
  *
  * @param queryStates A mapping of query identifiers to their associated weak references
  *                        and refresher wrappers.
+ * @param keepUrlsAfterRefsGone If true, URLs will be kept in cache even when all references
+ *                              are gone, and will only be cleaned up based on access expiry time.
  */
 class QuerySpecificCachedTable(
     expiration: Long,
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -98,6 +98,8 @@ class CachedTable(
  *
  * @param queryStates A mapping of query identifiers to their associated weak references
  *                        and refresher wrappers.
+ * @param keepUrlsAfterRefsGone If true, URLs will be kept in cache even when all references
+ *                              are gone, and will only be cleaned up based on access expiry time.
```
charlenelyu-db:
```suggestion
 *                              are gone, and will only be cleaned up based on cache expiry time.
```

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -534,6 +570,9 @@ class CachedTableManager(
    *                            timestamp of the idToUrl.
    * @param refreshToken an optional refresh token that can be used by the refresher to retrieve
    *                     the same set of files with refreshed urls.
+   * @param keepUrlsAfterRefsGone If true, for QuerySpecificCachedTable, URLs will be kept in cache
+   *                              even when all references are gone, and will only be cleaned up
+   *                              based on access expiry time.
```
charlenelyu-db:
```suggestion
   *                              based on cache expiry time.
```

PR #760, opened 2025-07-29
Add logging for file ID mismatches in QuerySpecificCachedTable registration

littlegrasscao:
This PR adds logging in the `registerQuerySpecificCachedTable` method to detect and log when new file IDs returned from the server are not a superset of existing cached file IDs. This helps identify potential data inconsistency issues without breaking the application flow.


littlegrasscao:
Hold on merging the PR until more thinking goes through

PR #762, opened 2025-08-05
Implement URL and query state merging in QuerySpecificCachedTable registration

littlegrasscao:
This PR modifies the `registerQuerySpecificCachedTable` method to merge URLs and query states instead of replacing them when registering to an existing cached table entry. This change is essential for streaming queries that require complex file URL management.

### DIFF
--- a/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
+++ b/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
@@ -530,39 +530,51 @@ class CachedTableManager(
             )
         }
 
-        // Validate that the new resolvedIdToUrl is a superset of the existing idToUrl
+        // Merge idToUrl maps - new URLs take precedence for same file IDs.
+        // This merging is needed by streaming queries due to complex interactions
+        // in DataSharingSource.
+        val mergedIdToUrl = querySpecificCachedTable.idToUrl ++ resolvedIdToUrl
+
+        // Log information about the merge
         val existingIds = querySpecificCachedTable.idToUrl.keySet
         val newIds = resolvedIdToUrl.keySet
-        val missingIds = existingIds -- newIds
-        if (missingIds.nonEmpty) {
-          logError(
-            s"File ID mismatch detected for table $tablePath. " +
-            s"Old file count: ${existingIds.size}, new file count: ${newIds.size}. " +
-            s"Missing file IDs: ${missingIds.mkString(", ")}. " +
-            s"There is a chance that the current query fails with file id not found error." +
+        val addedIds = newIds -- existingIds
+        val updatedIds = newIds.intersect(existingIds)
+
+        if (addedIds.nonEmpty || updatedIds.nonEmpty) {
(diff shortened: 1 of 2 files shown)


client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -530,39 +530,51 @@ class CachedTableManager(
...
-        // when all references are null. All refresh wrappers are preserved as they may hold state
-        // required by the server to refresh URLs.
-        val newQueryStates = querySpecificCachedTable.queryStates +
-          (queryId -> (refs, refresherWrapper))
+        // Merge query states - if queryId already exists, merge the refs
+        val existingQueryState = querySpecificCachedTable.queryStates.get(queryId)
+        val mergedQueryStates = existingQueryState match {
+          case Some((existingRefs, existingWrapper)) =>
+            // Merge refs, avoiding duplicates
+            val mergedRefs = (existingRefs ++ refs).groupBy(_.get).values.map(_.head).toSeq
```
charlenelyu-db: Can we use `eq` similar to [here](https://github.com/delta-io/delta-sharing/blob/main/client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala#L679-L681) instead of `groupBy` (which uses `==` I believe). There could be a difference between the two as `==` invokes `equals` and the underlying class might have it overridden.

charlenelyu-db (approved):
One comment otherwise LGTM.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -530,39 +530,51 @@ class CachedTableManager(
...
           s"Registered to an existing QuerySpecificCachedTable in cache for table $tablePath, " +
-          s"queryId ${queryId}, newQueryStates size: ${newQueryStates.size}."
+          s"queryId ${queryId}, total queryStates: ${mergedQueryStates.size}, " +
+          s"refs size after merge: ${mergedQueryStates(queryId)._1.size}."
         )
-        // Update all attributes as the new version is always more up to date and file URLs can
-        // be shared across identical queries.
+
         new QuerySpecificCachedTable(
           expiration = resolvedExpiration,
```
charlenelyu-db: The expiration should take the min of existing and current then.

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: updated to use the same logic

client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
littlegrasscao: updated the logic to 

- Old entry has URLs not in new entry, use min expiration
- All old URLs are being replaced, use new expiration

PR #763, opened 2025-08-05
Switch OSSHR service to Central Publishing Portal

littlegrasscao:
The nexus repository manager is deprecated. We need to migrate to the new Central Publishing Portal.

The process includes
1. New credential
2. new OSSRH endpoints
3. POST call to transfer a deployment

### DIFF
--- a/dev/spark_release.sh
+++ b/dev/spark_release.sh
@@ -27,13 +27,16 @@ if build/sbt "release skip-tests"; then
   git tag v$VERSION
   echo "Created git tag: v$VERSION"
 
-  # Transfer deployment to Central Publisher Portal (required for Maven-API-like plugins)
+  # Transfer deployment to Central Publisher Portal
   echo "Transferring deployment to Central Publisher Portal..."
   NAMESPACE="io.delta"
   OSSRH_BASE="https://ossrh-staging-api.central.sonatype.com"
 
   # Make POST request to upload to Portal
   # This must be done from the same IP that was used for publishing
+  # You need these environment variables to be set:
+  # export SONATYPE_USERNAME=your-username
+  # export SONATYPE_PASSWORD=your-password
   curl -X POST \
     -H "Authorization: Bearer $(echo -n "$SONATYPE_USERNAME:$SONATYPE_PASSWORD" | base64)" \
     "$OSSRH_BASE/manual/upload/defaultRepository/$NAMESPACE" \


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "1.1.0"
+version in ThisBuild := "1.1.0-SNAPSHOT"
```
linzhou-db: What about try using a difference version? 
such as `0.0.1` 

@littlegrasscao

littlegrasscao: The doc says:

> The Central Publisher Portal and the OSSRH Staging API are two different services.
> https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#triaging-issues

My guess is that the published JAR files are stored in the Central Publisher Portal, and the OSSRH Staging API is just a middleman to help with the migration. SNAPSHOT publishing is supported only in the Central Publisher Portal, but not through the OSSRH Staging API.

If that’s the case, we won’t be able to test with SNAPSHOT and will need to publish an official release instead. Additionally, it looks like any publish to the Central Publisher Portal is final—unlike Nexus, we can’t drop a staging repository.

littlegrasscao: update: I wasn't able to publish snapshot, but i can publish formal release.

dev/spark_release.sh
```diff
@@ -14,11 +14,37 @@ build/sbt clean
...
+  # You need these environment variables to be set:
+  # export SONATYPE_USERNAME=your-username
+  # export SONATYPE_PASSWORD=your-password
+  curl -X POST \
+    -H "Authorization: Bearer $(echo -n "$SONATYPE_USERNAME:$SONATYPE_PASSWORD" | base64)" \
+    "$OSSRH_BASE/manual/upload/defaultRepository/$NAMESPACE" \
+    && echo "✓ Successfully transferred to Central Publisher Portal" \
+    || echo "✗ Failed to transfer to Central Publisher Portal"
+else
+  echo "✗ Release process failed, skipping transfer to Central Publisher Portal"
```
linzhou-db, line 46: wow, nice logs.

linzhou-db (approved):

linzhou-db:
let's backport to previous versions. 
And I'm thinking do we need to test on each backport PR -- the answer is yes, right?

littlegrasscao:
> let's backport to previous versions. And I'm thinking do we need to test on each backport PR -- the answer is yes, right?

yes, i will test branch-1.3.
For older branches, we don't have any new commits. Should we test them?

linzhou-db:
> > let's backport to previous versions. And I'm thinking do we need to test on each backport PR -- the answer is yes, right?
> 
> yes, i will test branch-1.3. For older branches, we don't have any new commits. Should we test them?

good point. or we wait until @vrinda-db 's changes to do real time test.

PR #771, opened 2025-08-13
[client] Add retry based on the httpStatusErrorCode in endStreamAction

vrinda-db:
RE


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1364,8 +1364,9 @@ object DeltaSharingRestClient extends Logging {
           s"Successfully verified endStreamAction in the response" + queryIdForLogging
         )
         if(lastEndStreamAction.errorMessage != null) {
-          throw new DeltaSharingServerException("Request failed during streaming response " +
-          s"with error message ${lastEndStreamAction.errorMessage}")
+          throw new DeltaSharingServerException(
+            s"Server Exception: ${lastEndStreamAction.errorMessage}",
+            Option(lastEndStreamAction.httpStatusErrorCode).map(_.intValue))
```
linzhou-db: nit question: is it better to put error code in front?

client/src/main/scala/io/delta/sharing/spark/DeltaSharingErrors.scala
```diff
@@ -18,9 +18,20 @@ package io.delta.sharing.spark
...
 import org.apache.spark.sql.types.StructType
 
+// Common base exception class that also encapsulates the associated errorCode
+abstract class DeltaSharingExceptionWithErrorCode(
+    message: String,
+    val statusCodeOpt: Option[Int]
+) extends IllegalStateException(message)
+
+object DeltaSharingExceptionWithErrorCode {
+  def unapply(e: DeltaSharingExceptionWithErrorCode): Option[Option[Int]] = Some(e.statusCodeOpt)
```
linzhou-db, line 28: why is it `Option[Option[Int]]`?

linzhou-db: where is unapply used?

client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
```diff
@@ -65,10 +63,10 @@ private[sharing] object RetryUtils extends Logging {
 
   def shouldRetry(t: Throwable): Boolean = {
     t match {
-      case e: UnexpectedHttpStatus =>
-        if (e.statusCode == 429) { // Too Many Requests
+      case DeltaSharingExceptionWithErrorCode(Some(statusCode)) =>
```
linzhou-db, line 66: and could it be a DeltaSharingExceptionWithErrorCode with `None` as statusCodeOpt? What if that, we don't retry?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
vrinda-db: Do you mean in the error message? Would it be useful?

client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
vrinda-db: yes it can be. It shall be caught in the catch all (`case _ => false`) statement at the end below.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingErrors.scala
vrinda-db: `unapply` is an extractor (type of method) used for pattern matching (match - case syntax) in scala. `unapply` in general must return an `Option[...]` to indicate whether the pattern matches or not. So in our case, the return type is `Option[Option[Int]]`. The outermost option is just from the definition of the `unapply` extractor and the inner one is because we are extracting statusCodeOpt whose type is `Option[Int]`. 

As for the use, we are only able to do the following because of the companion extractor. 
```
t : Throwable match {
      case DeltaSharingExceptionWithErrorCode(Some(statusCode)) => ...

```

client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
linzhou-db: is it tested?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: To me, errorCode is more abstracted and quick glance of what's happening. 

Also, do we know what's the user facing error with this new definition of `DeltaSharingServerException`? do they see the errorCode? it could also be helpful for debugging.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingErrors.scala
linzhou-db: thanks for the detailed explanation!

client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
vrinda-db: yes LOC 42 in `client/src/test/scala/io/delta/sharing/client/util/RetryUtilsSuite.scala`

`assert(!shouldRetry(new DeltaSharingServerException("error", None)))`

linzhou-db: cool, thanks!

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
vrinda-db: Right now, no I don't think the errorCode will be showing in the error message since we are not overriding any of the default converters from the RuntimeException class. 

I pushed a change to pre-pend the error message with the status code.

linzhou-db: nit: L1372 can use `errorCodeOpt`

linzhou-db: And also nit UX: Does "Server Exception500" look good?

linzhou-db (approved):

JialeTomTian (approved):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
vrinda-db: resolved!

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1364,8 +1364,10 @@ object DeltaSharingRestClient extends Logging {
           s"Successfully verified endStreamAction in the response" + queryIdForLogging
         )
         if(lastEndStreamAction.errorMessage != null) {
-          throw new DeltaSharingServerException("Request failed during streaming response " +
-          s"with error message ${lastEndStreamAction.errorMessage}")
+          val errorCodeOpt = Option(lastEndStreamAction.httpStatusErrorCode).map(_.intValue)
+          throw new DeltaSharingServerException(
+            s"Server Exception${errorCodeOpt.getOrElse("")}: " +
```
linzhou-db: Server Exception[${...}]?

linzhou-db (approved):

linzhou-db:
> After this code change, getFile could run longer than before. In a streaming query, it’s possible to have multiple getBatch calls to DeltaSharingDatasource. I’m wondering whether we need to make maybeGetFileChanges thread-safe within a datasource.

Hmmm, good point, what's the consequence if not?

PR #777, opened 2025-08-18
Handle empty String nextPageToken correctly in DeltaSharingClient

nightscape:
According to PROTOCOL.md, an empty String also means end of the list, so it should be handled identically to None.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -261,7 +261,7 @@ class DeltaSharingRestClient(
     if (response != null && response.items != null) {
       shares ++= response.items
     }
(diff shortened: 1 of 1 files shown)


nightscape:
@linzhou-db @zsxwing @pranavsuku-db @chakankardb @PatrickJin-db @littlegrasscao any chance of getting a review on this?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -261,7 +261,7 @@ class DeltaSharingRestClient(
     if (response != null && response.items != null) {
       shares ++= response.items
     }
-    while (response.nextPageToken.nonEmpty) {
+    while (response.nextPageToken.exists(_.nonEmpty)) {
```
chakankardb, line 264: Can we add a unit test for this?

Issue #483, opened 2024-04-24
load_as_spark using config as variable instead of file


stevenayers-bge:
Hi all,

I want to use `load_as_spark`, but instead of saving the config credentials as a file, I want to pull them from a secrets manager (Azure Vault, AWS SecretsManager, Databricks Secret) instead of storing the file on the filesystem. I want to avoid saving the json config to the filesystem for security reasons.

For example:
```python
delta_sharing_config = dbutils.secrets.get(scope="delta-sharing", key="my-config")

df = delta_sharing.load_as_spark(delta_sharing_config + '#<share name>.<schema>.<table name>')
```

Is this possible?

Thanks

aimtsou:
@stevenayers-bge: Would you be satisfied with building the json object while bringing the secrets from key vault and any other configuration from application resource configuration or even from the keyvault?

@linzhou-db: I can work on that we can modify the reading part

stevenayers:
@aimtsou initializing the rest client works fine, for example:

```python
import delta_sharing
from delta_sharing.protocol import DeltaSharingProfile

# GET SECRET VALUE
delta_sharing_config = dbutils.secrets.get(scope="delta-sharing", key="my-config")

# PASS SECRET CONFIG WITHOUT SAVING TO FILE
profile = DeltaSharingProfile.from_json(delta_sharing_config)
client = delta_sharing.SharingClient(profile)

# List tables available in share
client.list_all_tables()
```

The issue is when you go to read the data into a dataframe. You cannot use `SharingClient` or `DataSharingProfile` to authenticate when reading the dataframe. You have to pass the file in the `load_as_spark` function URL argument:

```python
import delta_sharing

share_file = "/tmp/my-secret-config-stored-in-a-file" # security problem!

df = delta_sharing.load_as_spark(share_file + '#nep_test.reference.settlement_period_calendar')
```

stevenayers:
@aimtsou @linzhou-db this old PR would solve the issue https://github.com/delta-io/delta-sharing/pull/103

aimtsou:
I see so you are proposing something different to pass the whole JSON as a profile, I proposed to build the JSON object in python which is closer to @zsxwing suggests. But both ways can be implemneted.

stevenayers-bge:
@aimtsou i don't mind how it's done 🙂 Im not sure your PR will help though. load_as_spark doesn't interact with the Python SharingClient.

[...]

PR #573, opened 2024-09-14
Support for delta sharing profile via options

stevenayers:
Add support for delta sharing profile via options (resolves #483).

Task List:
- [x] Add options to DeltaSharingOptions
- [x] Unit tests for DeltaSharingOptions
- [x] DeltaSharingProfile from DeltaSharingOptions
- [x] Unit tests for DeltaSharingProfileProvider
- [x] Conditional logic to create from options
- [x] Unit tests for conditional logic
- [x] Integration tests
- [x] Include Examples and Documentation

### Examples

```scala
// Bearer Token
val df = spark.read.format("deltaSharing")
  .option("shareCredentialsVersion", "1")
  .option("endpoint", "https://example.com/delta-sharing")
  .option("bearerToken", "foobar")
  .option("expirationTime", "2022-01-01T00:00:00Z")
  .load("<share-name>.<schema-name>.<table-name>")
  .limit(10)

// OAuth
val df = spark.read.format("deltaSharing")
  .option("shareCredentialsVersion", "2")
  .option("shareCredentialsType", "oauth_client_credentials")
  .option("endpoint", "https://example.com/delta-sharing")
  .option("tokenEndpoint", "https://example.com/delta-sharing/oauth/token")
  .option("clientId", "abc")
  .option("clientSecret", "xyz")
  .load("<share-name>.<schema-name>.<table-name>")
  .limit(10)
```

```python
# Bearer Token
df = (
    spark.read.format("deltaSharing")
    .option("shareCredentialsVersion", "1")
    .option("endpoint", "https://example.com/delta-sharing")
    .option("bearerToken", "foobar")
    .option("expirationTime", "2022-01-01T00:00:00Z")
    .load("<share-name>.<schema-name>.<table-name>")
    .limit(10)
)

# OAuth
df = (
    spark.read.format("deltaSharing")
    .option("shareCredentialsVersion", "2")
    .option("shareCredentialsType", "oauth_client_credentials")
    .option("endpoint", "https://example.com/delta-sharing")
    .option("tokenEndpoint", "https://example.com/delta-sharing/oauth/token")
    .option("clientId", "abc")
    .option("clientSecret", "xyz")
    .load("<share-name>.<schema-name>.<table-name>")
    .limit(10)
)

# load_as_spark or load_table_changes_as_spark
import delta_sharing
from delta_sharing.protocol import DeltaSharingProfile

profile = DeltaSharingProfile( # also supports OAuth
    share_credentials_version=1,
[...]

PR #786, opened 2025-11-07
Support for delta sharing profile via options

kkoteeswaran:
Adding support for delta sharing profile via options (reviving https://github.com/delta-io/delta-sharing/pull/573)
Separate PR for Python - https://github.com/delta-io/delta-sharing/pull/789

Task List:

- [x]  Add options to DeltaSharingOptions
- [x]  Unit tests for DeltaSharingOptions
- [x]  DeltaSharingProfile from DeltaSharingOptions
- [x]  Unit tests for DeltaSharingProfileProvider
- [x]  Conditional logic to create from options
- [x]  Unit tests for conditional logic
- [x]  Integration tests
- [x]  Include Examples and Documentation

Examples
```
// Bearer Token
val df = spark.read.format("deltaSharing")
  .option("shareCredentialsVersion", "1")
  .option("endpoint", "https://example.com/delta-sharing")
  .option("bearerToken", "foobar")
  .option("expirationTime", "2022-01-01T00:00:00Z")
  .load("<share-name>.<schema-name>.<table-name>")
  .limit(10)

// OAuth
val df = spark.read.format("deltaSharing")
  .option("shareCredentialsVersion", "2")
  .option("shareCredentialsType", "oauth_client_credentials")
  .option("endpoint", "https://example.com/delta-sharing")
  .option("tokenEndpoint", "https://example.com/delta-sharing/oauth/token")
  .option("clientId", "abc")
  .option("clientSecret", "xyz")
  .load("<share-name>.<schema-name>.<table-name>")
  .limit(10)
```

cc @prodeezy

### DIFF
--- a/spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
+++ b/spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
@@ -1018,4 +1018,31 @@ class RemoteDeltaLogSuite extends SparkFunSuite with SharedSparkSession {
     val snapshot2 = deltaLog2.snapshot()
     assert(snapshot2.getTablePath.toString.split("#")(1).split("_").length == 4)
   }
+
+  test("RemoteDeltaLog path with options") {
+    lazy val tablePath = "share.schema.table"
+    lazy val shareCredentialsOptions: Map[String, String] = Map(
+      "shareCredentialsVersion" -> "1",
+      "endpoint" -> "foo",
+      "bearerToken" -> "bar",
+      "expirationTime" -> "2021-11-12T00:12:29Z"
+    )
+
+    spark.sessionState.conf.setConfString(
+      "spark.delta.sharing.client.sparkParquetIOCache.enabled", "true")
+    // Same as the table path
+    val deltaLog1 = RemoteDeltaLog(tablePath, shareCredentialsOptions)
+    assert(deltaLog1.path.toString == tablePath)
+    val snapshot1 = deltaLog1.snapshot()
+    assert(snapshot1.getTablePath.toString == tablePath)
+
+    spark.sessionState.conf.setConfString(
+      "spark.delta.sharing.client.sparkParquetIOCache.enabled", "false")
+    // Append timestamp suffix
+    // share.schema.table_yyyyMMdd_HHmmss_uuid
+    val deltaLog2 = RemoteDeltaLog(tablePath, shareCredentialsOptions)
+    assert(deltaLog2.path.toString.split("\\.")(2).split("_").length == 4)
+    val snapshot2 = deltaLog2.snapshot()
+    assert(snapshot2.getTablePath.toString.split("\\.")(2).split("_").length == 4)
+  }
 }


stevenayers:
@kkoteeswaran ❤️

prodeezy:
This PR helps resolve https://github.com/delta-io/delta-sharing/issues/483

moderakh (dismissed):
LGTM. Thank you for your contribution. Please wait for @chakankardb and @linzhou-db review as well.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1305,17 +1305,29 @@ object DeltaSharingRestClient extends Logging {
    * Parse the user provided path `profile_file#share.schema.table` to
    * ParsedDeltaSharingTablePath.
    */
-  def parsePath(path: String): ParsedDeltaSharingTablePath = {
+  def parsePath(
```
chakankardb, line 1309: Consider updating the comment in the function header.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1384,19 +1396,23 @@ object DeltaSharingRestClient extends Logging {
...
       readerFeatures: String = ""
   ): DeltaSharingClient = {
     val sqlConf = SparkSession.active.sessionState.conf
 
-    val profileProviderClass = ConfUtils.profileProviderClass(sqlConf)
-    val profileProvider: DeltaSharingProfileProvider =
+    val profileProvider: DeltaSharingProfileProvider = if (shareCredentialsOptions.nonEmpty) {
+        new DeltaSharingOptionsProfileProvider(shareCredentialsOptions)
+    } else {
+      val profileProviderClass = ConfUtils.profileProviderClass(sqlConf)
```
chakankardb, line 1410: Perhaps validate here as well that both are not specified?

chakankardb (dismissed):

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1384,19 +1396,23 @@ object DeltaSharingRestClient extends Logging {
...
+      shareCredentialsOptions: Map[String, String],
       forStreaming: Boolean = false,
       responseFormat: String = RESPONSE_FORMAT_PARQUET,
       readerFeatures: String = ""
   ): DeltaSharingClient = {
     val sqlConf = SparkSession.active.sessionState.conf
 
-    val profileProviderClass = ConfUtils.profileProviderClass(sqlConf)
-    val profileProvider: DeltaSharingProfileProvider =
+    val profileProvider: DeltaSharingProfileProvider = if (shareCredentialsOptions.nonEmpty) {
```
littlegrasscao, line 1407: - Is this O2D sharing? And recipient will use databricks DBR? 
- Is the table not imported into the databricks?
- And you only plan to run the query through spark scala?
For context, we implement a different ProfileProvider in databricks DBR to handle UC tables. And the below configUtils is needed to use the UC provider.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
chakankardb: Never mind; read the code wrong.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
kkoteeswaran: @littlegrasscao The recipient will be using databricks DBR but we don't want to limit this feature to DBR. Their databricks workspace may not be UC-enabled though. Could you please clarify which type of ProfileProvider is required by DBR?

littlegrasscao: For tables not imported into Databricks, this new provider you are adding is good enough. DBR can also use this provider to query shared tables.
For tables imported into Databricks, in the catalog explorer UI, you can see a datasource as delta sharing, that will trigger a separate Provider. But since the share is already imported into UC as a catalog, the credential should be storing inside UC, thus no option is needed. 

Overall, if your recipient just want to query a shared table using spark.option().load() in DBR, this new provider should be sufficient. We can do some testing in DBR to verify.

littlegrasscao: @kkoteeswaran 
After this change, you’ll also need to backport it to branch-1.3, then bump the version in delta.io and update the following file in Delta:
🔗 https://github.com/delta-io/delta/blob/master/sharing/src/main/scala/io/delta/sharing/spark/DeltaSharingDataSource.scala#L246

This is the DataSource used in DBR 14+. The one you’re modifying here is for DBR 13-.
The new DataSource includes the implementation for the Delta format, which supports Delta tables with advanced features like deletion vectors.

Does your open provider also support Delta tables with those features?

- If yes, you’ll need to make similar changes there as well.
- If no, then simply upgrading the delta-sharing-client version and applying the small signature changes will be sufficient.

You don’t have permission to cut a new release — I can help with that once the backport to branch-1.3 is complete.

kkoteeswaran: @littlegrasscao here is the PR to backport these changes to branch-1.3 - https://github.com/delta-io/delta-sharing/pull/798
We want these changes in DBR 14+ as well so I will open a PR in Delta with the appropriate changes after the delta-sharing release is cut. Is the pending update in Delta a pre-requisite for incorporating these changes in DBR 13-?

littlegrasscao: To clarify, which DBR versions will **definitely require** this new Option? Databricks is very strict about adding new implementations to older Spark versions.

The change you make in Delta OSS will apply to the latest OSS Spark. I’ll then need to port it into Databricks (currently, master corresponds to DBR 18) and backport it to older DBR versions as needed.

For DBR 13-, the pending change in Delta OSS isn’t required. You just need to backport this PR to branch-0.7 in this repo, and I can cut a new release and update the version in Databricks Spark.

spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
```diff
@@ -999,11 +999,12 @@ class RemoteDeltaLogSuite extends SparkFunSuite with SharedSparkSession {
          |  "bearerToken": "xxxxx"
          |}""".stripMargin, UTF_8)
     val tablePath = s"${testProfileFile.getCanonicalPath}#share.schema.table"
+    lazy val shareCredentialsOptions: Map[String, String] = Map.empty
```
linzhou-db, line 1002: Are there additional unit test can be added in this file? 

cc @littlegrasscao

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
kkoteeswaran: We will need to support DBR versions that are compatible with Spark 3.2 to 3.5 so DBR versions 10 to 16.4. Are there any other branches in addition to branch-0.7 that need this PR to be backported?

littlegrasscao: https://docs.databricks.com/aws/en/release-notes/runtime/
Only DBR 12-17 are available now. 
12 will be deprecated soon. We would recommend we do from DBR 13 to 18 and only on LTS versions.
You would need to back-port to branch-0.7, branch-1.1, branch-1.2, branch-1.3

spark/src/test/scala/io/delta/sharing/spark/RemoteDeltaLogSuite.scala
kkoteeswaran: @linzhou-db I have added additional test to verify RemoteDeltaLog with options

PR #788, opened 2025-11-08
Add logging for unexpected line and make exception public

littlegrasscao:
- Make UnexpectedHttpStatus public so caller can catch it.
- Add debug logging when getMetadata and getFile return wrong responses

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -354,7 +354,7 @@ class DeltaSharingRestClient(
       table = getFullTableName(table)
     )
     if (response.lines.size != 2) {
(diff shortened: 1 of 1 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -391,6 +392,24 @@ class DeltaSharingRestClient(
     }
   }
 
+  private def logFirst10Lines(
```
chakankardb: Call it logKLines and pass some param? The param could be some static value for now.

littlegrasscao: updated to logFirstKLines

chakankardb (approved):

PR #789, opened 2025-11-10
Support for delta sharing profile via options in python

kkoteeswaran:
Adding support for delta sharing profile via options (reviving https://github.com/delta-io/delta-sharing/pull/573)
Separate PR for Scala - https://github.com/delta-io/delta-sharing/pull/786

Task List:

- [x]  Add options to DeltaSharingOptions
- [x]  Unit tests for DeltaSharingOptions
- [x]  DeltaSharingProfile from DeltaSharingOptions
- [x]  Unit tests for DeltaSharingProfileProvider
- [x]  Conditional logic to create from options
- [x]  Unit tests for conditional logic
- [x]  Integration tests
- [x]  Include Examples and Documentation

Examples
```
# Bearer Token
df = (
    spark.read.format("deltaSharing")
    .option("shareCredentialsVersion", "1")
    .option("endpoint", "https://example.com/delta-sharing")
    .option("bearerToken", "foobar")
    .option("expirationTime", "2022-01-01T00:00:00Z")
    .load("<share-name>.<schema-name>.<table-name>")
    .limit(10)
)

# OAuth
df = (
    spark.read.format("deltaSharing")
    .option("shareCredentialsVersion", "2")
    .option("shareCredentialsType", "oauth_client_credentials")
    .option("endpoint", "https://example.com/delta-sharing")
    .option("tokenEndpoint", "https://example.com/delta-sharing/oauth/token")
    .option("clientId", "abc")
    .option("clientSecret", "xyz")
    .load("<share-name>.<schema-name>.<table-name>")
    .limit(10)
)

# load_as_spark or load_table_changes_as_spark
import delta_sharing
from delta_sharing.protocol import DeltaSharingProfile

profile = DeltaSharingProfile( # also supports OAuth
    share_credentials_version=1,
    endpoint="https://example.com/delta-sharing",
    bearer_token="foobar",
    expiration_time="2022-01-01T00:00:00Z",
)

df = delta_sharing.load_as_spark(
    f"<share-name>.<schema-name>.<table-name>",
    version=version,
    timestamp=timestamp,
    delta_sharing_profile=profile
)
```

cc @prodeezy @stevenayers

### DIFF
--- a/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
+++ b/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
@@ -194,7 +194,10 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
         Table(name = "deletion_vectors_with_dvs_dv_property_on", schema = "default", share = "share8"),
         Table(name = "dv_and_cm_table", schema = "default", share = "share8"),
         Table(name = "timestampntz_cdf_table", schema = "default", share = "share8"),
-        Table(name = "12k_rows", schema = "default", share = "share8")
+        Table(name = "12k_rows", schema = "default", share = "share8"),
+        Table(name = "timestampntz_cdf_table", schema = "default", share = "share8"),
+        Table(name = "add_columns_non_partitioned_cdf", schema = "default", share = "share8"),
+        Table(name = "add_columns_partitioned_cdf", schema = "default", share = "share8")
       )
       assert(expected == client.listAllTables().toSet)
     } finally {
@@ -1403,7 +1406,7 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     checkErrorMessage(e, "BAD REQUEST: Error Occurred During Streaming")
   }
 
-  test("version mismatch check - getMetadata with mocked response") {
+  integrationTest("version mismatch check - getMetadata with mocked response") {
     Seq(true, false).foreach { isMSTQueryFlag =>
       val requestedVersion = 5L
       val returnedVersion = 10L
@@ -1460,7 +1463,7 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     }
   }
 
-  test("version mismatch check - getFiles with mocked response") {
+  integrationTest("version mismatch check - getFiles with mocked response") {
     // Test delta format where version validation is controlled by isMSTQuery
     Seq(true, false).foreach { isMSTQueryFlag =>
       val requestedVersion = 15L
--- a/python/delta-kernel-rust-sharing-wrapper/Cargo.toml
+++ b/python/delta-kernel-rust-sharing-wrapper/Cargo.toml
@@ -2,7 +2,7 @@
 name = "delta-kernel-rust-sharing-wrapper"
 edition = "2021"
 license = "Apache-2.0"
-version = "0.2.2"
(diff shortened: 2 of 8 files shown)


prodeezy:
This PR helps resolve https://github.com/delta-io/delta-sharing/issues/483

PatrickJin-db (commented):
Looks good, but let's add a test for good measure. I can do that tomorrow since you don't have access to the cloud credentials for our integration tests.

linzhou-db (approved):
@PatrickJin-db are you thinking of a test in the same PR or a separate one?

linzhou-db:
@kkoteeswaran  in which version of delta-sharing-python would you need this?

kkoteeswaran:
@linzhou-db is there any correlation between DBR and delta-sharing-python versions? We want to make sure this is available in the version that is tied to DBR 15.4

python/delta_sharing/delta_sharing.py
```diff
@@ -148,17 +148,25 @@ def load_as_pandas(
 
 
 def load_as_spark(
-    url: str, version: Optional[int] = None, timestamp: Optional[str] = None
+    url: str,
```
PatrickJin-db, line 201: Can we add a check that the delta sharing profile cannot be in both `url` and `delta_sharing_profile`? In other words, if `url` contains a profile path, then `delta_sharing_profile` should not be populated, and vice versa.

python/delta_sharing/delta_sharing.py
```diff
@@ -214,6 +249,25 @@ def load_table_changes_as_spark(
         "`load_table_changes_as_spark` requires running in a PySpark application."
     )
     df = spark.read.format("deltaSharing").option("readChangeFeed", "true")
+    if delta_sharing_profile is not None:
```
moderakh, line 290: this block of code is duplicate same as above can we have a helper method for this?

python/delta_sharing/delta_sharing.py
```diff
@@ -172,6 +180,25 @@ def load_as_spark(
         "`load_as_spark` requires running in a PySpark application."
     )
     df = spark.read.format("deltaSharing")
+    if delta_sharing_profile is not None:
```
moderakh, line 236: this block of code is duplicate same as below can we have a helper method for this?

python/delta_sharing/delta_sharing.py
kkoteeswaran: @PatrickJin-db the `load_as_spark()` passes the `url` to Spark and the validation is handled downstream in the Scala connector's [parsePath()](https://github.com/delta-io/delta-sharing/blob/main/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala#L1331-L1356) so it is not required here. Let me know if you think otherwise.

python/delta_sharing/delta_sharing.py
kkoteeswaran: @moderakh I have pulled that block of code out to a separate method

python/delta_sharing/delta_sharing.py
linzhou-db: a) since there is no guarantee on the spark version being called by this function, there could be versions that support and not support this, what's the behavior on each?
b) even so, fail it early and locally makes more sense to me, instead of assuming on downstream behaviors, which could change without control. 

cc @PatrickJin-db @moderakh

PatrickJin-db (approved):

PR #790, opened 2025-11-10
Add PreSigned URL access logging and retry on connection reset errors

littlegrasscao:
This PR adds two improvements to enhance observability and resilience of the Delta Sharing client:

- PreSigned URL Access Logging: Adds optional logging for PreSigned URL access with table path, file ID, URL, and expiration information
- Connection Reset Retry: Enables automatic retry on connection reset errors

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
+++ b/client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
@@ -146,6 +146,8 @@ private[sharing] class RandomAccessHttpInputStream(
     } else if (contentLength > 0L && pos > this.contentLength - 1L) {
       throw new EOFException(FSExceptionMessages.CANNOT_SEEK_PAST_EOF + " " + pos)
     } else {
+      logDebug(s"Opening file $uri at pos $pos")
+
       val errorLogger = if (logPreSignedUrlAccess) {
         Some((e: Exception) => {
           e match {


client/src/main/scala/org/apache/spark/delta/sharing/PreSignedUrlCache.scala
```diff
@@ -398,6 +399,12 @@ class CachedTableManager(
         s"with expiration ${new java.util.Date(cachedTable.expiration)}")
       throw new IllegalStateException(s"cannot find url for id $fileId in table $tablePath")
     })
+
+    if (logPreSignedUrlAccess) {
```
linzhou-db: let's only log on error like this: https://github.com/delta-io/delta-sharing/commit/df8895fd63fd0d57eb714073353445fff34a1ad6#diff-cb46d3468a1f6f7fd36f58c63934cae1eee39e58ac5fe6d68cc2eae9612fc79f

littlegrasscao: just realized that there is debug level log there already: https://github.com/delta-io/delta-sharing/blob/732c8708d7572a20de32f4b72c3540916f730ac9/client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala#L148

Let me revert my change and ask customer to give this a try

linzhou-db: do you know how to check debug logs?

littlegrasscao: I don't know how to turn on debug level for OSS spark. 
Based on google search, seems like either:
- log4j.properties under $SPARK_HOME/conf
- sparkContext.setLogLevel("DEBUG")

Should we change it to logInfo and use the option?

linzhou-db: - I'm not sure about how to view debug output. 
- I prefer to only log the url upon errors, as we did before redacting. and put it behind the new option.

littlegrasscao: updated to only log when error happens and use the option

client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
```diff
@@ -147,7 +148,27 @@ private[sharing] class RandomAccessHttpInputStream(
...
+          }
+        })
+      } else {
+        None
+      }
+
+      val entity = RetryUtils.runWithExponentialBackoff(
+        numRetries,
+        maxRetryDuration,
+        onError = errorLogger
```
linzhou-db, line 170: is this used?

littlegrasscao: inside client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala,  this line prints out the log `onError.foreach(_(e))`

client/src/main/scala/io/delta/sharing/client/util/RetryUtils.scala
```diff
@@ -30,7 +30,8 @@ private[sharing] object RetryUtils extends Logging {
   def runWithExponentialBackoff[T](
       numRetries: Int,
       maxDurationMillis: Long = Long.MaxValue,
-      retrySleepInterval: Long = 1000)(func: => T): T = {
+      retrySleepInterval: Long = 1000,
+      onError: Option[Exception => Unit] = None)(func: => T): T = {
```
linzhou-db, line 34: better name it as errorLogger?

PR #791, opened 2025-11-12
Add version validation for getMetadata and getFiles with versionAsOf parameter

littlegrasscao:
This PR adds a new isMSTQuery method in ProfileProvider to control whether the client validates that the server returns the exact version requested in getMetadata and getFiles.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -496,9 +496,7 @@ class DeltaSharingRestClient(
         respondedFormat = respondedFormat
       )
     }
-    require(
-      profileProvider.isMSTQuery() || versionAsOf.isEmpty || versionAsOf.get == version
-    )
+    require(versionAsOf.isEmpty || versionAsOf.get == version)
     val protocol = JsonUtils.fromJson[SingleAction](lines(0)).protocol
     checkProtocol(protocol)
     val metadata = JsonUtils.fromJson[SingleAction](lines(1)).metaData


PROTOCOL.md
```diff
@@ -1481,6 +1481,15 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
 **{table}**: The table name to query. It's case-insensitive.
 </td>
 </tr>
+<tr>
+<td>Query Parameters</td>
+<td>
+
+**version** (type: Long, optional): If set, will return metadata as of the specified version of the table. This is only supported on tables with history sharing enabled.
```
chakankardb: History sharing is not part of the protocol. Let's rephrase:

The server may decided to return an error if the queried version is incorrect, or if server does not permit time travel.

PROTOCOL.md
```diff
@@ -1481,6 +1481,15 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
...
+
+**timestamp** (type: String, optional): If set, will return metadata as of the table version corresponding to the specified timestamp in the [Timestamp Format](#timestamp-format). This is only supported on tables with history sharing enabled.
```
chakankardb: same as above; history sharing is a databricks proprietary thing.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -364,6 +364,11 @@ class DeltaSharingRestClient(
         s"with response format ${response.respondedFormat}" + getDsQueryIdForLogging
     )
 
+    // Validate that the response version matches the requested version
```
zhu-tom: What is the current behavior without this validation when tableVersion != requestedVersion?

PROTOCOL.md
littlegrasscao: I initially thought so as well. But we actually document it in queryTable:
https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md?plain=1#L2012

Should I update there as well?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: We would let it pass, and just continue processing, so we could be returning the data not client requested...

PROTOCOL.md
littlegrasscao: Updated the wording in both getMetadata and queryTable

PROTOCOL.md
littlegrasscao: Updated the wording in both getMetadata and queryTable

PROTOCOL.md
```diff
@@ -1481,6 +1481,15 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
 **{table}**: The table name to query. It's case-insensitive.
 </td>
 </tr>
+<tr>
+<td>Query Parameters</td>
+<td>
+
+**version** (type: Long, optional): If set, will return metadata as of the specified version of the table. The server may decided to return an error if the queried version is incorrect, or if server does not permit time travel.
```
chakankardb: Nit: The server may decided to return -> The server may return

PROTOCOL.md
```diff
@@ -1481,6 +1481,15 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
...
+
+**timestamp** (type: String, optional): If set, will return metadata as of the table version corresponding to the specified timestamp in the [Timestamp Format](#timestamp-format). The server may decided to return an error if the queried version is incorrect, or if server does not permit time travel.
+</td>
```
chakankardb: Same as above (and at other places).

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
chakankardb: Could this lead to existing queries breaking in prod? We need to worry about O2D as well.

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
```diff
@@ -1400,4 +1401,119 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
...
     checkErrorMessage(e, "BAD REQUEST: Error Occurred During Streaming")
   }
+
+  test("version mismatch check - getMetadata with mocked response") {
+    Seq(true, false).foreach { validationEnabled =>
+      val requestedVersion = 5L
+      val returnedVersion = 10L
+
+      val mockProfileProvider = new DeltaSharingFileProfileProvider(
+        new org.apache.hadoop.conf.Configuration(),
```
zhu-tom: nit: Import is a bit more readable

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -111,6 +111,9 @@ object ConfUtils {
   val LOG_PRESIGNED_URL_ACCESS_CONF = "spark.delta.sharing.client.logPreSignedUrlAccess.enabled"
   val LOG_PRESIGNED_URL_ACCESS_DEFAULT = false
 
+  val VERSION_MISMATCH_CHECK_ENABLED_CONF = "spark.delta.sharing.client.versionMismatchCheckEnabled"
```
zhu-tom: should we have test in `ConfUtilsSuite` ?

zhu-tom (approved):
Minor comments but overall LGTM

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
littlegrasscao: good suggestion, add tests

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
littlegrasscao: good catch!

zhu-tom (approved):

linzhou-db:
I realize one thing: could this happen, once the spark is set through some MST, will it stay true until condition X? what is condition X? How should we unset it for other non-MST queries?

littlegrasscao:
> I realize one thing: could this happen, once the spark is set through some MST, will it stay true until condition X? what is condition X? How should we unset it for other non-MST queries?

This is a very good point! I think it through, once the spark conf is set, it's across the whole spark session, which corresponds to the whole notebook session. 

I have another idea to introduce a helper method in ProfileProvider, which is per datasource basis.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -364,6 +365,15 @@ class DeltaSharingRestClient(
...
+    if (versionMismatchCheckEnabled) {
+      require(
+        versionAsOf.isEmpty || versionAsOf.get == response.version,
+        s"The returned table version ${response.version} does not match the requested " +
+          s"versionAsOf ${versionAsOf.get} in getMetadata" + getDsQueryIdForLogging
+      )
+    }
```
chakankardb: require(enabled && (empty || mismatch)).

can we refactor this validation?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -482,7 +501,9 @@ class DeltaSharingRestClient(
         respondedFormat = respondedFormat
       )
     }
-    require(versionAsOf.isEmpty || versionAsOf.get == version)
+    if (!versionMismatchCheckEnabled) {
+      require(versionAsOf.isEmpty || versionAsOf.get == version)
+    }
```
chakankardb: Is this correct?

littlegrasscao: Yes, the same check has been applied above for MST queries. So we don't need to perform again here. 

This is an old check already there for parquet format. I add a new check above for both parquet and delta formats for MST.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: Yes, should it be require(!MST || empty || match)?

require(enabled  would fail immediately if disabled

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -482,7 +496,9 @@ class DeltaSharingRestClient(
         respondedFormat = respondedFormat
       )
     }
-    require(versionAsOf.isEmpty || versionAsOf.get == version)
+    require(
+      profileProvider.isMSTQuery() || versionAsOf.isEmpty || versionAsOf.get == version
+    )
```
chakankardb: Are we modifying the existing check ?

littlegrasscao: This check would be redundant for MST since we already check above.

We could also leave it there and not touch it if you prefer. it doesn't hurt to do double check for MST

PR #802, opened 2025-11-19
Add support for collated strings in OpConverter

stefankandic:
## Summary

  Since its 4.0 release, Spark now supports parametrizing `StringType` with different collations which define how string data is compared. This PR adds backwards-compatible support for passing collation information in Delta Sharing predicate expressions, with version-specific implementations for Spark 3.5 (Scala 2.12) and Spark 4.0 (Scala 2.13).

### Core Implementation

  1. Added `ExprContext` case class to hold expression metadata, specifically `collationIdentifier` for string comparisons
  2. Extended comparison operations to accept optional exprCtx parameter:
    - `EqualOp`, `LessThanOp`, `LessThanOrEqualOp`, `GreaterThanOp`, `GreaterThanOrEqualOp`
    - Also applies to `In` expressions which are converted to `EqualOp` chains
  3. Created version-specific `CollationExtractor` implementations:
    - Scala 2.13 (Spark 4.0): Extracts collation information from Spark's`StringType` and populates `collationIdentifier` in format: provider.collationName.icuVersion (e.g., `icu.UNICODE_CI.75.1`, `spark.UTF8_LCASE.75.1`)
    - Scala 2.12 (Spark 3.5): Does not create `collationIdentifier` and instead defaults to `UTF8_BINARY` comparisons as collations are just a writer feature and delta.
  4. Updated `OpConverter` to:
    - Call `CollationExtractor.extractCollationIdentifier()` to extract collation information

 ### Backwards Compatibility

  - The `exprCtx` parameter is optional (`Option[ExprContext] = None`), ensuring existing code continues to work
  - The valueType field remains as plain "string" (not "string collate <name>"), maintaining compatibility with older clients
  - Collation information is stored separately in `ExprContext`, allowing non-collation-aware servers to ignore it
  - Default `UTF8_BINARY` collations (non-collated strings) work on both Spark 3.5 and 4.0

 ### Validation

  Added safety checks to prevent invalid comparisons:
  - Throws `IllegalArgumentException` when comparing strings with different collations

##  Protocol Documentation

  Updated PROTOCOL.md to document the new `exprCtx` field and collationIdentifier format with examples.


client/src/main/scala/io/delta/sharing/filters/JsonPredicates.scala
```diff
@@ -43,12 +43,25 @@ object OpDataTypes {
   val supportedTypes = Set(BoolType, IntType, LongType, StringType, DateType)
   val supportedTypesV2 = supportedTypes ++ Set(FloatType, DoubleType, TimestampType)
 
+  def stringTypeWithCollation(collationName: String): String = {
+    s"string collate $collationName"
+  }
```
charlenelyu-db: nit: can we move this to test file since it's only used in testing?

stefankandic:
> Is this ready for review?

Yes it should be. Although it is still not clear how we can have this change pass the 2.12 checks given that it uses new changes added in Spark 4.0.

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
```diff
@@ -186,7 +205,7 @@ object OpConverter {
       case SqlBooleanType => OpDataTypes.BoolType
       case SqlIntegerType => OpDataTypes.IntType
       case SqlLongType => OpDataTypes.LongType
-      case SqlStringType => OpDataTypes.StringType
+      case _: SqlStringType => OpDataTypes.StringType
```
charlenelyu-db, line 209: nit: can revert

client/src/test/scala/io/delta/sharing/filters/OpConverterSuite.scala
```diff
@@ -353,4 +356,190 @@ class OpConverterSuite extends SparkFunSuite {
...
+
+    val columnOp = op.children(0).asInstanceOf[ColumnOp]
+    val literalOp = op.children(1).asInstanceOf[LiteralOp]
+    assert(columnOp.valueType == OpDataTypes.StringType)
+    assert(literalOp.valueType == OpDataTypes.StringType)
+
+    assert(op.exprCtx.isEmpty)
+  }
+
+  test("collated string UNICODE_CI equal test") {
```
charlenelyu-db: Can we also add a test for toJson and fromJson somewhere to make sure the json payload is as expected?

charlenelyu-db (dismissed):
This looks good to me and thanks for the nice test coverage! Can you also update the `PROTOCOL.md` to reflect the new field? 

image

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
```diff
@@ -207,4 +226,50 @@ object OpConverter {
...
+          val version = s"${ICU_VERSION.getMajor}.${ICU_VERSION.getMinor}"
+          Some(s"$provider.$collationName.$version")
+        } else {
+          None
+        }
+
+      case _ =>
+        None
+    }
+  }
```
chakankardb: Can we guard this change via some config/param?

client/src/test/scala/io/delta/sharing/filters/OpConverterSuite.scala
stefankandic: Done!

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
stefankandic: We can't actually. Without this change this line was only matching the case object of `StringType` class, but we want it to match it as well as each instance we create for collated types eg. `StringType("UTF8_LCASE")`

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
stefankandic: We could but since OpConverter does not have a context of `SqlConfig` we would have to change a whole lot of code and public APIs to pipe this through to here. So I would say that this is not worth doing for this specific change.

PROTOCOL.md
```diff
@@ -3291,6 +3292,14 @@ ValueType | Description
...
 "timestamp" | Represents a timestamp in [Timestamp Format](#timestamp-format).
 
+The exprCtx structure:
+
+When present, the `exprCtx` field is a JSON object that may contain the following fields:
+
+Field Name | Description
+-|-
+collationIdentifier | Optional string that specifies the collation to use for string comparisons. The format is `provider.collationName.version`. The provider is either `icu` for ICU-based collations (e.g., UNICODE_CI) or `spark` for Spark-specific collations (e.g., UTF8_LCASE). The version specifies the version of collation used (e.g. ICU version "75.1"). Example: "icu.UNICODE_CI.75.1" or "spark.UTF8_LCASE.75.1". When collationIdentifier field is absent, the default UTF8_BINARY collation is used.
+
```
chakankardb, line 3302: It would be good to add a link to delta lake collations (https://delta.io/blog/delta-lake-4-0/#collations-coming-soon) which it seems not there yet.
Perhaps we can do so once they come to delta lake.

client/src/main/scala-2.12/io/delta/sharing/filters/CollationExtractor.scala
```diff
@@ -0,0 +1,55 @@
...
+            s"'${leftStr.typeName}' vs '${rightStr.typeName}'"
+          )
+        }
+
+        val typeName = leftStr.typeName
+        if (typeName.startsWith("string collate")) {
+          // The 2.12 client depends on Spark 3.5, which does not support collations.
+          // This means we cannot extract the collation identifier. In this case, we
+          // should throw an error so the filter is not converted. This avoids applying
+          // an incorrect filter and ensures we do not return wrong results.
```
chakankardb: Optional: since we cannot extract for 2.12, we could do the following for readability:

```
def extract(..) {
  // this comment.
  validate(..)
}
private def validate(..) {
   the code here.
}
```

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
chakankardb: Consider adding this as a comment.

PROTOCOL.md
stefankandic: This makes sense! I added the link to the delta RFC which has the design linked and protocol proposal: https://github.com/delta-io/delta/issues/2894

client/src/main/scala-2.12/io/delta/sharing/filters/CollationExtractor.scala
stefankandic: Done!

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
stefankandic: Done!

chakankardb (dismissed):

charlenelyu-db (approved):
LGTM. Thanks for the change!

linzhou-db:
this seems a very old PR, should we close it? 
Closing it. @stefankandic  please re-open if needed.

PR #804, opened 2025-11-20
wip delta-kernel-rs upgrade

PatrickJin-db:


python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -126,7 +126,7 @@ impl Scan {
         &self,
         engine_interface: &PythonInterface,
     ) -> DeltaPyResult<PyArrowType<Box<dyn RecordBatchReader + Send>>> {
-        let result_schema: ArrowSchemaRef = try_get_schema(self.0.schema())?;
+        let result_schema: ArrowSchemaRef = try_get_schema(self.0.logical_schema())?;
         let results = self.0.execute(engine_interface.0.clone())?;
```
nicklan, line 130: `Scan::execute` now filters the data for you, so you don't need to do any of the filtering you were doing manually in `try_create_record_batch_iter`. 

You can see the sort of change you need to make [here](https://github.com/delta-io/delta-kernel-rs/pull/1429/files#diff-d09623ffc3b2903bcd62d9de1ab34867e9d3b17d2d0ce9ea4403a412e06cef8f). I'd recommend looking at the changes made in the `kernel/examples/read-table-single-threaded/src/main.rs` file

linzhou-db: cc @PatrickJin-db

PR #806, opened 2025-11-20
Adding post-commit CI pipeline

littlegrasscao:
Adding post-commit for main branch

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - 'branch-*'
+      - 'python-branch-*'
   workflow_dispatch:
 jobs:
   build-and-test-server:


.github/workflows/build-and-test.yml
```diff
@@ -1,5 +1,10 @@
 name: Build and Test
-on: [pull_request, workflow_dispatch]
+on:
+  pull_request:
+  push:
+    branches:
```
PatrickJin-db, line 5: just main? should we include all branches so that other branches (e.g. branch-1.2) also get post-commit test coverage, such as when we backport?

littlegrasscao: yes, let's add branch-* as well

PatrickJin-db: let's also add `python-branch-*`

littlegrasscao: sounds good, adding python branches

PR #809, opened 2025-11-21
update delta_kernel to 0.18.1

zachschuermann:
### overview
update delta_kernel dependency to `0.18.1`. includes:
1. kernel update itself + API updates (largely attempted to keep the same existing APIs/implementation)
2. bumps `arrow` to 57 and explicitly takes `object_store` dependency
3. bumps `pyo3` version to `0.26.0`

### follow-ups
- [x] we now `collect` when doing Scans. This is a Bad Thing - should try to optimize but maybe requires kernel changes if we want to stick with `execute` API. We should probably just move doing `Scan`s off of `execute` here and instead use the normal `scan_metadata`. done, fixed kernel and updated here.
- [x] need to fix a failing test (and run the myriad skipped tests) - done, thanks patrick!

related: #804

### DIFF
--- a/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
+++ b/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
@@ -1403,7 +1403,7 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     checkErrorMessage(e, "BAD REQUEST: Error Occurred During Streaming")
   }
 
-  test("version mismatch check - getMetadata with mocked response") {
+  integrationTest("version mismatch check - getMetadata with mocked response") {
     Seq(true, false).foreach { isMSTQueryFlag =>
       val requestedVersion = 5L
       val returnedVersion = 10L
@@ -1460,7 +1460,7 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     }
   }
 
-  test("version mismatch check - getFiles with mocked response") {
+  integrationTest("version mismatch check - getFiles with mocked response") {
     // Test delta format where version validation is controlled by isMSTQuery
     Seq(true, false).foreach { isMSTQueryFlag =>
       val requestedVersion = 15L


python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -41,20 +37,20 @@ impl From<KernelError> for PyKernelError {
...
 
 #[pyclass]
-struct Table(delta_kernel::Table);
+struct Table(Url);
 
 #[pymethods]
 impl Table {
     #[new]
     fn new(location: &str) -> DeltaPyResult<Self> {
         let location = Url::parse(location).map_err(KernelError::InvalidUrl)?;
```
nicklan: should this not use delta_kernel::try_parse_uri?

PatrickJin-db (approved):

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -41,20 +37,26 @@ impl From<KernelError> for PyKernelError {
...
 impl Table {
     #[new]
     fn new(location: &str) -> DeltaPyResult<Self> {
+        // location must end in a trailing / so that it gets treated as a dir
+        let location = if location.ends_with('/') {
+            location
+        } else {
+            &format!("{location}/")
+        };
         let location = Url::parse(location).map_err(KernelError::InvalidUrl)?;
```
zachschuermann, line 52: should probably update to `try_parse_uri`? and/or kernel expose better API here (i'll take an AI)

python/delta-kernel-rust-sharing-wrapper/src/lib.rs
```diff
@@ -90,29 +92,23 @@ impl ScanBuilder {
...
-    results: impl Iterator<Item = DeltaResult<ScanResult>>,
+    results: impl Iterator<Item = DeltaResult<Box<dyn delta_kernel::EngineData>>>,
     result_schema: ArrowSchemaRef,
 ) -> RecordBatchIterator<impl Iterator<Item = Result<RecordBatch, ArrowError>>> {
-    let record_batches = results.map(|res| {
-        let scan_res = res.and_then(|res| Ok((res.full_mask(), res.raw_data?)));
-        let (mask, data) = scan_res.map_err(|e| ArrowError::from_external_error(Box::new(e)))?;
+    let record_batches = results.map(|data| {
         let record_batch: RecordBatch = data
+            .map_err(|e| ArrowError::from_external_error(Box::new(e)))?
```
zachschuermann, line 106: from nick: could use `EngineDataArrowExt`

PR #816, opened 2025-11-25
Improve type safety for shareCredentialsOptions by using Map[String, String]

littlegrasscao:
## Summary
This PR improves type safety by changing `shareCredentialsOptions` from `Map[String, Any]` to `Map[String, String]` throughout the codebase, and adds explicit type conversion in DeltaSharingOptionsProfileProvider

## Problem
The integration test **"table1 passing profile with read options"** in `DeltaSharingSuite` was failing with:
```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Integer
at io.delta.sharing.client.DeltaSharingProfileProvider.validate(DeltaSharingProfileProvider.scala:70)
```


littlegrasscao:
@kkoteeswaran This is the fix for the test failure

PR #824, opened 2025-12-04
Add configuration to guard DeltaSharingOptionsProfileProvider usage

littlegrasscao:
This PR adds a new configuration option to control whether credentials can be passed via options to Delta Sharing client, providing administrators more control over how users authenticate.

### DIFF
--- a/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
+++ b/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
@@ -126,7 +126,7 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     }
     assert(e2.getMessage.contains("you must provide a profile file path"))
     assert(e2.getMessage.contains("profile_file#share.schema.table"))
-    
+
     // Reset config for other tests
     spark.conf.set(ConfUtils.OPTIONS_PROFILE_PROVIDER_ENABLED_CONF, "true")
   }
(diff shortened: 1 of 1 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1347,6 +1347,24 @@ object DeltaSharingRestClient extends Logging {
      shareCredentialsOptions: Map[String, String]): ParsedDeltaSharingTablePath = {
     val shapeIndex = path.lastIndexOf('#')
 
+    val sqlConf = SparkSession.active.sessionState.conf
+    if (!ConfUtils.optionsProfileProviderEnabled(sqlConf)) {
```
linzhou-db, line 1351: is this the earliest place to apply the check?

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -111,6 +111,9 @@ object ConfUtils {
   val LOG_PRESIGNED_URL_ACCESS_CONF = "spark.delta.sharing.client.logPreSignedUrlAccess.enabled"
   val LOG_PRESIGNED_URL_ACCESS_DEFAULT = false
 
+  val OPTIONS_PROFILE_PROVIDER_ENABLED_CONF = "spark.delta.sharing.profile.optionsProvider.enabled"
+  val OPTIONS_PROFILE_PROVIDER_ENABLED_DEFAULT = true
```
linzhou-db, line 115: so we are ok with default true?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: yes, this is right after we load spark options. And passing it into DeltaSharingClient.

client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
littlegrasscao: I am fine with using it as a kill switch.

Please share your opinion on it.

chakankardb: Given that this is low risk, it should be ok; the kill switch is just an insurance.

linzhou-db: sg

Issue #782, opened 2025-10-10
Support Directory Based Access


zhu-tom:
This is a proposal to support directory based access in the Delta Sharing Protocol.

## Motivation

The [Delta Sharing protocol](https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md) currently grants temporary access to a tables file via its [QueryTable](https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#read-data-from-a-table) API. This operation can be expensive, as the server needs to unpack the delta log, discover parquet data files needed for the query, and generate pre-signed urls for them. A petabyte scale table can have millions of data files, which can put significant load on the server and cause query performance to suffer as well.

Similar in spirit to UC OSS API for [GenerateTemporaryTableCredentials](https://github.com/unitycatalog/unitycatalog/blob/main/api/Models/GenerateTemporaryTableCredential.md), we would like to support sharing tables with Cloud Tokens, which are directory (prefix) based STS tokens that grant temporary read access to the table’s root directory. This approach bypasses the pre-signing workflow, and instead provides direct read only access to the table. The query engines that are capable of processing the delta log get direct access to it, and can optimize query performance by leveraging their custom metadata optimizations, caching and distributed metadata processing.

We propose to add directory based access to the delta sharing protocol to enrich the open sharing ecosystem further.

## Protocol Changes

The table object returned by the Delta Sharing APIs will now contain metadata regarding the table root storage location and supported access modes, along with any auxiliary locations.

### List all Tables in a Share

HTTP Request | Value
-|-
Method | `GET`
URL | `{prefix}/shares/{share}/all-tables`

<details open>
<summary><b>200: The tables were successfully returned.</b></summary>

<table>
<tr>
<th>HTTP Response</th>
<th>Value</th>
</tr>
<tr>
<td>Header</td>
<td>

`Content-Type: application/json; charset=utf-8`

</td>
</tr>
<tr>
<td>Body</td>
<td>

```json
{
  "items": [
    {
      "name": "string",
      "schema": "string",
      "share": "string",
      "shareId": "string",
      "id": "string",
      "location": "{scheme}://some/path/to/table",
      "auxiliaryLocations": [
        "{scheme}://some/path/1",
        "{scheme}://some/path/2"
      ],
      "accessModes": ["url","dir"]
    }
  ],
  "nextPageToken": "string"
}
```

Note: `location` should point to the root directory of the table.

[...]

PR #835, opened 2026-01-02
Add GenerateTemporaryTableCredential to DeltaSharingClient

zhu-tom:
# Overview
Add GenerateTemporaryTableCredential in Delta Sharing Spark Client following the proposal outlined [here](https://github.com/delta-io/delta-sharing/issues/782). 

# Tests
Unit tests.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -1105,7 +1105,29 @@ class DeltaSharingRestClient(
       )
     }
 
-    JsonUtils.fromJson[TemporaryCredentials](response(0))
+    val tempCredentials = JsonUtils.fromJson[TemporaryCredentials](response(0))
+
+    // Validate that at least one credential type is set
+    if (tempCredentials.credentials == null) {
+      throw new IllegalStateException(
+        s"Credentials object is null in response for target:$target" + getDsQueryIdForLogging
+      )
+    }
+
+    val credentials = tempCredentials.credentials
+    val hasAwsCredentials = credentials.awsTempCredentials != null
+    val hasAzureCredentials = credentials.azureUserDelegationSas != null
+    val hasGcpCredentials = credentials.gcpOauthToken != null
+
+    if (!hasAwsCredentials && !hasAzureCredentials && !hasGcpCredentials) {
+      throw new IllegalStateException(
+        s"No valid credentials found in response. At least one of " +
+        s"awsTempCredentials, azureUserDelegationSas, or gcpOauthToken must be set. " +
+        s"Response: $response" + getDsQueryIdForLogging
+      )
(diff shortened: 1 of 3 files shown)


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1071,6 +1073,32 @@ class DeltaSharingRestClient(
     response
   }
 
+  override def generateTemporaryTableCredential(table: Table): TemporaryCredentials = {
+    val encodedShareName = URLEncoder.encode(table.share, "UTF-8")
+    val encodedSchemaName = URLEncoder.encode(table.schema, "UTF-8")
+    val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
+
+    val target =
```
littlegrasscao, line 1085: do we need to send location to provider?

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
```diff
@@ -1477,6 +1482,174 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
...
+          assert(result.credentials.gcpOauthToken.oauthToken ==
+            expectedCredentials.credentials.gcpOauthToken.oauthToken)
+        }
+      } finally {
+        client.close()
+      }
+    }
+  }
+
+  integrationTest("generateTemporaryTableCredential - error on not one-line response") {
```
littlegrasscao, line 1687: Maybe also test malformed credentials or missing fields? in that case, is an exception expected?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
zhu-tom: Added the location param

client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
zhu-tom: I don't think we should throw here, make it up to clients to handle how they would like. Added some test cases

littlegrasscao: For field missing, should we assert on one of three are returned?
  awsTempCredentials: AwsTempCredentials = null,
  azureUserDelegationSas: AzureUserDelegationSas = null,
  gcpOauthToken: GcpOauthToken = null,
  
azure and gcp are fine, but for aws, if fields are missing in AwsTempCredentials, I assume DBR would fail?

zhu-tom: Sure, given they are all optional in the class definition I'll add a check

PR #836, opened 2026-01-06
Add GenerateTemporaryTableCredential to DeltaSharingClient (#835)

zhu-tom:
* Add GenerateTemporaryTableCredential to DeltaSharingClient
* Cherry pick of f65b54e9533c4c760bfeee53c8ad5ba64e134ee9

### DIFF
--- a/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
+++ b/client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
@@ -1898,7 +1898,7 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     }
   }
 
-  test("version mismatch check - getMetadata with mocked response") {
+  integrationTest("version mismatch check - getMetadata with mocked response") {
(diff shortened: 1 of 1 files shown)


client/src/test/scala/io/delta/sharing/client/DeltaSharingRestClientSuite.scala
```diff
@@ -1473,7 +1478,427 @@ class DeltaSharingRestClientSuite extends DeltaSharingIntegrationTest {
     checkErrorMessage(e, "BAD REQUEST: Error Occurred During Streaming")
   }
 
-  test("version mismatch check - getMetadata with mocked response") {
+  integrationTest("generateTemporaryTableCredential - parse responses") {
```
littlegrasscao, line 1481: Adding integrationTest will make external contributors to skip this test case in pre-commit(because they don't have access to tokens).
Usually integrationTest is added when test case needs aws credential to load delta log. Your test case doesn't seem like it?

zhu-tom: Yeah this requires a bit of refactor in my main change as well, I'll followup with the fix in main

PR #842, opened 2026-01-24
Turn on data filter predicate hints

littlegrasscao:
Set spark.delta.sharing.jsonPredicateV2Hints.enabled default to true.

This flag has been default to true since branch-1.2:
https://github.com/delta-io/delta-sharing/blob/branch-1.2/client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala#L67

### DIFF
--- a/server/src/test/scala/io/delta/sharing/server/TestResource.scala
+++ b/server/src/test/scala/io/delta/sharing/server/TestResource.scala
@@ -305,6 +305,7 @@ object TestResource {
     serverConfig.setSsl(SSLConfig(selfSigned = true, null, null, null))
     serverConfig.setEvaluatePredicateHints(true)
     serverConfig.setEvaluateJsonPredicateHints(true)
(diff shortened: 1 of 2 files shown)


client/src/main/scala/io/delta/sharing/client/util/ConfUtils.scala
```diff
@@ -61,7 +61,7 @@ object ConfUtils {
   val JSON_PREDICATE_DEFAULT = "true"
 
   val JSON_PREDICATE_V2_CONF = "spark.delta.sharing.jsonPredicateV2Hints.enabled"
-  val JSON_PREDICATE_V2_DEFAULT = "false"
```
chakankardb, line 64: Do we need to update any tests?

littlegrasscao: yes, updated 2 tests

PR #845, opened 2026-02-12
Add Directory Based Access To Protocol

### LINKED ISSUE #782: Support Directory Based Access

zhu-tom:
Update protocol to include details for GenerateTemporaryTableCredential and required metadata updates. Closes #782

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -102,6 +102,8 @@ The following table describes behavior when the **server** returns (or omits) `a
 </tr>
 </table>
 
+Note: For backward compatibility, if `accessModes` is not present, the client should assume the server only supports URL-based access. Legacy clients that do not understand `accessModes` will directly issue QueryTable requests; if the server only supports directory-based access for a table, such calls will fail.
+
 ## REST APIs
 
 Here are the list of APIs used by Delta Sharing Protocol. All of the REST APIs use [bearer tokens for authorization](https://tools.ietf.org/html/rfc6750). The `{prefix}` of each API is configurable, and servers hosted by different providers may pick up different prefixes.
@@ -849,7 +851,7 @@ Note: the `shareId` field is optional. If `shareId` is populated for a table, it
 
 Note: `location` if present must point to the root directory of the table where the delta log exists. If the server supports `dir` based access for the table, this field must be present (see `accessModes`).
 
-Note: `auxiliaryLocations` is optional and lists extra storage locations for table files (usually no more than one). These should be supported in the `auxiliaryLocation` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. Most tables use only the root directory, but if some files are stored elsewhere, the delta log in the root will include absolute paths to them. If a client can't read from an auxiliary location, it should fall back to URL access (if available) or fail the request.
+Note: `auxiliaryLocations` is optional and lists extra storage locations for table files (usually no more than one). These should be supported in the `location` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. Most tables use only the root directory, but if some files are stored elsewhere, the delta log in the root will include absolute paths to them. If a client can't read from an auxiliary location, it should fall back to URL access (if available) or fail the request.
 
 Note: `accessModes` represents the supported access modes for the table. This can be `url`, `dir`, or both. If `url` is present, the [QueryTable](#read-data-from-a-table) API should be implemented for the table. If `dir` is present, the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) API should be implemented for the table. If this field is not present, the client will assume that the server only supports url based access.
 
@@ -1101,7 +1103,7 @@ Note: the `shareId` field is optional. If `shareId` is populated for a table, it
 
 Note: `location` if present must point to the root directory of the table where the delta log exists. If the server supports `dir` based access for the table, this field must be present (see `accessModes`).
 
-Note: `auxiliaryLocations` is optional and lists extra storage locations for table files (usually no more than one). These should be supported in the `auxiliaryLocation` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. Most tables use only the root directory, but if some files are stored elsewhere, the delta log in the root will include absolute paths to them. If a client can't read from an auxiliary location, it should fall back to URL access (if available) or fail the request. 
+Note: `auxiliaryLocations` is optional and lists extra storage locations for table files (usually no more than one). These should be supported in the `location` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. Most tables use only the root directory, but if some files are stored elsewhere, the delta log in the root will include absolute paths to them. If a client can't read from an auxiliary location, it should fall back to URL access (if available) or fail the request. 
 
 Note: `accessModes` represents the supported access modes for the table. This can be `url`, `dir`, or both. If `url` is present, the [QueryTable](#read-data-from-a-table) API should be implemented for the table. If `dir` is present, the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) API should be implemented for the table. If this field is not present, the client will assume that the server only supports url based access.
 
@@ -1288,9 +1290,6 @@ Example:
       "shareId": "edacc4a7-6600-4fbb-85f3-a62a5ce6761f",
       "id": "74be6365-0fc8-4a2f-8720-0de125bb5832",
       "location": "s3://deltasharing/vaccine_share/acme_vaccine_patient_data/vaccine_patients",
-      "auxiliaryLocations": [
-        "s3://secondary/deltasharing/vaccine_share/acme_vaccine_patient_data/vaccine_patients",
-      ],
       "accessModes": ["dir"]
     }
   ],
@@ -2628,7 +2627,7 @@ Accepted timestamp format by a delta sharing server: in the ISO8601 format, in t
 
 ### Generate Temporary Table Credential
 
-This API returns Cloud Tokens, which are directory (prefix) based STS tokens that grant temporary read access to the table’s root directory. This approach bypasses the pre-signing workflow, and instead provides direct read only access to the table. The query engines that are capable of processing the delta log get direct access to it, and can optimize query performance by leveraging their custom metadata optimizations, caching and distributed metadata processing. The response follows the format of [GenerateTemporaryTableCredential](https://github.com/unitycatalog/unitycatalog/blob/main/api/Apis/TemporaryCredentialsApi.md#generatetemporarytablecredentials) in UC OSS. The `location` field is also added to introduce a potentially lightweight approach which avoids the metadata call and pre-processing the delta log. It should be the location which the credentials are generated for. Clients that do not support reading from a cloud vendor can throw an error.
+This API returns Cloud Tokens, which are directory (prefix) based STS tokens that grant temporary read access to the table’s root directory. This approach bypasses the pre-signing workflow, and instead provides direct read only access to the table. The query engines that are capable of processing the delta log get direct access to it, and can optimize query performance by leveraging their custom metadata optimizations, caching and distributed metadata processing. The response follows the format of [GenerateTemporaryTableCredential](https://github.com/unitycatalog/unitycatalog/blob/main/api/Apis/TemporaryCredentialsApi.md#generatetemporarytablecredentials) in UC OSS. The `location` should be the location which the credentials are generated for. Clients that do not support reading from the cloud vendor this location belongs to can throw an error.
 
 <table>
 <tr>
@@ -2649,8 +2648,6 @@ This API returns Cloud Tokens, which are directory (prefix) based STS tokens tha
 `Authorization: Bearer {token}`
 
 Optional: `Content-Type: application/json; charset=utf-8`
-
-Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=deletionvectors;accessModes=url,prefix`
 </td>
 </tr>
 <tr>


PROTOCOL.md
```diff
@@ -1044,6 +1060,12 @@ Note: check the format of the `name`, `schema`, and `share` fields in the sharin
 
 Note: the `shareId` field is optional. If `shareId` is populated for a table, its value should be unique across the sharing server and immutable through the table's lifecycle.
 
+Note: `location` should point to the root directory of the table.
```
chakankardb: This filed if present must point to the root directory of the table where the delta log exists. If the server supports dir based access for the table, this field must be present (see accessModes).

PROTOCOL.md
```diff
@@ -1044,6 +1060,12 @@ Note: check the format of the `name`, `schema`, and `share` fields in the sharin
...
+
+Note: `auxiliaryLocations` is an optional field which represents any auxiliary storage locations for the table. These should be supported in the `auxiliaryLocation` field of the Generate Temporary Table Credential request body
+
```
chakankardb, line 1107: These are only needed if the table's data files are stored in multiple locations. Most tables are expected to have all the files in the root directory. Some tables may have subset of data files in these additional auxiliary locations (usually no more than 1). In such cases, the delta log of the table in the root directory will contain absolute paths to data files in these auxiliary locations.

If the client does not support reading from one or more auxiliary locations, they should either fall back to url based access if the server supports it, or fail the request.

PROTOCOL.md
```diff
@@ -1044,6 +1060,12 @@ Note: check the format of the `name`, `schema`, and `share` fields in the sharin
...
+Note: `accessModes` represents the supported access modes for the table. This can be `url`, `dir`, or both. If `url` is present, the [QueryTable](#read-data-from-a-table) endpoint should be implemented for the table. If `dir` is present, the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) endpoint should be implemented for the table.
```
chakankardb: For backward compatibility, If this field is not present, the client will assume that the server only supports url based access.

PROTOCOL.md
```diff
@@ -1216,14 +1238,25 @@ Example:
...
       "name" : "vaccine_ingredients",
       "shareId": "edacc4a7-6600-4fbb-85f3-a62a5ce6761f",
-      "id": "2f9729e9-6fcf-4d34-96df-bf72b26dfbe9"
+      "id": "2f9729e9-6fcf-4d34-96df-bf72b26dfbe9",
+      "location": "s3://deltasharing/vaccine_share/acme_vaccine_ingredient_data/vaccine_ingredients",
+      "auxiliaryLocations": [
+        "s3://secondary/deltasharing/vaccine_share/acme_vaccine_ingredient_data/vaccine_ingredients",
+        "s3://tertiary2/delta_project/alt_share/alternate_schema/alternate_table",
+      ],
+      "accessModes": ["url","dir"]
```
chakankardb, line 1284: In this example, let's only use location.

PROTOCOL.md
```diff
@@ -1740,6 +1773,12 @@ delta-table-version: 123
 {
   "metaData": {
     "id": "f8d5c169-3d01-4ca3-ad9e-7dc3355aedb2",
+    "location": "s3://my-bucket/tables/customer",
+    "auxiliaryLocations": [
+       "s3://secondary-bucket/tables/customer",
+       "s3://other-bucket/tables/customer"
```
chakankardb: Let's only use 1  aux location since we don't really support multiple.

PROTOCOL.md
```diff
@@ -2552,6 +2591,324 @@ content-type: application/x-ndjson; charset=utf-8
...
+**{schema}**: The schema name to query. It's case-insensitive.
+
+**{table}**: The table name to query. It's case-insensitive.
+</td>
+</tr>
+<tr>
+<td>Request Body</td>
+<td>
+
+The `location` field is optional and specifies the location URL path to generate temporary credentials for. If this field is not provided, the response should contain credentials for the table's main location. If the main location is specified the server should still respond with the credential.
```
chakankardb: Perhaps add more details:

This api should be called for the root location as well as all the aux locations. If a table has auxiliary locations and a client does not support reading from multiple locations, they should either fall back to url based access via QueryTable api or throw an error.

PROTOCOL.md
```diff
@@ -803,6 +808,12 @@ Note: check the format of the `name`, `schema`, and `share` fields in the sharin
 
 Note: the `shareId` field is optional. If `shareId` is populated for a table, its value should be unique across the sharing server and immutable through the table's lifecycle.
 
+Note: `location` if present must point to the root directory of the table where the delta log exists. If the server supports `dir` based access for the table, this field must be present (see `accessModes`).
+
+Note: `auxiliaryLocations` is optional and lists extra storage locations for table files (usually no more than one). These should be supported in the `auxiliaryLocation` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. Most tables use only the root directory, but if some files are stored elsewhere, the delta log in the root will include absolute paths to them. If a client can't read from an auxiliary location, it should fall back to URL access (if available) or fail the request.
```
chakankardb: Is there an auxiliaryLocation field in GTTC? (it is just location).

PROTOCOL.md
```diff
@@ -1044,6 +1060,14 @@ Note: check the format of the `name`, `schema`, and `share` fields in the sharin
 
 Note: the `shareId` field is optional. If `shareId` is populated for a table, its value should be unique across the sharing server and immutable through the table's lifecycle.
 
+Note: `location` if present must point to the root directory of the table where the delta log exists. If the server supports `dir` based access for the table, this field must be present (see `accessModes`).
+
+Note: `auxiliaryLocations` is optional and lists extra storage locations for table files (usually no more than one). These should be supported in the `auxiliaryLocation` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. Most tables use only the root directory, but if some files are stored elsewhere, the delta log in the root will include absolute paths to them. If a client can't read from an auxiliary location, it should fall back to URL access (if available) or fail the request. 
+
+If the client does not support reading from one or more auxiliary locations, they should either fall back to url based access if the server supports it, or fail the request.
```
chakankardb: This block seems to be missing from the earlier comments.
Is there a way to dedup these comments?

PROTOCOL.md
```diff
@@ -63,6 +63,45 @@ This document is a specification for the Delta Sharing Protocol, which defines t
...
 - Sharing Server: A server that implements this protocol.
 
+### Access Modes
+
+Delta Sharing supports two ways for clients to read table data: **[URL-based access](#read-data-from-a-table)** and **[directory-based access](#generate-temporary-table-credential)**. With URL-based access (`url`), the server returns pre-signed URLs for individual data files and the client fetches them via the [Query Table](#read-data-from-a-table) API. With directory-based access (`dir`), the server issues temporary cloud credentials (e.g., STS) via the [Generate Temporary Table Credential](#generate-temporary-table-credential) API so the client can read from the table’s root (and optional [auxiliary](#generate-temporary-table-credential) locations); the client then reads the Delta log and data files using the cloud storage API. Table metadata returned by [Query Table Metadata](#query-table-metadata) (and list-table APIs) includes an `accessModes` array indicating which modes the server supports for that table. The client reads `accessModes` from the server’s response and chooses how to read the table based on that and its own capabilities (e.g., URL-only, dir-only, or both).
```
littlegrasscao, line 68: is our client current able to handle auxiliary location?

PROTOCOL.md
```diff
@@ -63,6 +63,45 @@ This document is a specification for the Delta Sharing Protocol, which defines t
...
+<th>Server returns accessModes=url</th>
+<th>Server returns accessModes=dir</th>
+<th>Server returns accessModes=url,dir</th>
+</tr>
+<tr>
+<th>Client only supports URL-based access</th>
+<td>Proceeds with URL-based access.</td>
+<td>Proceeds with URL-based access.</td>
+<td>Must fail or error (server does not offer URL access).</td>
+<td>Proceeds with URL-based access.</td>
```
littlegrasscao, line 87: why are there 3 
<td>Proceeds with URL-based access.</td>?

PROTOCOL.md
```diff
@@ -63,6 +63,45 @@ This document is a specification for the Delta Sharing Protocol, which defines t
...
+</tr>
+<tr>
+<th>Client only supports directory-based access</th>
+<td>Must fail or error (server implies URL-only when accessModes is omitted).</td>
+<td>Must fail or error (server only supports URL).</td>
+<td>Proceeds with directory-based access.</td>
+<td>Proceeds with directory-based access.</td>
```
littlegrasscao, line 94: duplicate <td>Proceeds with directory-based access.</td>?

PROTOCOL.md
```diff
@@ -63,6 +63,45 @@ This document is a specification for the Delta Sharing Protocol, which defines t
...
+</tr>
+<tr>
+<th>Client supports both URL and directory access</th>
+<td>Proceeds with URL-based access.</td>
+<td>Proceeds with URL-based access.</td>
```
littlegrasscao, line 99: duplicate <td>Proceeds with URL-based access.</td>?

PROTOCOL.md
littlegrasscao: hmm, we haven't implemented auxiliaryLocation in client yet?

PROTOCOL.md
```diff
@@ -2552,6 +2625,324 @@ content-type: application/x-ndjson; charset=utf-8
 ### Timestamp Format
 Accepted timestamp format by a delta sharing server: in the ISO8601 format, in the UTC timezone, such as `2022-01-01T00:00:00Z`.   
 
+
+### Generate Temporary Table Credential
+
+This API returns Cloud Tokens, which are directory (prefix) based STS tokens that grant temporary read access to the table’s root directory. This approach bypasses the pre-signing workflow, and instead provides direct read only access to the table. The query engines that are capable of processing the delta log get direct access to it, and can optimize query performance by leveraging their custom metadata optimizations, caching and distributed metadata processing. The response follows the format of [GenerateTemporaryTableCredential](https://github.com/unitycatalog/unitycatalog/blob/main/api/Apis/TemporaryCredentialsApi.md#generatetemporarytablecredentials) in UC OSS. The `location` field is also added to introduce a potentially lightweight approach which avoids the metadata call and pre-processing the delta log. It should be the location which the credentials are generated for. Clients that do not support reading from a cloud vendor can throw an error.
```
littlegrasscao: > The `location` field is also added to introduce a potentially lightweight approach which avoids the metadata call and pre-processing the delta log.

Why would location avoid metadata call? The client would still need to call one of the metadata api to get location and pass to GTTC?

PROTOCOL.md
```diff
@@ -2552,6 +2625,324 @@ content-type: application/x-ndjson; charset=utf-8
...
+</tr>
+<tr>
+<td>Headers</td>
+<td>
+
+`Authorization: Bearer {token}`
+
+Optional: `Content-Type: application/json; charset=utf-8`
+
+Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=deletionvectors;accessModes=url,prefix`
```
littlegrasscao: The capability doesn't matter here for GTTC call?

PROTOCOL.md
chakankardb: No; currently our server is also not able to issue them.

PROTOCOL.md
chakankardb: It is a table; perhaps view it using: https://github.com/zhu-tom/delta-sharing/blob/2ac6ffe7bfbb09c017614cdf578143de3910cf49/PROTOCOL.md

PROTOCOL.md
```diff
@@ -63,6 +63,45 @@ This document is a specification for the Delta Sharing Protocol, which defines t
...
+<td>Proceeds with directory-based access.</td>
+</tr>
+<tr>
+<th>Client supports both URL and directory access</th>
+<td>Proceeds with URL-based access.</td>
+<td>Proceeds with URL-based access.</td>
+<td>Proceeds with directory-based access.</td>
+<td>Client may choose either mode.</td>
+</tr>
+</table>
```
chakankardb, line 103: Perhaps add a comment about backward compatibility; legacy clients will ignore the accessModes and directly issue QueryTable, and the server only supports dir based access such calls will fail.

PROTOCOL.md
chakankardb: Let's simplify/rephrase: 
The `location` should be the location which the credentials are generated for. Clients that do not support reading from the cloud vendor this location belongs to can throw an error.

PROTOCOL.md
chakankardb: +1, we don't need  the capabilities header right since  this call already assumes dir based access (responseFormat, readerFeatures etc not relevant since we are giving direct access).

PROTOCOL.md
```diff
@@ -2725,6 +3116,9 @@ Field Name | Data Type | Description | Optional/Required
...
+location | String | The root directory of the table. For `dir` tables, the server must support directory based access for the table and metadata must include the directory of the table. In the case that the client does not support directory based access, this field is optional. However, we recommend that this field be included to support recipients with network restrictions to allow these locations to be accessed. | Optional
+auxiliaryLocations | Array<String> | An array containing any auxiliary storage locations for the table. This field is optional; if not present, there are no auxiliary locations. These should be supported in the `location` field of the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) request body. | Optional
+accessModes | Array<String> | An array of the supported access modes for the table. Valid values are `url` and/or `dir`. If `url` is present, the [QueryTable](#read-data-from-a-table) API should be implemented for the table. If `dir` is present, the [GenerateTemporaryTableCredential](#generate-temporary-table-credential) API should be implemented for the table. | Optional
```
chakankardb, line 3118: Is this needed anymore?

PROTOCOL.md
littlegrasscao: If both server/client don't support auxiliary location, do we need to mention it here?

PROTOCOL.md
zhu-tom: `accessModes` is needed for reconciliation in the case the table is created in table level reconciliation and it may be helpful for clients to have this in the table metadata response.

PROTOCOL.md
chakankardb: SAP wanted us to add these for future compatibility.

chakankardb (approved):

PR #846, opened 2026-02-14
Support spark.option('callerOrg') in Delta Sharing client

littlegrasscao:
- Add userAgent option to DeltaSharingOptions and pass through RemoteDeltaLog and DeltaSharingRestClient to the HTTP User-Agent header (after QueryId, before Hadoop segment). It will show as `callerOrg-xxx`.
- Add tests: DeltaSharingOptionsSuite (option parsing), DeltaSharingRestClientSuite (RemoteDeltaLog path and header assertion).

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -1311,9 +1311,14 @@ class DeltaSharingRestClient(
     } else {
       "Delta-Sharing-Spark"
     }
-    val customUserAgentPart = if (callerOrg.nonEmpty) s" callerOrg-$callerOrg" else ""
-    s"$sparkAgent/$VERSION" + s" $sparkVersionString" + s" $getQueryIdString" +
-      customUserAgentPart + USER_AGENT
+    val parts = Seq(
+      Some(s"$sparkAgent/$VERSION"),
+      Some(sparkVersionString),
+      Some(getQueryIdString),
+      Option(callerOrg).filter(_.nonEmpty).map(o => s"callerOrg-$o"),
+      Some(USER_AGENT)
+    ).flatten
+    parts.mkString(" ")
   }
 
   private def getQueryIdString: String = {
@@ -1599,7 +1604,7 @@ object DeltaSharingRestClient extends Logging {
         java.lang.Integer.valueOf(tokenExchangeMaxRetries),
         java.lang.Integer.valueOf(tokenExchangeMaxRetryDurationInSeconds),
         java.lang.Integer.valueOf(tokenRenewalThresholdInSeconds),
-        callerOrg.filter(_.nonEmpty).getOrElse("")
+        callerOrg.getOrElse("")
       ).asInstanceOf[DeltaSharingClient]
   }
 }


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1593,7 +1598,8 @@ object DeltaSharingRestClient extends Logging {
         java.lang.Long.valueOf(asyncQueryMaxDurationMillis),
         java.lang.Integer.valueOf(tokenExchangeMaxRetries),
         java.lang.Integer.valueOf(tokenExchangeMaxRetryDurationInSeconds),
-        java.lang.Integer.valueOf(tokenRenewalThresholdInSeconds)
+        java.lang.Integer.valueOf(tokenRenewalThresholdInSeconds),
+        callerOrg.filter(_.nonEmpty).getOrElse("")
```
zhu-tom: Nit: seems equivalent to `callerOrg.getOrElse("")`?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1310,7 +1311,9 @@ class DeltaSharingRestClient(
     } else {
       "Delta-Sharing-Spark"
     }
-    s"$sparkAgent/$VERSION" + s" $sparkVersionString" + s" $getQueryIdString" + USER_AGENT
+    val customUserAgentPart = if (callerOrg.nonEmpty) s" callerOrg-$callerOrg" else ""
+    s"$sparkAgent/$VERSION" + s" $sparkVersionString" + s" $getQueryIdString" +
```
zhu-tom: List is getting big making this line a bit hard to read, I wonder if `Seq(...).mkString(" ")` is slightly cleaner

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -1310,7 +1311,9 @@ class DeltaSharingRestClient(
     } else {
       "Delta-Sharing-Spark"
     }
-    s"$sparkAgent/$VERSION" + s" $sparkVersionString" + s" $getQueryIdString" + USER_AGENT
+    val customUserAgentPart = if (callerOrg.nonEmpty) s" callerOrg-$callerOrg" else ""
```
zhu-tom: Any reason you decided to omit instead of include with `not-set` like it is done for QueryId?

littlegrasscao: Not something specific. 
I think initially it would be only adobe using it. Adding it for everyone would increase unnecessary log size.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: good suggestion, updating it

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: yes, this filter is redundant.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
zhu-tom: I see, that raises another question: do we truncate the header anywhere? Would a long callerOrg result in the loss of the `USER_AGENT` part in logs?

littlegrasscao: yes we truncate in delta sharing kpi log(no truncation in audit log), that's why i want to put it at this spot, instead of the end.

zhu-tom (approved):

PR #848, opened 2026-02-21
Server Implementation for Directory Based Access

zhu-tom:
This implements directory based access in the Delta Sharing Reference Server by adding support for the new temporary table credentials endpoint. The credential vending mirrors the Unity Catalog OSS implementation. Run and tested locally.

### DIFF
--- a/server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
+++ b/server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
@@ -3724,13 +3724,13 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
     )
   }
 
-  integrationTest("table1 - cannot query table changes") {
+  integrationTest("table2 - cannot query table changes (historyShared is false)") {
     assertHttpError(
-      url = requestPath("/shares/share1/schemas/default/tables/table1/changes"),
+      url = requestPath("/shares/share2/schemas/default/tables/table2/changes"),
       method = "GET",
       data = None,
       expectedErrorCode = 400,
-      expectedErrorMessage = "cdf is not enabled on table share1.default.table1"
+      expectedErrorMessage = "cdf is not enabled on table share2.default.table2"
     )
   }
 
@@ -3909,7 +3909,7 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
     verifyPreSignedUrl(actualFiles(0).url, 568)
   }
 
-  integrationTest("temporary-table-credentials endpoint") {
+  integrationTest("temporary-table-credentials endpoint - AWS") {
     val url = requestPath("/shares/share1/schemas/default/tables/table1/temporary-table-credentials")
     val response = readHttpContent(
       url,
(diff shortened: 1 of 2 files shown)


server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -611,6 +619,51 @@ class DeltaSharingService(serverConfig: ServerConfig) {
     )
   }
 
+  @Post("/shares/{share}/schemas/{schema}/tables/{table}/temporary-table-credentials")
+  def generateTemporaryTableCredential(
+      req: HttpRequest,
+      @Param("share") share: String,
+      @Param("schema") schema: String,
+      @Param("table") table: String,
+      aggregatedRequest: AggregatedHttpRequest): HttpResponse = processRequest {
```
littlegrasscao, line 628: should this request be the same format as listFiles:
```
 @Post("/shares/{share}/schemas/{schema}/tables/{table}/query")
  @ConsumesJson
  def listFiles(
      req: HttpRequest,
      request: QueryTableRequest): HttpResponse = processRequest
```

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -611,6 +619,51 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+    val request = parseTemporaryTableCredentialRequest(aggregatedRequest)
+    val tableConfig = sharedTableManager.getTable(share, schema, table)
+    if (!tableConfig.historyShared) {
```
littlegrasscao, line 631: how about partition filter?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -611,6 +619,51 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+    }
+    val uri = request.location
+      .map(s => new URI(s))
+      .getOrElse(new URI(tableConfig.getLocation))
+    val conf = new Configuration()
+    val vendor = new StorageCredentialVendor(conf)
+    val result = vendor.vendCredential(
+      uri,
+      Set(Privilege.SELECT),
+      serverConfig.preSignedUrlTimeoutSeconds
```
littlegrasscao, line 644: should we reuse url expiration here or a different one?

server/src/main/scala/io/delta/sharing/server/credential/gcp/GcpCredentialVendor.scala
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.hadoop.fs.Path
+
+import io.delta.sharing.server.credential.CredentialContext
+import io.delta.sharing.server.model.{Credentials, GcpOauthToken}
+
+/**
+ * Vends GCP OAuth2 access tokens, optionally downscoped to the context location via CAB.
+ * Mirrors Unity Catalog's GcpCredentialVendor.
+ */
+class GcpCredentialVendor {
```
littlegrasscao, line 33: are these vendor code directly copied from OSS UC?

Do they actually work or just mocking?

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -3908,4 +3908,41 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
...
+      result.credentials.gcpOauthToken != null
+    assert(hasCreds, "Expected one of awsTempCredentials, azureUserDelegationSas, or gcpOauthToken")
+  }
+
+  integrationTest("temporary-table-credentials with location override") {
+    val url = requestPath("/shares/share1/schemas/default/tables/table1/temporary-table-credentials")
+    val response = readHttpContent(
+      url,
+      Some("POST"),
+      Some("""{"location": "s3a://test-bucket/custom/path"}"""),
```
littlegrasscao, line 3987: is this a real location? and does the result contain real credentials?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
zhu-tom: I don't see a notion of that for TableConfig in OSS, so it doesn't seem to apply in this case

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
zhu-tom: The `@ConsumesJson` doesn't handle the possibility of empty POST body well (but we allow it for GTTC), so this is the cleanest solution I found.

PR #854, opened 2026-02-26
Server Implementation for Directory Based Access

zhu-tom:
This implements directory based access in the Delta Sharing Reference Server by adding support for the new temporary table credentials endpoint. The credential vending mirrors the Unity Catalog OSS implementation. Run and tested locally.

### DIFF
--- a/cli/test_cli.py
+++ b/cli/test_cli.py
@@ -1,0 +1,617 @@
+#!/usr/bin/env python3
+"""Unit tests for delta-sharing CLI."""
+
+import argparse
+import json
+import os
+import tempfile
+import unittest
+
+import cli
+
+
+class TestUrl(unittest.TestCase):
+    def test_simple(self):
+        self.assertEqual(
+            cli._url("https://host/prefix", "shares"),
+            "https://host/prefix/shares",
+        )
+
+    def test_trailing_slash(self):
+        self.assertEqual(
+            cli._url("https://host/prefix/", "shares"),
+            "https://host/prefix/shares",
+        )
+
+    def test_multiple_parts(self):
+        self.assertEqual(
+            cli._url("https://h", "shares", "s1", "schemas", "sc1", "tables"),
+            "https://h/shares/s1/schemas/sc1/tables",
+        )
+
+    def test_query_params(self):
+        url = cli._url("https://h", "shares", maxResults=10, pageToken="abc")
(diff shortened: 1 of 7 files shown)


server/src/main/scala/io/delta/sharing/server/credential/CredentialContext.scala
```diff
@@ -0,0 +1,95 @@
...
+ * limitations under the License.
+ */
+
+package io.delta.sharing.server.credential
+
+import java.net.URI
+
+/**
+ * Storage scheme used to select which cloud credential vendor to use.
+ * Mirrors Unity Catalog's UriScheme for credential vending.
```
chakankardb: Is this Open UC ? If so, perhaps mention that; otherwise we should perhaps remove this ref.

server/src/main/scala/io/delta/sharing/server/credential/CredentialContext.scala
```diff
@@ -0,0 +1,95 @@
...
+  type Privilege = Value
+  val SELECT, UPDATE = Value
+}
+
+/**
+ * Context for generating cloud storage credentials.
+ * Encapsulates scheme, storage base, privileges, and locations so that
+ * cloud-specific vendors can produce appropriately scoped temporary credentials.
+ *
+ * Mirrors Unity Catalog's CredentialContext (path + privileges -> context).
```
chakankardb: Same as above.

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -611,6 +619,51 @@ class DeltaSharingService(serverConfig: ServerConfig) {
...
+    val request = parseTemporaryTableCredentialRequest(aggregatedRequest)
+    val tableConfig = sharedTableManager.getTable(share, schema, table)
+    if (!tableConfig.historyShared) {
+      throw new DeltaSharingIllegalArgumentException(
+        "Temporary table credentials are not supported because history sharing is not enabled " +
+        s"on table: $share.$schema.$table")
+    }
+    val uri = request.location
+      .map(s => new URI(s))
+      .getOrElse(new URI(tableConfig.getLocation))
```
chakankardb, line 638: should we validate that the location matches table's location ?

PR #856, opened 2026-03-02
OSS server: support fileidhash on queryTable and queryCDF

littlegrasscao:
If clients send fileIdHash in the HTTP header, use the specified hash function to generate the fileId (either md5 or sha256). Also return the hash function used in the HTTP response header. If not specified, use md5 for Parquet format responses and sha256 for Delta format responses.

Add test cases to cover this change.

### DIFF
--- a/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
+++ b/server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
@@ -647,6 +647,7 @@ class DeltaSharingService(serverConfig: ServerConfig) {
       normalized
     }
     // scalastyle:on caselocale
+  }
 
   @Post("/shares/{share}/schemas/{schema}/tables/{table}/temporary-table-credentials")
   def generateTemporaryTableCredential(


server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
```diff
@@ -652,7 +688,31 @@ object DeltaSharingService {
...
   val DELTA_SHARING_RESPONSE_FORMAT = "responseformat"
+
+  /**
+   * Hash a file path to produce a file ID. When the client explicitly requests an algorithm
+   * via `fileIdHash`, that algorithm is used. Otherwise the default is sha256 for the delta
+   * response format and md5 for parquet (backward compatibility).
+   */
+  def hashFileId(
+      path: String,
+      fileIdHash: Option[String],
```
zhu-tom: Should we throw if fileIdHash is non-empty but not sha256 or md5?

server/src/main/scala/io/delta/sharing/kernel/internal/DeltaSharedTableKernel.scala
```diff
@@ -723,10 +729,10 @@ class DeltaSharedTableKernel(
       numRecords -= dvDescriptor.cardinality
     }
 
+    val fileId = DeltaSharingService.hashFileId(addFileVectorPath, fileIdHash, respondedFormat)
```
zhu-tom: Should `hashFileId` be in a helper separate from DeltaSharingService? Feels like an anti-pattern for DeltaSharingService to depend on DeltaSharedTableKernel while also having DeltaSharedTableKernel depend on DeltaSharingService

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
```diff
@@ -2998,6 +3168,28 @@ class DeltaSharingServiceSuite extends FunSuite with BeforeAndAfterAll {
     }
   }
 
+  integrationTest("queryCDF file id signed with requested fileidhash (md5 and sha256)") {
+    val baseUrl = requestPath(
+      s"/shares/share8/schemas/default/tables/cdf_table_cdf_enabled/changes?startingVersion=0&endingVersion=3")
+    val responseMd5 = readNDJson(baseUrl, Some("GET"), None, Some(0), RESPONSE_FORMAT_PARQUET,
+      requestFileIdHash = Some("md5"))
+    val responseSha256 = readNDJson(baseUrl, Some("GET"), None, Some(0), RESPONSE_FORMAT_PARQUET,
+      requestFileIdHash = Some("sha256"))
```
zhu-tom, line 3189: Should we assert the headers here too?

server/src/main/scala/io/delta/sharing/server/DeltaSharingService.scala
littlegrasscao: it throws under getRequestFileIdHash

server/src/test/scala/io/delta/sharing/server/DeltaSharingServiceSuite.scala
littlegrasscao: This readDNJson doesn't return response header.

The test above that uses readNDJsonWithFileIdHashHeader validates response header

server/src/main/scala/io/delta/sharing/kernel/internal/DeltaSharedTableKernel.scala
littlegrasscao: moved to a query util file

PR #865, opened 2026-03-20
Document fileidhash request header in PROTOCOL.md

littlegrasscao:
Add a new "File ID Hash Header" section to PROTOCOL.md documenting the fileidhash HTTP request header, which allows clients to request a specific hash algorithm for generating file id values in responses.


PROTOCOL.md
```diff
@@ -3042,6 +3043,43 @@ The server can:
...
+
+When the client does not send the header, the server should
+not include `fileidhash` in the response (the client does not validate the response header in this case).
+
+### Client Verification
+
+When a client sends the `fileidhash` header, it should verify that the server echoes the same value in
+the response. If the response header is missing or has a different value, the client should treat
+this as an error (the server may not support the feature or may have used a different algorithm).
+
```
chakankardb, line 3075: Do we need this?

Why can't the client re-generate this hash and use that to order files as needed?

littlegrasscao: Do you mean the client should rewrite file IDs in the server's response?
A few concerns with that approach:

1. Scale — The file list could be large, making client-side regeneration expensive.
2. Unknown ID scheme — For open servers, file IDs may not be derived from file paths. Forcing client-side regeneration based on file paths could produce incorrect IDs. 
3. Source of truth — The server should own file ID generation and keep them in sync.

PROTOCOL.md
```diff
@@ -3042,6 +3043,44 @@ The server can:
...
+[Query Table Changes](#read-change-data-feed-from-a-table).
+
+### Supported Values
+
+Value | Description
+-|-
+`parquet` | File `id` values use the Parquet-aligned scheme: a stable, deterministic encoding of the file (encoding details are implementation-defined).
+`delta` | File `id` values use the Delta-aligned scheme: a stable, deterministic encoding of the file (encoding details are implementation-defined).
+
+The value is case-insensitive. The server must return HTTP 400 if an unsupported value is provided.
```
linzhou-db, line 3063: for unsupported value: could the server ignore it and return the value for what's used?

PROTOCOL.md
```diff
@@ -3042,6 +3043,44 @@ The server can:
...
+
+### Delta response format and structured streaming
```
linzhou-db: my preference is no need to add this section.

PROTOCOL.md
```diff
@@ -3042,6 +3043,44 @@ The server can:
...
+
+The `fileidhash` header matters most when the response uses the **Delta** format. A structured streaming query needs to **migrate** from the Parquet to the Delta response format when the shared table turns on **advanced Delta reader features** that the Parquet format cannot handle, for example Deletion Vectors. In Spark, that migration can be handled automatically for streaming when [`spark.sql.delta.sharing.streamingAutoResolveResponseFormat`](https://github.com/delta-io/delta/blob/194e707e64b8aadca2fd1b06ec35a641741d90d0/spark/src/main/scala/org/apache/spark/sql/delta/sources/DeltaSQLConf.scala#L3020) is enabled. Sending a consistent `fileidhash` across those requests preserves **data integrity**.
+
+### Response Header
+
+When the client sends a valid `fileidhash` header, the server must echo the same header and value
+(normalized to lowercase) in the response. 
+
+When the client does not send the header, the server should
```
linzhou-db: It's ok for the server to still include the header in the response, in this case. 

if agreed, we can remove this paragraph.

PROTOCOL.md
littlegrasscao: remove section

PROTOCOL.md
littlegrasscao: that makes sense. even if server returns a response header, client just ignores it. so it doesn't matter

PROTOCOL.md
littlegrasscao: that might be confusing for the client? client sends a value, but server returns a different value. 

Also this means an unauthorized request is going through the system

linzhou-db: makes sense. could you do a quick check on what's the server behavior today for other headers? say unsupported value for responseFormat.

littlegrasscao: responseFormat is throwing an exception today.

readerFeatures doesn't have any validation.

linzhou-db: I see, responseFormat is a two way communication, readerFeatures is about client ability. 
That makes sense.

linzhou-db: The new header is also a two way communication, it makes sense to follow the pattern of responseFormat.

linzhou-db (approved):

PR #324, opened 2023-06-09
Throw exception on Trigger.availableNow in DeltaSharingSource

linzhou-db:
Throw exception on Trigger.availableNow in DeltaSharingSource

PR #881, opened 2026-03-27
Support Trigger.AvailableNow in DeltaSharingSource

zhu-tom:
## Description

  Implement native `Trigger.AvailableNow` support in `DeltaSharingSource` by adding the `SupportsTriggerAvailableNow` trait.

  Previously, when a user specified `Trigger.AvailableNow` with a Delta Sharing source, Spark's `AvailableNowSourceWrapper` would wrap the source. This wrapper calls `latestOffset(null, allAvailable)` during preparation — before `getBatch`
  sets `previousOffset` — causing the source to recompute offsets from scratch and produce duplicate rows (see #324).

  This PR fixes the issue by implementing `SupportsTriggerAvailableNow` directly, which avoids the wrapper entirely. The approach:

  - **Freeze the server version at query start**: `prepareForTriggerAvailableNow()` captures the current server version into `frozenServerVersionForAvailableNow` before setting the `isTriggerAvailableNow` flag, ensuring no new writes to the
  table after query start affect the run.
  - **Return frozen version on subsequent calls**: `getOrUpdateLatestTableVersion` returns the frozen value when in AvailableNow mode, skipping RPCs to the server. This is sufficient to bound the query — `maybeGetFileChanges` derives
  `endingVersionForQuery` from this value, so no separate guard on the fetch loop is needed.
  - **Natural termination**: the query terminates once all data through the frozen version is consumed, with no changes to the existing rate limiting (`maxFilesPerTrigger`), incremental fetching (`maxVersionsPerRpc`), or offset computation.

  Changes:
  - Add `SupportsTriggerAvailableNow` to import and `DeltaSharingSource` class definition
  - Add `isTriggerAvailableNow` and `frozenServerVersionForAvailableNow` state variables
  - Implement `prepareForTriggerAvailableNow()`
  - Guard `getOrUpdateLatestTableVersion` to return frozen version when in AvailableNow mode

  ## Tests

  5 new integration tests in `DeltaSharingSourceSuite`:

  1. **Basic AvailableNow** — verifies all data is processed in a single batch and the query self-terminates
  2. **maxFilesPerTrigger multi-batch** — verifies rate limiting produces the correct number of micro-batches with correct row counts per batch, with a full `checkAnswer` on the output
  3. **Checkpoint file counts** — verifies AvailableNow terminates with exactly one `offsets/N` and `commits/N` file per source file (no extra server fetches past the frozen version)
  4. **Clean restart — no duplicates** — verifies a second AvailableNow run on the same checkpoint emits 0 rows when the table is unchanged
  5. **Mid-stream crash restart** — simulates a crash after `offsets/0` is committed but before `commits/0` is written; verifies the engine re-runs batch 0 correctly and all rows appear with no data loss (file sink idempotency prevents
  duplicate writes)

**TESTS ARE PASSING LOCALLY**


client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -271,6 +292,13 @@ case class DeltaSharingSource(
     }
 
     val currentLatestVersion = getOrUpdateLatestTableVersion
+    // In AvailableNow mode, stop fetching once we've reached the frozen version.
+    // Using >= defensively; in practice the fetched version should only equal the frozen version
+    // since endingVersionForQuery is capped at frozenServerVersionForAvailableNow.
+    if (isTriggerAvailableNow && fromVersion > frozenServerVersionForAvailableNow) {
```
littlegrasscao: why is it > not >=?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -583,6 +583,134 @@ class DeltaSharingSourceSuite extends QueryTest
...
+    withTempDirs { (checkpointDir, outputDir) =>
+      runAvailableNowQuery(
+        withStreamReaderAtVersion(),
+        checkpointDir,
+        outputDir
+      )
+
+      // All data from the shared table should be processed.
+      val result = spark.read.format("parquet").load(outputDir.getCanonicalPath)
+      assert(result.count() > 0, "Expected data to be processed")
```
littlegrasscao: should we validate result is correct as well?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -583,6 +583,134 @@ class DeltaSharingSourceSuite extends QueryTest
...
+      )
+
+      val countAfterSecondRun =
+        spark.read.format("parquet").load(outputDir.getCanonicalPath).count()
+      assert(countAfterSecondRun === countAfterFirstRun,
+        s"Expected no new data on restart, but got $countAfterSecondRun vs $countAfterFirstRun")
+    }
+  }
+
+  integrationTest("Trigger.AvailableNow - terminates when no new data") {
```
littlegrasscao: this test and the one above looks the same?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -583,6 +583,134 @@ class DeltaSharingSourceSuite extends QueryTest
...
+        .option("checkpointLocation", checkpointDir.getCanonicalPath)
+        .trigger(Trigger.AvailableNow())
+        .start(outputDir.getCanonicalPath)
+
+      try {
+        query.processAllAvailable()
+        // With maxFilesPerTrigger=1, there should be multiple batches
+        // (one per file). Verify query completed and processed data.
+        val progress = query.recentProgress.filter(_.numInputRows != 0)
+        assert(progress.length >= 1, "Expected at least one batch with data")
```
littlegrasscao: should we assert on more than 1 batch?

zhu-tom:
@littlegrasscao this PR became out of date as the delta PR evolved. Updated with the latest which should address your comments

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -443,6 +443,172 @@ class DeltaSharingSourceSuite extends QueryTest
...
+        query.processAllAvailable()
+      } finally {
+        query.stop()
+      }
+
+      // Simulate a crash: delete commits/0 so the engine believes batch 0 was
+      // never committed to the sink and re-runs it on the next start.
+      // FileStreamSink.addBatch skips batches already in its own sink log, so
+      // the re-run does not produce duplicate rows in the output.
+      new java.io.File(checkpointDir, "commits/0").delete()
```
littlegrasscao: you delete commit, but not output data file. 
shouldn't the output data file already being generated in previous run? And running again generate duplicate data files?

zhu-tom: I think I should just remove it. processAllAvailable() runs to completion, so any post-hoc checkpoint manipulation just tests FileStreamSink idempotency, not our SupportsTriggerAvailableNow logic.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -186,6 +208,9 @@ case class DeltaSharingSource(
   // RPC. Adding a minimum interval of QUERY_TABLE_VERSION_INTERVAL_MILLIS between two consecutive
   // rpcs to avoid traffic jam on the delta sharing server.
   private def getOrUpdateLatestTableVersion: Long = {
+    if (isTriggerAvailableNow) {
```
linzhou-db, line 211: nit but interestingly: this logic here and L166/L167 depend on each other. 

is it safer to update the check to be `if (isTriggerAvailableNow && frozenServerVersionForAvailableNow != -1)` or consider some other options to reduce this dependency, or you feel this is ok.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
```diff
@@ -186,6 +208,9 @@ case class DeltaSharingSource(
...
+      return frozenServerVersionForAvailableNow
+    }
```
linzhou-db, line 213: and no more changes in the Source class, curious how does the stream stop?

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
```diff
@@ -443,6 +443,128 @@ class DeltaSharingSourceSuite extends QueryTest
...
+      checkAnswer(spark.read.format("parquet").load(outputDir.getCanonicalPath), expected)
+    }
+  }
+
+  integrationTest("Trigger.AvailableNow with maxFilesPerTrigger - success") {
+    withTempDirs { (checkpointDir, outputDir) =>
+      val query = withStreamReaderAtVersion()
+        .option("maxFilesPerTrigger", "1")
+        .load().writeStream.format("parquet")
+        .trigger(Trigger.AvailableNow())
```
linzhou-db, line 483: so this test will fail if this uses Trigger.Once?

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
zhu-tom: The only reason this would happen is if a fresh RPC was not sent for the initial prepare, which should never be the case as prepare is called at query start before any prior RPC. `getOrUpdateLatestTableVersion` validates that a valid version is returned.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
zhu-tom: Uses the existing logic in `maybeGetFileChanges`. The latest version will return the frozen version so we know when to stop.

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
zhu-tom: Yes - we will not see 4 batches with trigger once

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: maybe walk me through the work flow? 
Like Spark Streaming calls A, and A calls delta sharing B->C->D, then the response to A is XXX, then Spark Streaming stops..

spark/src/test/scala/io/delta/sharing/spark/DeltaSharingSourceSuite.scala
linzhou-db: sg.

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
linzhou-db: kk.

linzhou-db (approved):

littlegrasscao (approved):

client/src/main/scala/io/delta/sharing/spark/DeltaSharingSource.scala
zhu-tom: Good rundown from Isaac: 

**Setup**: `prepareForTriggerAvailableNow` → RPC → freeze version 50. 

**Each micro-batch**: `latestOffset` → `getNextOffsetFromPreviousOffset` → `maybeGetFileChanges` → `getOrUpdateLatestTableVersion` returns frozen 50 → RPC chunk up to `min(50, from + maxVersionsPerRpc)` → fills `sortedFetchedFiles` → rate-limit walk → `getBatch` → `createDataFrameFromOffset` [:749] sets `previousOffset = endOffset` [:760], Spark commits to checkpoint, returns DataFrame. 

**Termination**: `sortedFetchedFiles` exhausted → `buildOffsetFromIndexedFile` [:925] bumps `previousOffset` to `(51, -1)` → next `latestOffset` [:998]: `fromVersion(51) > frozenVersion(50)` → nothing fetched → `getNextOffsetFromPreviousOffset` [:886] returns `(51, -1)` as **available** — same as **committed** `(51, -1)` from checkpoint → `committed == available` → `TERMINATED`

linzhou-db: i see. thanks!

Putaek84 (commented):
Putaek84

PR #913, opened 2026-04-10
Migrate Python CI to uv with locked dependencies and pyproject.toml

linzhou-db:
- Replace `setup.py` with `pyproject.toml` and conda CI with `uv sync`
  - `uv.lock` pins all dependencies (including transitive) with integrity hashes                                                                                                 
  - `--exclude-newer` prevents pulling packages published after a known-good date
  - Update `lint-python` to scope checks to `delta_sharing/` and remove `setuptools.extern` dependency                                                                           
  - Update release scripts to use `python -m build`                                                   
                                                                                                                                                                                 
  ## Local test steps                                                                                                                                                            
                                                                                                                                                                                 
  ```bash                                                                                                                                                                        
  cd python                              
  rm -rf .venv
  uv sync     
  uv run maturin develop --manifest-path delta-kernel-rust-sharing-wrapper/Cargo.toml
  uv run dev/lint-python                                                             
  uv run dev/pytest
  ```                                                                                                                                                                                 

  Test plan
                                                                                                                                                                                 
  - uv sync + uv run dev/lint-python + uv run dev/pytest pass locally


python/pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+    "numpy<2.0",
+    "pandas==2.0.3",
+    "pyarrow==16.1.0",
+    "pytest<=9.0.2",
+    "requests<=2.32.5",
+    "types-requests<=2.32.0.20260311",
+    "yarl>=1.6.0,<=1.23.0",
+    "setuptools>=64",
+]
+
```
PatrickJin-db: let's also add a [dependency cooldown](https://docs.astral.sh/uv/concepts/resolution/#dependency-cooldowns) and regenerate the lockfile

python/uv.lock
```diff
@@ -0,0 +1,4539 @@
...
+    { url = "https://files.pythonhosted.org/packages/79/73/fa6d1a96ab7fd6e6d1c3500700963eab46813847f01ef0ccbaa726181dd5/frozenlist-1.5.0-cp312-cp312-macosx_10_13_universal2.whl", hash = "sha256:31115ba75889723431aa9a4e77d5f398f5cf976eea3bdf61749731f62d4a4a21", size = 94026, upload-time = "2024-10-23T09:46:58.601Z" },
+    { url = "https://files.pythonhosted.org/packages/ab/04/ea8bf62c8868b8eada363f20ff1b647cf2e93377a7b284d36062d21d81d1/frozenlist-1.5.0-cp312-cp312-macosx_10_13_x86_64.whl", hash = "sha256:7437601c4d89d070eac8323f121fcf25f88674627505334654fd027b091db09d", size = 54150, upload-time = "2024-10-23T09:46:59.608Z" },
+    { url = "https://files.pythonhosted.org/packages/d0/9a/8e479b482a6f2070b26bda572c5e6889bb3ba48977e81beea35b5ae13ece/frozenlist-1.5.0-cp312-cp312-macosx_11_0_arm64.whl", hash = "sha256:7948140d9f8ece1745be806f2bfdf390127cf1a763b925c4a805c603df5e697e", size = 51927, upload-time = "2024-10-23T09:47:00.625Z" },
+    { url = "https://files.pytho
```
PatrickJin-db, line 2618: some of the dependencies explicitly link to https://pypi-proxy.dev.databricks.com, are we sure we want that? i don't think that domain is accessible outside of our VPN.

dev/release.sh
```diff
@@ -33,7 +33,7 @@ git checkout v$VERSION
 
 # Generate Python artifacts
 cd python/
-python3 setup.py sdist bdist_wheel
```
PatrickJin-db, line 36: shouldn't we be using `uv build` now?

dev/python_release.sh
```diff
@@ -25,7 +25,7 @@ git tag "py-v$VERSION"
 
 # Generate Python artifacts
 pushd python
-python3 setup.py sdist bdist_wheel
```
PatrickJin-db, line 28: same thing here

python/pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+    "aiohttp<=3.13.3",
+    "black==24.3.0",
+    "build>=1.0,<2.0",
+    "flake8<=7.3.0",
+    "fsspec>=0.7.4,<=2026.2.0",
+    "jwcrypto>=1.5.6,<=1.5.6",
+    "maturin<=1.12.6",
+    "mypy==0.981",
+    "numpy<2.0",
+    "pandas==2.0.3",
```
PatrickJin-db: uv fails to install this if you are on python 3.12+. if we want to keep such old dependencies, let's add a `.python-version` file via `uv python pin 3.11`. this file is dev-only and won't affect the python versions required to install this package

python/pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+Issues = "https://github.com/delta-io/delta-sharing/issues"
+                                                                                                                                                                                 
+[tool.setuptools.dynamic]                                 
+version = {attr = "delta_sharing.version.__version__"}
+                                                                                                                                                                                 
+[tool.setuptools.packages.find]
+include = ["delta_sharing"]
+
+[dependency-groups]
+dev = [
```
PatrickJin-db: more generally, do we want to pin our dev dependencies to such old versions? some of these are >2 years old (like `pandas==2.0.3`, `numpy<2.0`). Our github workflow pinned some pretty old versions but I don't think there is any good reason to keep on testing with such old versions.

dev/release.sh
```diff
@@ -12,7 +12,7 @@ git clean -fdx
 # Clean existing artifacts
 build/sbt clean
 cd python
-python3 setup.py clean --all
+rm -rf build dist
```
PatrickJin-db: nit: combine this with the line below

python/pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+adl = ["adlfs"]
+gcs = ["gcsfs"]
+gs = ["gcsfs"]
+
+[project.urls]
+Source = "https://github.com/delta-io/delta-sharing"
+Documentation = "https://github.com/delta-io/delta-sharing"                                                                                                                    
+Issues = "https://github.com/delta-io/delta-sharing/issues"
+                                                                                                                                                                                 
+[tool.setuptools.dynamic]
```
PatrickJin-db: nit: why do some lines in this file have so many trailing whitespaces?

littlegrasscao (dismissed):

python/pyproject.toml
linzhou-db: added

python/uv.lock
linzhou-db: updated

dev/release.sh
linzhou-db: updated.

dev/python_release.sh
linzhou-db: updated.

python/pyproject.toml
linzhou-db: added

python/pyproject.toml
linzhou-db: This is not addressed, will check separately, but wonder any suggestion?

dev/release.sh
linzhou-db: updated

python/pyproject.toml
linzhou-db: from copy paste of isaac answers, removed.

python/uv.lock
PatrickJin-db: hmm i still see them. I did a bit of digging into `uv lock` and I don't think there is an easy way to manage the registry used. I guess the only solution is to keep the proxy URLs and have outside contributors regenerate the uv.lock file. We should add that somewhere in the README.

python/pyproject.toml
PatrickJin-db: I think we should just reuse the dependencies in `project` when possible. Our tests should run against the dependencies used when actually installing `delta_sharing`.

We can add more dependency groups if we later identify any changes that are backwards incompatible with older versions.

linzhou-db: updated

python/uv.lock
linzhou-db: added to `CONTRIBUTING.md`

python/pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+    "Operating System :: OS Independent",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+]
+dependencies = [
+    "delta-kernel-rust-sharing-wrapper>=0.3.1,<1.0",
+    "pandas>=2.0.3,<=3.0.1",
```
PatrickJin-db: looks like our tests fail with pandas 3.0.0+ (will investigate this separately). let's change the upper bound to <3.0.0.

PatrickJin-db (dismissed):

python/pyproject.toml
linzhou-db: updated

PatrickJin-db (approved):

PR #919, opened 2026-04-14
Deprecate Python 3.8 and 3.9 support

PatrickJin-db:
These python versions are EoL and are preventing us from updating many old dependencies, several of which contain CVEs.

There should be no backwards incompatible changes since older relases will continue to work with 3.8/3.9.

This also resolves [CVE-2026-21441](https://github.com/urllib3/urllib3/security/advisories/GHSA-38jv-5279-wg99) in `urllib3` by dropping 1.26.20, which was only used for python <3.10.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -8,7 +8,7 @@ dynamic = ["version"]
 description = "Python Connector for Delta Sharing"
 readme = "README.md"
 license = {text = "Apache-2.0"}
-requires-python = ">=3.8"
+requires-python = ">=3.10"
 authors = [
     {name = "The Delta Lake Project Authors", email = "delta-users@googlegroups.com"},
 ]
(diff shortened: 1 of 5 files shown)


python/pyproject.toml
```diff
@@ -30,8 +28,7 @@ dependencies = [
     "fsspec>=0.7.4,<=2026.2.0",
     "requests>=2.20.0,<=2.32.5",
     "aiohttp>=3.0.0,<=3.13.3",
-    "dataclasses;python_version<'3.8'",
-    "yarl>=1.6.0,<=1.23.0",
+"yarl>=1.6.0,<=1.23.0",
```
littlegrasscao: format?

littlegrasscao (dismissed):

python/pyproject.toml
linzhou-db: @PatrickJin-db

linzhou-db (dismissed):
Q: should we do this separately for other python-branch?

python/pyproject.toml
PatrickJin-db: fixed

linzhou-db (approved):

PR #924, opened 2026-04-17
Fix python build-and-test jobs

PatrickJin-db:
Changelog:
* Get rid of the setup-python action because setup-uv already also allows setting python version
* Have the build-kernel-wheels job use uv too
* Use --manifest-path in `maturin` commands rather than cd-ing into `delta-kernel-rust-sharing-wrapper`. This ensures that uv uses the global `pyproject.toml` instead of `delta-kernel-rust-sharing-wrapper/pyproject.toml`.

### DIFF
--- a/.github/workflows/build-kernel-wheels.yml
+++ b/.github/workflows/build-kernel-wheels.yml
@@ -34,56 +34,58 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
 
-      - name: Set up Python
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
+      - name: Set up Python and uv
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57  # v8.0.0
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-kernel-wheels.yml
```diff
@@ -34,56 +34,56 @@ jobs:
...
         shell: bash
 
       - name: Build wheel (ARM macOS)
         if: matrix.os == 'macos-latest' && matrix.arch == 'arm64'
         run: |
           rustup target add aarch64-apple-darwin
-          cd python/delta-kernel-rust-sharing-wrapper
-          maturin build --release --target aarch64-apple-darwin
+          cd python
+j         maturin build --manifest-path delta-kernel-rust-sharing-wrapper/Cargo.toml --release --target aarch64-apple-darwin
```
littlegrasscao: typo j?

python/.python-version
```diff
@@ -1 +1 @@
-3.11
+3.8
```
littlegrasscao: isn't 3.8 going away?

PatrickJin-db: changed for testing, forgot to remove

littlegrasscao (approved):

linzhou-db (approved):

PR #926, opened 2026-04-17
Align client fileidhash with parquet/delta header values (#910)

littlegrasscao:
Validate and document fileidhash as parquet or delta (replacing md5/sha256). Update RestClient tests. Server must accept these values (merge server PR first).


linzhou-db:
adding @chakankardb to review in case he has different thought on the naming of the key/value of the header.

linzhou-db (approved):

PR #944, opened 2026-05-01
Fix broken async query test

andyl-db:
Previously GetQueryStatus with query id 1234 would just load instantly. However in my last PR that added more async query tests, it made it so it required at least four polls before it would return the table.

This PR fixes broken async query test by polling until the lines exceeds 1 line (table has loaded)


PatrickJin-db (approved):
What about the failing spark tests?

PR #947, opened 2026-05-07
Upgrade delta-kernel-rs to 0.22.0

PatrickJin-db:
Title. The most important fix in this upgrade is to properly handle DB presigned URLs when using the Delta format.

One interface change is required, which is using `DefaultEngineBuilder` to initialize the default engine.

Some Cargo.lock dependencies are updated due to transitive dependency changes in delta-kernel-rs. In particular,
* `rustls` path now uses `reqwest/rustls` instead of `reqwest/rustls-tls-native-roots` + `reqwest/http2`
* New dependencies on `percent-encoding`, `rand`, `tracing-subscriber`.
* Removal of `comfy-table` 

Did not see any performance regressions when testing locally on some shared tables.

### DIFF
--- a/python/delta-kernel-rust-sharing-wrapper/Cargo.lock
+++ b/python/delta-kernel-rust-sharing-wrapper/Cargo.lock
@@ -143,7 +143,6 @@ dependencies = [
  "atoi",
  "base64",
  "chrono",
- "comfy-table",
  "half",
  "lexical-core",
  "num-traits",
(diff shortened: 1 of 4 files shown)


python/delta-kernel-rust-sharing-wrapper/Cargo.lock
```diff
@@ -326,6 +325,28 @@ version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
+[[package]]
+name = "aws-lc-rs"
```
linzhou-db, line 329: why do we need these new dependency changes?

linzhou-db: let's add the reason in the PR description.

linzhou-db (commented):
And how do we test the new change (support for dburl)?

python/delta-kernel-rust-sharing-wrapper/Cargo.lock
PatrickJin-db: added. tl;dr they're changes in transitive dependencies of delta-kernel-rs

linzhou-db (approved):
Still have the question of how to test the support of dburl

PatrickJin-db:
I locally tested reading a DBURL table too. Though it's true that adding an integration test for it is nontrivial.

PR #862, opened 2026-03-11
[Python] Add table handles and snapshot object model

zacdav-db:
Summary:
Add the core object model for Python snapshot reads.

This PR introduces SharingClient.table, DeltaSharingTable, TableSnapshot, table.snapshot, and table.to_pandas as a pure full-snapshot materializer.

Scope:
- table handles
- snapshot configuration
- pandas materialization
- legacy parity coverage for load_as_pandas vs table.snapshot(...).to_pandas
- minimal pandas quickstart and README updates

Not included:
- Arrow APIs
- lazy batch readers
- object-based CDF APIs
- DuckDB examples

Testing:
- py_compile on touched Python files
- focused pytest subset for table creation, snapshot wiring, parity, and direct real-table snapshot reads

Part of #860.

Issue #860, opened 2026-03-08
Python connector: add first-class table handle and Arrow-native read APIs


zacdav-db:
## Summary

The Python connector currently makes users construct `"<profile>#<share>.<schema>.<table>"` strings for common read paths. That works, but it is awkward, easy to get wrong, and makes object-oriented workflows difficult.

At the same time, Arrow-native consumption is underexposed. This makes integration with Arrow consumers such as DuckDB less direct than it should be, and it forces users toward eager materialization even when they want batch-oriented reads.

## Motivation

Today, a representative Python workflow looks like this:

```python
profile_file = "recipient.share"
table_url = profile_file + "#share.schema.table"
data = delta_sharing.load_as_pandas(table_url, limit=10)
```

Pain points:

- Users must manually build and parse `table_url` strings.
- The API shape does not reflect the underlying concepts already present in the connector (`SharingClient`, `Table`, scans).
- Arrow-native use cases are not first-class.
- Lazy batch-oriented consumption for engines like DuckDB is not easy to discover.

## Proposal

Add an additive object-based API alongside the existing URL-based API.

### New surface

```python
client = delta_sharing.SharingClient("recipient.share")
table = client.table("share.schema.table")

pdf = table.to_pandas(limit=10)
arrow_table = table.to_arrow(limit=10)
batches = table.to_record_batches(limit=10)
reader = table.to_record_batch_reader(limit=10)
```

Also add a URL-based Arrow helper for parity:

```python
arrow_table = delta_sharing.load_as_arrow("recipient.share#share.schema.table", limit=10)
```

### Design goals

- Keep the existing URL-based APIs working unchanged.
- Make the new API additive, not a replacement.
- Support both eager Arrow materialization and lazy Arrow batch consumption.
- Make it easy for engines like DuckDB to consume a `RecordBatchReader` directly.

## Compatibility requirements

This should not disrupt existing users.

- `load_as_pandas(...)` remains supported.
- The `"<profile>#<share>.<schema>.<table>"` format remains supported.
- New examples should demonstrate the object-based API.
- Existing syntax should remain documented as a compatibility path.

## Implementation notes

The implementation should ideally make pandas an adapter over shared reader logic rather than the only primary surface.

In particular:

- `SharingClient.table("share.schema.table")` should return a first-class table handle.
- `to_arrow()`, `to_record_batches()`, and `to_record_batch_reader()` should share a common Arrow read path.
[...]

PR #949, opened 2026-05-19
[Python] Add Arrow materializers for table snapshots

zacdav-db:
**Summary**
Adds Arrow-native materializers to Python table snapshots, following the staged object API plan in #860 and building on the table-handle APIs from #862.

This PR adds snapshot support through:

- `to_arrow()`
- `to_record_batches()`
- `to_record_batch_reader()`

It also keeps the URL-based `load_as_arrow(...)` helper available for parity with the existing legacy read APIs.

**Scope**
- snapshot Arrow materialization
- lazy `RecordBatch` iteration and `RecordBatchReader` support
- Delta `timestamp` columns normalized to `timestamp[us, UTC]` so parquet and delta-kernel paths return consistent schemas
- focused coverage for object API wiring, legacy parity, schema normalization, partition columns, timestamp normalization, and kernel-path limits

CDF Arrow materialization is intentionally split into a follow-up PR.

**Not Included**
- CDF Arrow materialization
- table-handle interface changes already merged in #862
- docs and examples
- DuckDB quickstart
- dependency changes
- changes to existing pandas or Spark API behavior

**Testing**
- Python test suite: `87 passed, 126 skipped`
- `dev/lint-python` (compile, black, pycodestyle, flake8, mypy): passed
- pydoc checks skipped because `sphinx-build` is not installed
- `git diff --check upstream/main`: passed

Part of #860; builds on #862.

### DIFF
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -14,10 +14,11 @@
 # limitations under the License.
 #
 from itertools import chain
-from typing import BinaryIO, List, Optional, Sequence, TextIO, Tuple, Union
+from typing import BinaryIO, Iterator, List, Optional, Sequence, TextIO, Tuple, Union
 from pathlib import Path
 
 import pandas as pd
+import pyarrow as pa
 
 from delta_sharing.protocol import CdfOptions, Protocol, Metadata
 
@@ -155,6 +156,43 @@ def load_as_pandas(
     ).to_pandas()
 
 
+def load_as_arrow(
+    url: str,
+    limit: Optional[int] = None,
+    version: Optional[int] = None,
+    timestamp: Optional[str] = None,
+    jsonPredicateHints: Optional[str] = None,
+    use_delta_format: Optional[bool] = None,
+) -> pa.Table:
+    """
+    Load the shared table using the given url as a PyArrow Table.
+
+    :param url: a url under the format "<profile>#<share>.<schema>.<table>"
+    :param limit: a non-negative int. Load only the ``limit`` rows if the parameter is specified.
+      Use this optional parameter to explore the shared table without loading the entire table to
+      the memory.
+    :param version: an optional non-negative int. Load the snapshot of table at version
+    :param timestamp: an optional string. Load the snapshot of table at version corresponding
+      to the timestamp.
+    :param jsonPredicateHints: Predicate hints to be applied to the table. For more details see:
+      https://github.com/delta-io/delta-sharing/blob/main/PROTOCOL.md#json-predicates-for-filtering
+    :param use_delta_format: Whether to use delta format or parquet format. If not set, table
+      metadata will be used to determine whether to use delta or parquet format.
+    :return: A PyArrow Table representing the shared table.
+    """
+    profile_json, share, schema, table = _parse_url(url)
+    profile = DeltaSharingProfile.read_from_file(profile_json)
+    return DeltaSharingReader(
+        table=Table(name=table, share=share, schema=schema),
+        rest_client=DataSharingRestClient(profile),
+        jsonPredicateHints=jsonPredicateHints,
+        limit=limit,
+        version=version,
+        timestamp=timestamp,
+        use_delta_format=use_delta_format,
+    ).to_arrow()
+
+
 class TableSnapshot:
     def __init__(
         self,
@@ -189,6 +227,15 @@ class TableSnapshot:
 
     def to_pandas(self, convert_in_batches: bool = False) -> pd.DataFrame:
         return self._reader(convert_in_batches=convert_in_batches).to_pandas()
+
+    def to_arrow(self) -> pa.Table:
+        return self._reader().to_arrow()
+
+    def to_record_batches(self) -> Iterator[pa.RecordBatch]:
+        return self._reader().to_record_batches()
+
+    def to_record_batch_reader(self) -> pa.RecordBatchReader:
+        return self._reader().to_record_batch_reader()
 
     def _table_name(self) -> str:
         return f"{self._table.share}.{self._table.schema}.{self._table.name}"
@@ -341,6 +388,15 @@ class DeltaSharingTable:
 
     def to_pandas(self, convert_in_batches: bool = False) -> pd.DataFrame:
         return self.snapshot().to_pandas(convert_in_batches=convert_in_batches)
+
+    def to_arrow(self) -> pa.Table:
+        return self.snapshot().to_arrow()
+
+    def to_record_batches(self) -> Iterator[pa.RecordBatch]:
+        return self.snapshot().to_record_batches()
+
+    def to_record_batch_reader(self) -> pa.RecordBatchReader:
+        return self.snapshot().to_record_batch_reader()
 
     def to_spark(self) -> "PySparkDataFrame":  # noqa: F821
         return self.snapshot().to_spark()
--- a/python/delta_sharing/delta_sharing.py
+++ b/python/delta_sharing/delta_sharing.py
@@ -170,7 +170,7 @@ def load_as_arrow(
     :param url: a url under the format "<profile>#<share>.<schema>.<table>"
     :param limit: a non-negative int. Load only the ``limit`` rows if the parameter is specified.
       Use this optional parameter to explore the shared table without loading the entire table to
-      the memory.
+      memory.
     :param version: an optional non-negative int. Load the snapshot of table at version
     :param timestamp: an optional string. Load the snapshot of table at version corresponding
       to the timestamp.
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -286,6 +286,8 @@ class DeltaSharingReader:
             left = self._limit
             for index, file in enumerate(response.add_files):
                 file_limit = left
+                # Reuse the first dataset opened above to resolve column casing;
+                # each remaining file needs its own dataset for batch scanning.
                 pa_dataset = (
                     first_dataset if index == 0 else DeltaSharingReader._parquet_dataset(file.url)
                 )
--- a/python/delta_sharing/reader.py
+++ b/python/delta_sharing/reader.py
@@ -266,7 +266,6 @@ class DeltaSharingReader:
 
         response, schema_json = self._list_files()
         schema = to_arrow_schema(schema_json)
-        converters = to_converters(schema_json)
 
         if len(response.add_files) == 0 or self._limit == 0:
             return schema, iter(())
@@ -275,9 +274,7 @@ class DeltaSharingReader:
             left = self._limit
             for file in response.add_files:
                 file_limit = left
-                for batch in DeltaSharingReader._to_record_batches(
-                    file, schema_json, converters, file_limit
-                ):
+                for batch in DeltaSharingReader._to_record_batches(file, schema_json, file_limit):
                     yield batch
                     if left is not None:
                         left -= batch.num_rows
@@ -597,7 +594,6 @@ class DeltaSharingReader:
     def _to_record_batches(
         action: FileAction,
         schema_json: dict,
-        converters: Dict[str, Callable[[str], Any]],
         limit: Optional[int],
     ) -> Iterator[pa.RecordBatch]:
         filesystem = DeltaSharingReader._parquet_filesystem(action.url)
@@ -612,7 +608,7 @@ class DeltaSharingReader:
             if limit is not None and rows_read + batch.num_rows > limit:
                 batch = batch.slice(0, limit - rows_read)
 
-            yield DeltaSharingReader._normalize_record_batch(batch, action, schema_json, converters)
+            yield DeltaSharingReader._normalize_record_batch(batch, action, schema_json)
             rows_read += batch.num_rows
 
     @staticmethod
@@ -635,7 +631,6 @@ class DeltaSharingReader:
         batch: pa.RecordBatch,
         action: FileAction,
         schema_json: dict,
-        converters: Dict[str, Callable[[str], Any]],
     ) -> pa.RecordBatch:
         columns = []
         names = []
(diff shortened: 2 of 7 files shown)


zacdav-db:
@PatrickJin-db this follows on from #862 (once merged)

python/delta_sharing/reader.py
```diff
@@ -148,6 +148,47 @@ def __to_pandas_kernel(self):
...
+                    timestamp=self._timestamp,
+                )
+            finally:
+                self._rest_client.remove_delta_format_header()
+            table_path = self.__write_temp_delta_log_snapshot(temp_dir.name, response.lines)
+            interface = delta_kernel_rust_sharing_wrapper.PythonInterface(table_path)
+            table = delta_kernel_rust_sharing_wrapper.Table(table_path)
+            snapshot = table.snapshot(interface)
+            scan = delta_kernel_rust_sharing_wrapper.ScanBuilder(snapshot).build()
+            scan_result = scan.execute(interface)
```
PatrickJin-db: can we extract this into a helper function? this is basically all copy pasted from `__to_pandas_kernel`

python/delta_sharing/reader.py
```diff
@@ -387,6 +495,111 @@ def __table_changes_to_pandas_kernel(self, cdfOptions: CdfOptions) -> pd.DataFra
...
+            os.makedirs(log_dir)
+            self.__write_temp_delta_log_cdf(
+                log_dir,
+                delta_protocol,
+                min_version,
+                max_version,
+                version_to_metadata,
+                version_to_actions,
+                version_to_timestamp,
+            )
```
PatrickJin-db: similar comment here about the code duplication with `__table_changes_to_pandas_kernel`

python/delta_sharing/reader.py
```diff
@@ -514,6 +747,156 @@ def _to_pandas(
...
+                    raise ValueError("Cannot partition on binary or complex columns")
+                value = converter(action.partition_values[field_name])
+                columns.append(pa.array([value] * num_rows, type=field_type))
+            else:
+                columns.append(pa.nulls(num_rows, type=field_type))
+
+        return pa.RecordBatch.from_arrays(columns, names=names)
+
+    @staticmethod
+    def _to_arrow_type(schema_type):
```
PatrickJin-db: would prefer if this can be moved to `delta_sharing/converter.py`

python/delta_sharing/reader.py
```diff
@@ -530,8 +913,8 @@ def _commit_version_col_name():
 
     @staticmethod
     def _add_special_cdf_schema(schema_json: dict) -> dict:
```
PatrickJin-db, line 686: just want to double check, the changes to this function are not supposed to change its behavior right?

python/delta_sharing/reader.py
```diff
@@ -213,6 +254,73 @@ def to_pandas(self) -> pd.DataFrame:
...
+            return self.__record_batches_kernel()
+
+        response = self._rest_client.list_files_in_table(
+            self._table,
+            predicateHints=self._predicateHints,
+            jsonPredicateHints=self._jsonPredicateHints,
+            limitHint=self._limit,
+            version=self._version,
+            timestamp=self._timestamp,
+        )
```
PatrickJin-db: would also prefer if this bit is extracted to a helper method, but my opinion here is not as strong as in other places

python/delta_sharing/reader.py
```diff
@@ -514,6 +747,156 @@ def _to_pandas(
...
+        url = urlparse(action.url)
+        if "storage.googleapis.com" in (url.netloc.lower()):
+            import delta_sharing._yarl_patch  # noqa: F401
+
+        protocol = url.scheme
+        proxy = getproxies()
+        if len(proxy) != 0:
+            filesystem = fsspec.filesystem(protocol, client_kwargs={"trust_env": True})
+        else:
+            filesystem = fsspec.filesystem(protocol)
```
PatrickJin-db: this part is duplicated with _to_pandas, can we make a helper function for it?

python/delta_sharing/reader.py
```diff
@@ -514,6 +747,156 @@ def _to_pandas(
...
+    def _to_arrow_schema(schema_json: dict) -> pa.Schema:
+        return pa.schema(
+            [
+                pa.field(field["name"], DeltaSharingReader._to_arrow_type(field["type"]))
+                for field in schema_json["fields"]
+            ]
+        )
+
+    @staticmethod
+    def _normalize_record_batch(
```
PatrickJin-db, line 630: i don't think this method is preserving the column order in schema_json if there are partition columns. you would need to either add logic here to create the columns in the right order, or make sure all callers use the right column order afterwards.

zacdav-db: Done, `_normalize_record_batch` builds columns in `schema_json` order, and I added a test covering a partition column in the middle of the schema.

python/delta_sharing/reader.py
zacdav-db: Extracted the shared snapshot delta-kernel scan setup into `__snapshot_scan_kernel` and reused it from both pandas and Arrow batch paths.

python/delta_sharing/reader.py
zacdav-db: Extracted the shared CDF delta-kernel scan setup into `__table_changes_scan_kernel`

python/delta_sharing/reader.py
zacdav-db: The returned schema is unchanged, but we avoid mutating the input. Added a focused test for this.

zacdav-db: As far as I can tell the mutating behaviour was not relied upon currently and this is more predictable.

python/delta_sharing/delta_sharing.py
```diff
@@ -155,6 +156,43 @@ def load_as_pandas(
     ).to_pandas()
 
 
+def load_as_arrow(
```
PatrickJin-db, line 159: are we adding a `load_table_changes_as_arrow` function or will that be in a different PR?

python/delta_sharing/reader.py
```diff
@@ -213,6 +243,78 @@ def to_pandas(self) -> pd.DataFrame:
...
+                            raise RuntimeError(
+                                "'_to_record_batches' returned more rows than the "
+                                f"requested limit of {file_limit}"
+                            )
+                        if left == 0:
+                            return
+
+        return schema, iterator()
+
+    def _list_files_for_arrow(self):
```
PatrickJin-db: seems like the only part of this function that is related to arrow is the `to_arrow_schema` call. in that case, i feel like extracting the just the `self._rest_client.list_files_in_table(...)` and `schema_json = ...` parts is a better idea so that it can be used in the pandas codepaths too.

python/delta_sharing/reader.py
```diff
@@ -514,6 +679,106 @@ def _to_pandas(
...
+                        if action.timestamp is not None
+                        else pa.nulls(num_rows, type=field_type)
+                    )
+                    continue
+
+            if field_name in action.partition_values:
+                converter = converters[field_name]
+                if converter is None:
+                    raise ValueError("Cannot partition on binary or complex columns")
+                value = converter(action.partition_values[field_name])
```
PatrickJin-db: hmm, the `converters` passed in here is made for converting columns values from Delta types to pandas types, are you sure we want to use that here? why doesn't `columns.append(pa.array([value] * num_rows, type=field_type))` on the original value work?

python/delta_sharing/tests/test_reader.py
```diff
@@ -67,6 +145,318 @@ def remove_delta_format_header(self):
...
+        RestClientMock(),
+        limit=2,
+        use_delta_format=True,
+    )
+
+    result = pa.Table.from_batches(list(reader.to_record_batches()))
+    assert result.column("a").to_pylist() == [1, 2]
+
+
+def test_to_arrow_non_partitioned(tmp_path):
```
PatrickJin-db: instead of creating a bunch of arrow tests, how about we take the existing tests and modify them to test reading them as both arrow and pandas? there are methods to convert to/from pandas and arrow, which hopefully make it easier to verify that the arrows and pandas results are the same.

PatrickJin-db (commented):
Can we split the snapshot and CDF methods into separate PRs? This PR is a bit large.

python/delta_sharing/delta_sharing.py
zacdav-db: Will ensure thats part of the split PR for CDF.

python/delta_sharing/reader.py
zacdav-db: Good call. That is better.

python/delta_sharing/tests/test_delta_sharing.py
PatrickJin-db: nit: the test name says `to_pandas` but in actuality it's testing `to_arrow` as well now

python/delta_sharing/tests/test_reader.py
PatrickJin-db, line 1: can we adapt all of the existing snapshot tests to test both arrow and python? (and rename the test names accordingly)

python/delta_sharing/tests/test_delta_sharing.py
PatrickJin-db, line 1: similarly, we can compare the pandas and arrow results in the integration tests here. i can run them locally once they're added.

PatrickJin-db (commented):
Logic seems fine now but I would strongly prefer more test coverage. We also need to split up reader.py but that can be done after all the arrow PRs are merged.

zacdav-db:
@PatrickJin-db Reused the existing snapshot coverage for pandas/Arrow parity, including parquet and kernel paths.

python/delta_sharing/delta_sharing.py
```diff
@@ -155,6 +156,43 @@ def load_as_pandas(
...
+    jsonPredicateHints: Optional[str] = None,
+    use_delta_format: Optional[bool] = None,
+) -> pa.Table:
+    """
+    Load the shared table using the given url as a PyArrow Table.
+
+    :param url: a url under the format "<profile>#<share>.<schema>.<table>"
+    :param limit: a non-negative int. Load only the ``limit`` rows if the parameter is specified.
+      Use this optional parameter to explore the shared table without loading the entire table to
+      the memory.
```
PatrickJin-db: nit: "to the memory" -> "to memory"

python/delta_sharing/tests/test_delta_sharing.py
```diff
@@ -59,6 +61,38 @@
...
+    pdf: pd.DataFrame,
+    arrow_table: pa.Table,
+    *,
+    sort_by: Optional[Sequence[str]] = None,
+) -> None:
+    # Arrow follows logical Delta dtypes and uses UTC-aware Delta timestamps, while
+    # the legacy pandas path may preserve parquet dtypes and naive UTC timestamps.
+    arrow_pdf = arrow_table.to_pandas()
+    if sort_by is not None:
+        pdf = pdf.sort_values(by=sort_by).reset_index(drop=True)
```
PatrickJin-db, line 87: since we're already sorting the both pdfs, can we also get rid of unnecessary sorts that the test cases perform before calling `_assert_arrow_matches_pandas`

python/delta_sharing/tests/test_reader.py
```diff
@@ -30,8 +39,51 @@
...
+        return self._response
+
+
+def _assert_snapshot_materializers(
+    reader: DeltaSharingReader,
+    expected: pd.DataFrame,
+    *,
+    expected_arrow_pdf: Optional[pd.DataFrame] = None,
+    check_arrow_dtype: bool = True,
+) -> pa.Table:
```
PatrickJin-db: nit: i think it would make more sense if this returned both the pandas df and arrow table. i know right now only the arrow table is being used but returning both would help check specific properties of the pandas df if needed.

python/delta_sharing/tests/test_delta_sharing.py
PatrickJin-db: just ran them, they all passed

python/delta_sharing/tests/test_reader.py
```diff
@@ -214,15 +435,18 @@ def autoresolve_query_format(self, table: Table):
...
     expected1["B"] = "x"
     expected2 = pdf2.copy()
     expected2["B"] = "y"
     expected = pd.concat([expected1, expected2]).reset_index(drop=True)
 
-    pd.testing.assert_frame_equal(pdf, expected)
+    _assert_snapshot_materializers(
+        reader,
+        expected,
+        expected_arrow_pdf=expected.rename(columns={"a": "A"}),
```
PatrickJin-db, line 448: hmm so the pandas table has the `a` column lowercase but the arrow table has it uppercase? @linzhou-db which behavior is intended?

PatrickJin-db (dismissed):

python/delta_sharing/tests/test_reader.py
PatrickJin-db: would prefer if the existing lowercasing behavior is preserved but it's not a blocker

python/delta_sharing/tests/test_reader.py
zacdav-db: Will make that adjustment now.

zacdav-db:
@PatrickJin-db addressed your comments:

- Made Arrow match pandas column naming: Parquet columns keep their physical casing, while added partition columns keep the Delta schema casing.
- Kept those names consistent across Arrow tables, record batches, and record batch readers, and reused the first Parquet dataset.
- Cleaned up the snapshot test helper and removed duplicate sorting.
- Fixed the `load_as_arrow` wording.

python/delta_sharing/reader.py
```diff
@@ -213,6 +234,95 @@ def to_pandas(self) -> pd.DataFrame:
...
+
+        if len(response.add_files) == 0 or self._limit == 0:
+            return schema, iter(())
+
+        first_dataset = DeltaSharingReader._parquet_dataset(response.add_files[0].url)
+        physical_names = {name.lower(): name for name in first_dataset.schema.names}
+        # Match pandas column naming: preserve physical Parquet casing for data
+        # columns, while columns synthesized from Delta metadata (such as
+        # partition columns) retain their Delta schema casing.
+        schema = pa.schema(
```
PatrickJin-db: Sorry, I actually didn't realize that the pandas path was relying on the casing in the parquet files (I thought there was an explicit conversion somewhere but I misremembered). This fix is quite unwieldy as a result. Now I'd prefer to keep the code simpler and live with different column casing between pandas and arrow (arguably it's the pandas codepath that's buggy here).

zacdav-db: I've reverted this adjustment now.

PatrickJin-db (approved):

PR #962, opened 2026-06-23
Retry transient errors during RandomAccessHttpInputStream stream reads

littlegrasscao:
## Summary

- Failures thrown from `currentStream.read(...)` in `RandomAccessHttpInputStream` (e.g. `SocketTimeoutException`, connection-reset `SocketException`, mid-stream `IOException`) are currently propagated to the caller without retry, even though the read is logically idempotent: `pos` is only advanced after a successful read, so re-issuing the byte-range request from the unchanged `pos` yields the bytes the caller still needs.
- Add an opt-in retry that wraps both `read()` and `read(buf, off, len)` in `RetryUtils.runWithExponentialBackoff` and aborts the broken `currentStream` on failure so the next attempt reopens a fresh HTTP connection at the same `pos`. The exception is propagated to `RetryUtils.shouldRetry`, so only its existing transient set (timeouts, connection resets, retryable HTTP statuses, generic `IOException`) is retried, while `InterruptedException` / `InterruptedIOException` / non-retryable status codes still surface immediately.
- The new behavior is gated by a new conf, `spark.delta.sharing.client.streamReadRetry.enabled` (default `false`), and reuses the existing `spark.delta.sharing.network.numRetries` / `spark.delta.sharing.network.maxRetryDuration` budget.

## Why this is safe to retry

- `pos` advances only when `currentStream.read` returns successfully, so a retry re-issues the same byte range from the byte the caller still needs.
- On exception we explicitly abort the broken stream (`abortCurrentStream()`) so each retry attempt goes through `reopen(pos)` and gets a fresh HTTP connection rather than reusing a half-broken one.
- `RetryUtils.shouldRetry` already filters non-transient exceptions (e.g. `InterruptedException`, `ConnectionClosedException`, non-retryable HTTP statuses).
- `assertNotClosed()` runs on every attempt, so a refreshed pre-signed URL or a stream close racing with the read is still observed.

## Test plan

- [x] `client/testOnly io.delta.sharing.client.RandomAccessHttpInputStreamSuite` &mdash; new tests cover retry-on-success, no-retry-by-default, and the single-byte `read()` path.
- [x] `client/testOnly io.delta.sharing.client.util.ConfUtilsSuite` &mdash; new test covers the conf accessors for both `Configuration` and `SQLConf`.
- [x] `client/testOnly io.delta.sharing.client.util.RetryUtilsSuite` &mdash; existing retry behavior still passes.
- [x] `client/Test/compile` &mdash; scalastyle and compilation clean.
- Other client suite failures observed locally (`DeltaSharingFileSystemSuite::"file system should be cached"`, `DeltaSharing*Suite` integration tests) reproduce on `origin/main` without this change and are unrelated.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
+++ b/client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
@@ -44,7 +44,7 @@ import java.nio.charset.StandardCharsets.UTF_8
 
 import org.apache.commons.io.IOUtils
 import org.apache.hadoop.fs.{FileSystem, FSExceptionMessages, FSInputStream}
-import org.apache.http.HttpStatus
+import org.apache.http.{HttpEntity, HttpStatus}
 import org.apache.http.client.HttpClient
 import org.apache.http.client.methods.{HttpGet, HttpRequestBase}
 import org.apache.http.conn.EofSensorInputStream
@@ -115,7 +115,7 @@ private[sharing] class RandomAccessHttpInputStream(
               s"error: ${e.getMessage}")
           }
           if (retryStreamReadOnError) {
-            // Drop the broken stream so the next retry attempt opens a fresh connection
+            // Drop the broken stream so the retry attempt opens a fresh connection
             // at the still-unchanged `pos`.
             abortCurrentStream()
           }
@@ -153,7 +153,7 @@ private[sharing] class RandomAccessHttpInputStream(
               s"length: $len, error: ${e.getMessage}")
           }
           if (retryStreamReadOnError) {
-            // Drop the broken stream so the next retry attempt opens a fresh connection
+            // Drop the broken stream so the retry attempt opens a fresh connection
             // at the still-unchanged `pos`.
             abortCurrentStream()
           }
@@ -170,16 +170,18 @@ private[sharing] class RandomAccessHttpInputStream(
   }
 
   /**
-   * Wraps a single underlying-stream read in `RetryUtils.runWithExponentialBackoff` when
-   * `retryStreamReadOnError` is enabled. The read is logically idempotent: `pos` is only
-   * advanced on a successful read, so re-issuing the byte-range request from the unchanged
-   * `pos` yields the bytes the caller still needs. The caller is responsible for aborting the
-   * broken `currentStream` (so that the next attempt reopens a fresh connection at `pos`)
-   * before letting the exception propagate.
+   * When `retryStreamReadOnError` is enabled, retries a failed stream read. The read is
+   * idempotent because `pos` only advances on success, and the caller aborts the broken
+   * stream so the retry reopens a fresh connection at `pos`. The short `retrySleepInterval`
+   * keeps a concurrent `close()` from blocking on the monitor during the retry sleep.
    */
   private def doStreamRead(readOp: => Int): Int = {
     if (retryStreamReadOnError) {
-      RetryUtils.runWithExponentialBackoff(numRetries, maxRetryDuration) {
+      RetryUtils.runWithExponentialBackoff(
+        numRetries,
+        maxRetryDuration,
+        retrySleepInterval = 500L
+      ) {
         readOp
       }
     } else {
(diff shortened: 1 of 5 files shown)


linzhou-db (dismissed):
Looks good! 
Especially aborting the current stream on exception, which avoid duplicate read. 

Why are we making this change?

client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
```diff
@@ -100,19 +101,26 @@ private[sharing] class RandomAccessHttpInputStream(
...
+            logInfo(s"Error reading from stream - uri: $uri, position: $pos, " +
+              s"error: ${e.getMessage}")
+          }
+          if (retryStreamReadOnError) {
+            // Drop the broken stream so the next retry attempt opens a fresh connection
+            // at the still-unchanged `pos`.
+            abortCurrentStream()
+          }
+          throw e
+      }
```
chakankardb, line 123: read() and read(buf, off, len) are both synchronized methods. 
doStreamRead calls RetryUtils.runWithExponentialBackoff, which calls sleeper(sleepMs) (Thread.sleep by default, up to maxRetryDuration = 10 min). 
The thread holds the object monitor for the entire retry sleep.
                                                                                                                    
Consequence: any concurrent close() call (also synchronized) blocks for the entire retry duration.
In Spark, task cancellation typically closes input streams — this could cause a task cancel to hang for minutes. 

The description claims assertNotClosed() "runs on every attempt, so a stream close racing with the read is still observed" — but that's not true since close() can't acquire the lock while the reading thread is sleeping.

client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
```diff
@@ -131,19 +139,26 @@ private[sharing] class RandomAccessHttpInputStream(
...
+            logInfo(s"Error reading from stream - uri: $uri, position: $pos, offset: $off, " +
+              s"length: $len, error: ${e.getMessage}")
+          }
+          if (retryStreamReadOnError) {
+            // Drop the broken stream so the next retry attempt opens a fresh connection
+            // at the still-unchanged `pos`.
+            abortCurrentStream()
+          }
+          throw e
+      }
```
chakankardb, line 161: reopen(pos) is called inside the doStreamRead retry lambda (because currentStream was nulled by abortCurrentStream() on the failed attempt). 
reopen itself calls RetryUtils.runWithExponentialBackoff(numRetries, maxRetryDuration). So when retryStreamReadOnError =  true and the HTTP connection itself is broken:                                                                                                                                 
  - reopen tries numRetries + 1 times → exhausts, throws  
  - doStreamRead's outer retry catches that, tries reopen again numRetries more times
  - Total connection attempts: (numRetries + 1)² — with default numRetries = 3, that's 16 attempts instead of 4

littlegrasscao:
> Looks good! Especially aborting the current stream on exception, which avoid duplicate read.
> 
> Why are we making this change?

There is a ES ticket around this code showing intermittent failures, so I am adding retry here to see if it helps

client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
littlegrasscao: I did:
1. shorten the sleep time to 500ms, 3 retries would have 3.5s wait time in total
2. remove reopen retry if there is an outer retry already.

client/src/main/scala/io/delta/sharing/client/RandomAccessHttpInputStream.scala
littlegrasscao: Giving this a bit more thought, the existing reopen has a retry, the method would be blocked by it already up to 10 mins. 

The new retry block includes assertNotClosed, so on every retry, it would check assertNotClosed. The only wait time is the retry sleep, which I reduced to 500ms and double each time.

linzhou-db (approved):
looks good! though feels a bit awkward, I just confirmed that this new retry covers more cases.

PR #964, opened 2026-06-24
[PROTOCOL] Document version/timestamp query params on Query Table Metadata

littlegrasscao:
## Summary
- The Delta Sharing reference client issues `GET .../tables/{table}/metadata` with optional `version=<v>` or `timestamp=<t>` URL query parameters (see `getEncodedMetadataParams` in `client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala`), but the **Query Table Metadata** section of `PROTOCOL.md` only documented the path segments.
- This PR adds documentation for those optional query parameters, their semantics, and the fact that `version` and `timestamp` are mutually exclusive — mirroring how the corresponding `version` / `timestamp` fields are described on the Read Data from a Table request body.
- No code or wire-format changes; documentation only.

## Test plan
- [x] `git diff` reviewed — only `PROTOCOL.md` changed.
- [ ] Rendered Markdown reviewed on the PR page to confirm the new "Query Parameters" row renders correctly inside the existing HTML `<table>`.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1576,7 +1576,7 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
 <td>URL</td>
 <td>
 
-`{prefix}/shares/{share}/schemas/{schema}/tables/{table}/metadata`
+`{prefix}/shares/{share}/schemas/{schema}/tables/{table}/metadata[?version={version}|&timestamp={timestamp}]`
 
 </td>
 </tr>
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -1576,7 +1576,7 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
 <td>URL</td>
 <td>
 
-`{prefix}/shares/{share}/schemas/{schema}/tables/{table}/metadata`
+`{prefix}/shares/{share}/schemas/{schema}/tables/{table}/metadata[?version={version}|&timestamp={timestamp}]`
```
linzhou-db: nit but to keep consistenc: In other GET endpoints, we are not showcasing the query parameters in this line, shall we consider skipping the change here?

littlegrasscao: Good point — done in 233ad340b. Reverted the URL line back to just the path so it matches the other GET endpoints; kept the new "Query Parameters" row with the `version` / `timestamp` docs.

linzhou-db (approved):

PR #966, opened 2026-06-25
Add includeHistoricalProtocol CDF query parameter (delta-format only)

littlegrasscao:
## Summary
Mirror the existing `includeHistoricalMetadata` plumbing for Protocol actions, so a streaming client can detect when a table's Protocol / reader features change across versions in a CDF (`/changes`) query.

The new flag only affects **`responseformat=delta`** responses. For `responseformat=parquet`, the parquet wire shape has no representation for additional Protocol lines, so the server silently drops the flag and keeps the response shape backwards compatible.

## Wire surface
- **Client** (`DeltaSharingClient.getCDFFiles`) takes a new `includeHistoricalProtocol: Boolean = false` parameter and (when true) emits `includeHistoricalProtocol=true` on the `/changes` URL alongside any `includeHistoricalMetadata=true`.
- **Server** (`DeltaSharingService.listCdfFiles`) accepts the new `includeHistoricalProtocol` query parameter and propagates it through `DeltaSharedTableProtocol.queryCDF` and the standalone implementation into `DeltaSharingCDCReader.queryCDF`. The kernel implementation just carries the new arg (it still throws \`Unsupported\`).
- **Cache key**: the new parameter is included in `QueryParamChecksum` so paginated queries with the flag don't collide with cached page tokens from requests without it.

## Reader semantics
`DeltaSharingCDCReader` emits intermediate `Protocol` actions seen in the delta log under the same `v > start` rule that already gates `Metadata`, in all three CDF branches (CDC, hotfix metadata-only, AddFile/RemoveFile). `DeltaSharedTable.queryCDF` gates the entire feature on the response format being `delta`.

## Backwards compatibility
- The new client parameter has a default of `false`, so existing call sites (`spark/RemoteDeltaCDFRelation`, `RemoteDeltaLogSuite`, `DeltaSharingSource`, the python client, etc.) compile and behave unchanged.
- The new server query param is optional and defaults to `false`.
- For parquet responses the server silently ignores the flag.

## Docs
- `PROTOCOL.md`: the **Read Change Data Feed from a Table** section now documents `includeHistoricalProtocol` and explicitly notes that it only affects `responseformat=delta`.
- `delta-sharing-protocl-api-description.yml`: added the new query parameter to the OpenAPI spec.

## Tests
- Two client unit tests assert the URL the REST client sends contains `includeHistoricalProtocol=true` only when requested, and that it is absent (while `includeHistoricalMetadata=true` is still present) when not.
- One server integration test confirms the new query parameter is accepted on `/changes` and produces an unchanged response shape for a table with no historical Protocol upgrades (no Protocol actions injected, single head Protocol preserved).

## Test plan
- [x] `client/src/main/scala/...` and `server/src/main/scala/...` compile under both scala 2.12 (server) and 2.13 (client/spark).
- [x] New client unit tests pass locally: \`testOnly io.delta.sharing.client.DeltaSharingRestClientSuite -- -z "includeHistoricalProtocol"\` -> 2 succeeded.
- [ ] Existing server integration tests under `streaming_notnull_to_null` need S3 credentials and only run in CI; they are unchanged.
- [ ] CI to run full client/server/spark test suites against the new flag.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -2349,7 +2349,8 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
  **startingTimestamp** (type: String, optional): The starting timestamp of the query, a string in the [Timestamp Format](#timestamp-format), which will be converted to a version created greater or equal to this timestamp. <br>
  **endingVersion** (type: Long, optional): The ending version of the query, inclusive. <br>
  **endingTimestamp** (type: String, optional): The ending timestamp of the query, a string in the [Timestamp Format](#timestamp-format), which will be converted to a version created earlier than or at the timestamp. <br>
- **includeHistoricalMetadata** (type: Boolean, optional): If set to true, return the historical metadata if seen in the delta log. This is for the streaming client to check if the table schema is still read compatible.</td>
+ **includeHistoricalMetadata** (type: Boolean, optional): If set to true, return the historical metadata if seen in the delta log. This is for the streaming client to check if the table schema is still read compatible. <br>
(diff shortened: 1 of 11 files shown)


littlegrasscao:
Split into two smaller PRs for easier review: client wire-only #967 and server + protocol docs #968. Closing this one in favor of those.

PR #967, opened 2026-06-25
[Client] Add includeHistoricalProtocol param on getCDFFiles

littlegrasscao:
## Summary
Client-side half of `includeHistoricalProtocol` support for the CDF endpoint. Adds a new `includeHistoricalProtocol: Boolean = false` parameter on `DeltaSharingClient.getCDFFiles`. When set, the REST client appends `includeHistoricalProtocol=true` to the `/changes` URL alongside any existing `includeHistoricalMetadata=true`, mirroring how `includeHistoricalMetadata` is encoded today.

The matching server change (parse the new query param and emit historical `Protocol` actions in delta-format CDF responses) is split into a companion PR.

## Backwards compatibility
- Default is `false`, so existing call sites (`RemoteDeltaCDFRelation`, `DeltaSharingSource`, `RemoteDeltaLogSuite`, etc.) compile and behave unchanged.
- When the flag is `false`, the URL byte-for-byte matches today's request, so old servers see no difference.
- When the flag is `true` against an old server, the unknown query param is ignored.
- `TestDeltaSharingClient` (a spark-test mock implementing the trait) is updated to match the new signature so the `spark` module still compiles.

## Test plan
- [x] `client` + `spark` modules compile and test-compile cleanly.
- [x] New client unit tests pass: `testOnly io.delta.sharing.client.DeltaSharingRestClientSuite -- -z "includeHistoricalProtocol"` -> 2 succeeded.
- [ ] CI to run the full client and spark suites.

## Companion PR
Server + protocol-doc half: forthcoming companion PR (will be linked here once opened).

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -89,7 +89,8 @@ trait DeltaSharingClient {
       table: Table,
       cdfOptions: Map[String, String],
       includeHistoricalMetadata: Boolean,
-      fileIdHash: Option[String]): DeltaTableFiles
+      fileIdHash: Option[String],
+      includeHistoricalProtocol: Boolean = false): DeltaTableFiles
 
   def generateTemporaryTableCredential(
     table: Table,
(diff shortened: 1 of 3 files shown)


littlegrasscao:
Companion server + protocol PR: #968

littlegrasscao:
Re-opening with the head branch on `delta-io/delta-sharing` (rather than the `littlegrasscao` fork) so the full CI suite, including the secret-gated integration tests, can run.

PR #968, opened 2026-06-25
[Server/PROTOCOL] Accept includeHistoricalProtocol on CDF (delta-format only)

littlegrasscao:
## Summary
Server-side half of `includeHistoricalProtocol` support for the CDF endpoint. Adds a new optional `includeHistoricalProtocol` query parameter to `GET .../changes` and plumbs it through `DeltaSharedTableProtocol.queryCDF`, `DeltaSharedTable.queryCDF`, and `DeltaSharingCDCReader.queryCDF`. The kernel implementation just carries the new arg (it still throws Unsupported).

The matching client-side wire encoding lands in a separate companion PR; this PR is fully usable on its own with a normal (older) client (the param defaults to `false`).

## Reader semantics
- `DeltaSharingCDCReader` emits intermediate `Protocol` actions seen in the delta log under the same `v > start` rule used today for `Metadata`, in all three CDF branches (CDC, hotfix metadata-only, AddFile/RemoveFile).
- `DeltaSharedTable.queryCDF` gates the entire feature on the response format being `delta`. For `parquet` responses the flag is silently dropped \u2014 the parquet wire shape has no representation for additional Protocol lines, so we keep it backwards compatible by never producing them.
- The new parameter is included in `QueryParamChecksum` so paginated queries with the flag don't collide with cached page tokens from requests without it.

## Docs
- `PROTOCOL.md`: the **Read Change Data Feed from a Table** section now documents `includeHistoricalProtocol` and explicitly notes that it only affects `responseformat=delta` responses.
- `delta-sharing-protocl-api-description.yml`: added the new query parameter to the OpenAPI spec.

## Backwards compatibility
- Existing clients don't send the parameter; the server treats it as `false`; behavior unchanged.
- For parquet responses the server silently ignores the flag.

## Test plan
- [x] `server` module compiles and test-compiles cleanly.
- [x] One server integration test added under `streaming_notnull_to_null` confirms the new query parameter is accepted on `/changes` and produces an unchanged response shape for a table with no historical Protocol upgrades. (Existing `streaming_notnull_to_null` integration tests need S3 credentials locally and only run in CI.)
- [ ] CI to run the full server suite against the new flag.

## Companion PR
Client-side wire encoding: forthcoming companion PR.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -2349,7 +2349,8 @@ Optional: `delta-sharing-capabilities: responseformat=delta;readerfeatures=delet
  **startingTimestamp** (type: String, optional): The starting timestamp of the query, a string in the [Timestamp Format](#timestamp-format), which will be converted to a version created greater or equal to this timestamp. <br>
  **endingVersion** (type: Long, optional): The ending version of the query, inclusive. <br>
  **endingTimestamp** (type: String, optional): The ending timestamp of the query, a string in the [Timestamp Format](#timestamp-format), which will be converted to a version created earlier than or at the timestamp. <br>
- **includeHistoricalMetadata** (type: Boolean, optional): If set to true, return the historical metadata if seen in the delta log. This is for the streaming client to check if the table schema is still read compatible.</td>
+ **includeHistoricalMetadata** (type: Boolean, optional): If set to true, return the historical metadata if seen in the delta log. This is for the streaming client to check if the table schema is still read compatible. <br>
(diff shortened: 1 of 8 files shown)


littlegrasscao:
Companion client PR: #967

littlegrasscao:
Re-opening with the head branch on `delta-io/delta-sharing` (rather than the `littlegrasscao` fork) so the full CI suite, including the secret-gated integration tests, can run.

PR #969, opened 2026-06-25
[Client] Add includeHistoricalProtocol param on getCDFFiles and streaming getFiles

littlegrasscao:
## Summary

Client-side half of `includeHistoricalProtocol` support for both delta-sharing endpoints that can stream a range of versions:

- `getCDFFiles` (`GET .../changes`) — flag is appended to the **URL query string** via `getEncodedCDFParams`, next to `includeHistoricalMetadata`.
- streaming `getFiles(table, startingVersion, endingVersion, fileIdHash)` (`POST .../query`) — flag is set on the **`QueryTableRequest` POST body** (it logically belongs alongside `startingVersion` / `endingVersion`).

Both paths get a new `includeHistoricalProtocol: Boolean = false` parameter on the `DeltaSharingClient` trait. The `TestDeltaSharingClient` mock implementing the trait is updated to match.

The matching server change (parse the new param, emit historical `Protocol` actions in delta-format responses, version-stamp each Protocol) is in companion PR #970.

### `getCDFFiles` (`GET .../changes`)

- `getEncodedCDFParams` appends `includeHistoricalProtocol=true` only when the caller opted in **and** the client's `responseFormat` includes `delta`.

### Streaming `getFiles` (`POST .../query`)

- `QueryTableRequest` case class gains `includeHistoricalProtocol: Option[Boolean] = None`. `getFiles(startingVersion, ...)` sets this to `Some(true)` only when the caller opted in **and** the client's `responseFormat` includes `delta`. The target URL is unchanged.

### Parquet-format gating (both endpoints)

For parquet-only clients the flag is a no-op on the server — the parquet response returns the same single Protocol regardless. The client suppresses the flag (URL on `/changes`, body field on `/query`) for parquet-only clients to keep parquet-format requests byte-identical to today's. Clients negotiated for `parquet,delta` still forward the flag, because the server can pick the delta-format response.

## Backwards compatibility

- Default is `false`, so existing call sites (`RemoteDeltaCDFRelation`, `DeltaSharingSource`, `RemoteDeltaLogSuite`, etc.) compile and behave unchanged.
- When the flag is `false`, both the URL and the POST body are byte-for-byte identical to today's request, so old servers see no difference.
- When the flag is `true` against an old server, the unknown URL query string parameter (on `/changes`) and the unknown body field (on `/query`) are simply ignored.
- `TestDeltaSharingClient` (spark-test mock) is updated to match the new signatures so the `spark` module still compiles.

## Tests

`testOnly io.delta.sharing.client.DeltaSharingRestClientSuite -- -z "includeHistoricalProtocol"` → **7 succeeded** locally:

- 2 for `getCDFFiles` (URL contains/doesn't contain `includeHistoricalProtocol=true`).
- 2 for streaming `getFiles` (body JSON contains/doesn't contain `"includeHistoricalProtocol":true`; URL never contains it).
- 1 for `getCDFFiles` parquet suppression (caller opts in, parquet-only client, URL does not contain the flag).
- 1 for streaming `getFiles` parquet suppression (caller opts in, parquet-only client, body does not contain the flag).
- 1 for `getCDFFiles` parquet+delta forwarding (caller opts in, `responseFormat=parquet,delta`, URL contains the flag).

## Companion PR

Server + protocol-doc half: #970.

(Supersedes #967, which was opened from a fork branch and therefore could not run secret-gated CI.)

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -161,8 +161,7 @@ private[sharing] case class QueryTableRequest(
   pageToken: Option[String],
   includeRefreshToken: Option[Boolean],
   refreshToken: Option[String],
-  idempotency_key: Option[String],
-  includeHistoricalProtocol: Option[Boolean] = None
+  idempotency_key: Option[String]
 ) extends NextPageRequest {
   override def clone(
         maxFiles: Option[Int],
@@ -553,8 +552,10 @@ class DeltaSharingRestClient(
     val encodedShareName = URLEncoder.encode(table.share, "UTF-8")
     val encodedSchemaName = URLEncoder.encode(table.schema, "UTF-8")
     val encodedTableName = URLEncoder.encode(table.name, "UTF-8")
+    val queryString = if (includeHistoricalProtocol) "?includeHistoricalProtocol=true" else ""
     val target = getTargetUrl(
-      s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/query")
+      s"/shares/$encodedShareName/schemas/$encodedSchemaName/tables/$encodedTableName/query" +
+        queryString)
     val request: QueryTableRequest = QueryTableRequest(
       predicateHints = Nil,
       limitHint = None,
@@ -567,8 +568,7 @@ class DeltaSharingRestClient(
       pageToken = None,
       includeRefreshToken = None,
       refreshToken = None,
-      idempotency_key = None,
-      includeHistoricalProtocol = if (includeHistoricalProtocol) Some(true) else None
+      idempotency_key = None
     )
 
     val (version, respondedFormat, lines, responseFileIdHash) = if (queryTablePaginationEnabled) {
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -83,7 +83,8 @@ trait DeltaSharingClient {
     table: Table,
     startingVersion: Long,
     endingVersion: Option[Long],
-    fileIdHash: Option[String]): DeltaTableFiles
+    fileIdHash: Option[String],
+    includeHistoricalProtocol: Boolean = false): DeltaTableFiles
 
   def getCDFFiles(
       table: Table,
@@ -160,7 +161,8 @@ private[sharing] case class QueryTableRequest(
   pageToken: Option[String],
   includeRefreshToken: Option[Boolean],
   refreshToken: Option[String],
-  idempotency_key: Option[String]
+  idempotency_key: Option[String],
+  includeHistoricalProtocol: Option[Boolean] = None
 ) extends NextPageRequest {
   override def clone(
         maxFiles: Option[Int],
@@ -545,7 +547,8 @@ class DeltaSharingRestClient(
       table: Table,
       startingVersion: Long,
       endingVersion: Option[Long],
-      fileIdHash: Option[String]): DeltaTableFiles = {
+      fileIdHash: Option[String],
+      includeHistoricalProtocol: Boolean = false): DeltaTableFiles = {
     val start = System.currentTimeMillis()
     val encodedShareName = URLEncoder.encode(table.share, "UTF-8")
     val encodedSchemaName = URLEncoder.encode(table.schema, "UTF-8")
@@ -564,7 +567,8 @@ class DeltaSharingRestClient(
       pageToken = None,
       includeRefreshToken = None,
       refreshToken = None,
-      idempotency_key = None
+      idempotency_key = None,
+      includeHistoricalProtocol = if (includeHistoricalProtocol) Some(true) else None
     )
 
     val (version, respondedFormat, lines, responseFileIdHash) = if (queryTablePaginationEnabled) {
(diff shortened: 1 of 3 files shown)


littlegrasscao:
Companion server + protocol PR: #970

littlegrasscao:
Extended this PR to also cover the streaming `getFiles(table, startingVersion, endingVersion, ...)` path. `includeHistoricalProtocol` is now a new optional field on `QueryTableRequest` that the client only sets when true (`Include.NON_ABSENT` keeps the wire byte-identical when not set). See commit 0c4c10765.

littlegrasscao:
Per feedback, moved the streaming `getFiles` half off the `QueryTableRequest` body and onto a URL query string parameter, so the wire encoding matches `getEncodedCDFParams` on `/changes`.

- `QueryTableRequest` case class: dropped the `includeHistoricalProtocol: Option[Boolean]` field added in the previous commit.
- Streaming `getFiles`: when `includeHistoricalProtocol = true`, the `/query` target URL is suffixed with `?includeHistoricalProtocol=true`; otherwise both the URL and the POST body are byte-identical to today's.
- Trait signatures (`getFiles`, `TestDeltaSharingClient`) unchanged from the previous commit.
- Unit tests now capture the target URL passed to `getNDJsonPost` and assert: when enabled, URL has the param and body does NOT mention it; when disabled, neither URL nor body mentions it. All 4 unit tests (2 CDF + 2 streaming) pass locally.

See commit 5b87cf1b5. Companion server change in #970.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -83,13 +83,15 @@ trait DeltaSharingClient {
     table: Table,
     startingVersion: Long,
     endingVersion: Option[Long],
-    fileIdHash: Option[String]): DeltaTableFiles
+    fileIdHash: Option[String],
+    includeHistoricalProtocol: Boolean = false): DeltaTableFiles
```
linzhou-db, line 87: I'm thinking who is the caller and how does the caller decide whether to set this to true or false.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -544,13 +546,16 @@ class DeltaSharingRestClient(
       table: Table,
       startingVersion: Long,
       endingVersion: Option[Long],
-      fileIdHash: Option[String]): DeltaTableFiles = {
+      fileIdHash: Option[String],
+      includeHistoricalProtocol: Boolean = false): DeltaTableFiles = {
```
linzhou-db, line 551: once `includeHistoricalProtocol` is set to true, will the server return a separate line for `protocol` when it appears? 
And is that new line in the response being handled?

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: The caller would be DBR/Delta OSS. Only streaming and CDF caller would set includeHistoricalProtocol=true. 

In this way, we can only set it to true only in supported DBR versions

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: I am planning to do:
1. only apply to delta format response
2. only when client send includeHistoricalProtocol, which means client is able to handle reponse
3. delta format response are lines. So we will have 1st line: version -> initial protocol; followed by metadata/addfile; if there is a protocol change in a mid-version, new line: version -> second protocol is sending back. 

This OSS server PR would demonstrate that better:
https://github.com/delta-io/delta-sharing/pull/970/changes

linzhou-db: "This only matters for delta responseFormat," -- how is this defined and enforced?

if you plan to define it to be only applied to delta format, then should we only set it when responseFormat=delta is set in the header?

littlegrasscao: On the client side, we will only set includeHistoricalProtocol=true in runtime, that enforces delta format.
Server knows response format and can handle accordingly.

littlegrasscao: updated the logic to check both includeHistoricalProtocol input and responseFormatSet contains delta

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: I see so it mostly depend on the use case to decide the true/false, instead of flag.

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: > 3. delta format response are lines. So we will have 1st line: version -> initial protocol; followed by metadata/addfile; if there is a protocol change in a mid-version, new line: version -> second protocol is sending back.

Is this code already in place?

littlegrasscao: number 3 is server side response - see server impl 

How to write new lines into block manager will be implemented in Runtime/Delta OSS

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
littlegrasscao: I will also add a flag in runtime/Delta OSS, so we check both use case and flag

client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
linzhou-db: right, sg.
I just confirmed that lines are directly returned and not parsed in DeltaSharingClient through `return DeltaTableFiles(`

linzhou-db (approved):

PR #970, opened 2026-06-25
[Server/PROTOCOL] Accept includeHistoricalProtocol on CDF and streaming /query (delta-format only)

littlegrasscao:
## Summary

Server + PROTOCOL.md half of `includeHistoricalProtocol` support for both delta-sharing endpoints that can stream a range of versions:

- `GET .../changes` (CDF) — flag is carried as a **URL query string** parameter.
- `POST .../query` with `startingVersion` (streaming version range) — flag is carried in the **`QueryTableRequest` POST body** (it logically belongs alongside `startingVersion` / `endingVersion`).

The flag is plumbed through `DeltaSharedTableProtocol`, `DeltaSharedTable`, and `DeltaSharingCDCReader`. The kernel implementations just carry the new arg (behaviorally unchanged on the kernel paths). On `responseformat=delta` responses, `Protocol` actions carry the delta-log `version` they correspond to, mirroring how `Metadata` already does it; this version stamping is gated on the same `includeHistoricalProtocol` flag.

The matching client-side wire encoding lands in companion PR #969. This PR is fully usable on its own with an older client (the flag defaults to `false` everywhere and is purely additive).

### `GET .../changes` (CDF)

- New `@Param("includeHistoricalProtocol") @Nullable includeHistoricalProtocol: String` on `listCdfFiles`, parsed via `Try(_.toBoolean).getOrElse(false)`, mirroring the existing `includeHistoricalMetadata` URL param.

### `POST .../query` (streaming version range)

- `QueryTableRequest.proto` gains `optional bool includeHistoricalProtocol = 16`. `DeltaSharingService.listFiles` reads `request.includeHistoricalProtocol.getOrElse(false)` and forwards it to both `query(...)` call sites (the kernel and non-kernel paths). No new URL parameters; the target URL is unchanged for clients that don't set the field.

### Delta-format Protocol wrappers

- `DeltaResponseProtocol` (standalone path) and `DeltaFormatResponseProtocol` (kernel path) gain an optional `version: java.lang.Long = null` field. Jackson's `Include.NON_ABSENT` keeps the field off the wire when it's `null`.
- `version` is populated **only when `includeHistoricalProtocol=true`** and the response is in delta format. Both the head Protocol (stamped with `startingVersion` for streaming `/query` and `snapshot.version` for `/changes`) and any inlined historical Protocols (stamped with their actual commit version) carry it. Older clients that don't opt in see byte-identical delta-format wire shape.
- The parquet `model.Protocol` is **not** changed — parquet responses only ever carry the head Protocol with no version, so the parquet wire shape is preserved.

## Reader semantics

- `DeltaSharingCDCReader.queryCDF` emits intermediate `Protocol` actions seen in the delta log under the same `v > start` rule used today for `Metadata`, in all three CDF branches (CDC, hotfix metadata-only, AddFile/RemoveFile).
- `DeltaSharedTable.queryDataChangeSinceStartVersion` (called from streaming `/query`) emits intermediate `Protocol` actions seen in the log under the same `v > startingVersion` rule used today for `Metadata`.
- The new flag is included in `QueryParamChecksum` so paginated queries with the flag don't collide with cached page tokens from requests without it.
- `includeHistoricalProtocol` is gated on `responseFormat == delta`; for parquet responses the flag is a no-op and no additional Protocol lines (or version stamps) are emitted.
- Internal helpers (`queryDataChangeSinceStartVersion`, `DeltaSharingCDCReader.queryCDF`) receive the value as `emitHistoricalProtocol` to make it visually clear that the response-format gate has already been applied by the caller.

## Docs

- `PROTOCOL.md`:
  - **Read Change Data Feed from a Table** lists `includeHistoricalProtocol` under Query Parameters.
  - **Read Data from a Table** documents `includeHistoricalProtocol` under the request body bullets (where `startingVersion` / `endingVersion` already live).
  - **Protocol in Delta Format** documents the new optional `version` field with a comprehensive JSON example showing `readerFeatures` and `writerFeatures`.
  - Both endpoint descriptions note that the flag only affects `responseformat=delta` responses.
- `delta-sharing-protocl-api-description.yml`:
  - `GET /changes` adds `includeHistoricalProtocol` as an `in: query` parameter alongside `includeHistoricalMetadata`.
  - `POST /query` keeps `parameters: *tableParameters` (no new URL params); `includeHistoricalProtocol` is added to the `QueryRequest` body schema.

## Backwards compatibility

- Existing clients don't send the flag on either endpoint; the server treats it as `false`; behavior is unchanged.
- For parquet responses the server silently ignores the flag on both endpoints; the parquet `Protocol` case class is unchanged.
- Delta-format responses are byte-identical to today's for clients that don't opt in — the head Protocol's `version` field is gated on the flag, so legacy delta-format clients that assert on exact response shape continue to pass.

## Tests

- New unit suites that hand-craft a tiny local Delta log with a Protocol upgrade at `v=2` and assert on the wrapped `DeltaFormatResponseProtocol` objects via JSON round-trip:
  - `DeltaSharedTableCDFSuite` (3 tests): head Protocol carries no version when flag is off; head + intermediate Protocol@v=2 each stamped with its delta-log version when flag is on; starting from the upgrade version itself emits only the head (`v > start` gate).
  - `DeltaSharedTableVersionRangeSuite` (3 tests): same coverage for streaming `query` in version-range mode.
  - Both share `DeltaSharedTableTestBase`, which adds a `protected def newFileSigner()` seam on `DeltaSharedTable` so the suite can run against the local Delta log with a no-op signer.
- All 6 unit tests pass locally; scalastyle is clean.

## Companion PR

Client-side wire encoding: #969.

(Supersedes #968, which was opened from a fork branch and therefore could not run secret-gated CI.)


littlegrasscao:
Companion client PR: #969

littlegrasscao:
Extended this PR to also cover the streaming `POST .../query` path. New `optional bool includeHistoricalProtocol = 16` on `QueryTableRequest.proto`; `DeltaSharedTable.query` now plumbs the flag through `queryDataChangeSinceStartVersion` to emit historical `Protocol` actions inline (delta-format only). PROTOCOL.md updated to document the new field next to `startingVersion`. See commit b613a0a42.

littlegrasscao:
Per feedback, moved the streaming `/query` half off the request body and onto a URL query string parameter, so the wire encoding matches the `/changes` endpoint added in the first commit.

- `QueryTableRequest.proto`: removed `optional bool includeHistoricalProtocol = 16` (no proto field).
- `DeltaSharingService.listFiles`: now reads the flag from `@Param("includeHistoricalProtocol") @Nullable includeHistoricalProtocol: String` using the same `Try(_.toBoolean).getOrElse(false)` pattern as `listCdfFiles`, and forwards it to both `query(...)` call sites.
- Internal reader plumbing (`DeltaSharedTable.query`, `DeltaSharedTableProtocol.query`, `DeltaSharedTableKernel.query`, `queryDataChangeSinceStartVersion`, `QueryParamChecksum`) is unchanged — only how the flag is carried across HTTP changes.
- `PROTOCOL.md`: under `Read Data from a Table`, removed the body-bullet and added a new `Query Parameters` row.
- `delta-sharing-protocl-api-description.yml`: the `POST /query` operation now lists `includeHistoricalProtocol` as a query parameter.
- Integration test updated to send `?includeHistoricalProtocol=true` in the URL instead of a body field.

See commit 1207c3e13. Companion client change in #969.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTable.scala
```diff
@@ -644,6 +666,9 @@ class DeltaSharedTable(
             numSignedFiles += 1
           case (p: Protocol, _) =>
             assertProtocolRead(p)
+            if (includeHistoricalProtocol && v > startingVersion) {
```
linzhou-db: we don't need to check `responseFormat` here?

server/src/main/scala/io/delta/standalone/internal/DeltaSharingCDCReader.scala
```diff
@@ -255,6 +262,8 @@ class DeltaSharingCDCReader(val deltaLog: DeltaLogImpl, val conf: Configuration)
               selectedActions.append(c)
             case m: Metadata if (includeHistoricalMetadata && v > start) =>
               selectedActions.append(m)
+            case p: Protocol if (includeHistoricalProtocol && v > start) =>
```
linzhou-db: we don't need to check `responseFormat` here?

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTable.scala
littlegrasscao: The caller of this method queryDataChangeSinceStartVersion already checks responseFormat when pass in includeHistoricalProtocol at line 429

server/src/main/scala/io/delta/standalone/internal/DeltaSharingCDCReader.scala
littlegrasscao: the caller at DeltaSharedTable.scala line 747 already checks responseFormat before pass in includeHistoricalProtocol

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTable.scala
linzhou-db: then we should use `emitHistoricalProtocol`?

server/src/main/scala/io/delta/standalone/internal/DeltaSharingCDCReader.scala
linzhou-db: then we should use `emitHistoricalProtocol`?

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTable.scala
littlegrasscao: updated the param name to make it more clear.

server/src/main/scala/io/delta/standalone/internal/DeltaSharingCDCReader.scala
littlegrasscao: updated the param name to make it more clear.

server/src/main/scala/io/delta/standalone/internal/DeltaSharedTable.scala
```diff
@@ -91,7 +92,11 @@ class DeltaSharedTable(
     }
   }
 
-  private val fileSigner = withClassLoader {
+  private val fileSigner: CloudFileSigner = withClassLoader { newFileSigner() }
+
+  // Extracted so tests can override with a no-op signer when running against a local Delta log.
```
linzhou-db, line 97: why do we want to use a no-op signer in tests?

server/src/test/scala/io/delta/standalone/internal/DeltaSharedTableTestBase.scala
```diff
@@ -0,0 +1,163 @@
...
+      partitionValues = Map.empty,
+      size = 100L,
+      modificationTime = 1700000004000L,
+      dataChange = true)))
+
+    tableDir
+  }
+
+  /** Write a single Delta commit at `version` as NDJSON action lines. */
+  private def writeCommit(logDir: File, version: Long, actions: Seq[Action]): Unit = {
```
linzhou-db, line 94: interesting to see that we are using this to produce a fake delta log for testing.

linzhou-db (approved):
Looking good! 

Do we have a real table with different protocol in different versions, so we can test the response for client side changes as well?

PR #983, opened 2026-07-07
Add UnsupportedOp for unsupported predicate sub-expressions

chakankardb:
Previously OpConverter threw IllegalArgumentException for unsupported SQL expressions (e.g. nested struct field access via GetStructField), causing the entire predicate to be dropped. This prevented pruning even when other parts of the predicate (e.g. AND siblings) were fully supported.

Instead, this change uses UnsupportedOp for unsupported sub-expressions, and then prunes away parts of expression tree based on them. This preserves the compound predicate structure so that supported siblings can still drive file pruning. UnsupportedOp evaluates conservatively to true (never prunes), which is propagated up via AND/OR/NOT. The UnsupportOp stays local to this client, and is never sent back to the server.

The change is guarded behind a flag.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
+++ b/client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
@@ -62,34 +62,45 @@ object OpConverter {
   // Converts a set of SQL expression trees into a json predicate op.
   // We perform an implicit And across the input set at the top level.
   //
-  // The method throws an exception if the conversion fails, and callers
-  // should skip filtering in that case.
+  // If partialFilterEnabled is true, unsupported sub-expressions are replaced with UnsupportedOp,
+  // allowing the rest of the predicate tree to still be used for pruning.
+  // If false (default), throws IllegalArgumentException on any unsupported expression.
   @throws[IllegalArgumentException]
-  def convert(expressions: Seq[SqlExpression]): Option[BaseOp] = {
-    expressions.map(convertOne(_)) match {
+  def convert(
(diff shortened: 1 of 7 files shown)


client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
```diff
@@ -62,34 +62,45 @@ object OpConverter {
...
-  //
-  // Throws an exception if conversion fails.
-  private def convertOne(expr: SqlExpression): BaseOp = {
-    // Try converting to a leaf op first, and it that fails try converting to
+  // Converts one SQL expression, returning UnsupportedOp for any unsupported sub-expression
+  // when partialFilterEnabled is true, or throwing IllegalArgumentException when false.
+  private def convertOne(expr: SqlExpression, partialFilterEnabled: Boolean): BaseOp = {
+    if (partialFilterEnabled) {
+      try { convertOneInternal(expr, partialFilterEnabled) }
+      catch { case _: IllegalArgumentException => UnsupportedOp() }
```
linzhou-db: nit: do we want to log the exception?

linzhou-db (dismissed):

client/src/main/scala/io/delta/sharing/filters/OpConverter.scala
chakankardb: Done.

linzhou-db (approved):

charlenelyu-db (approved):
LGTM

PR #990, opened 2026-07-14
[Client] Support backtick-quoted share/schema/table names with dots in parsePath

### LINKED ISSUE #646: Unable to load tables with delta sharing client if share has a dot (`.`) in its name
Loading a table using the delta sharing client fails if the share, schema, or table contains a dot (`.`) in their names.

```
>>> spark.read.format("deltaSharing").load("file:///tmp/profile.json#`my.share`.`my.schema`.`my.table`")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/.../py-envs/spark-3.5.1/lib/python3.11/site-packages/pyspark/sql/readwriter.py", line 307, in load
    return self._df(self._jreader.load(path))
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../py-envs/spark-3.5.1/lib/python3.11/site-packages/pyspark/python/lib/py4j-0.10.9.7-src.zip/py4j/java_gateway.py", line 1322, in __call__
  File "/Users/.../py-envs/spark-3.5.1/lib/python3.11/site-packages/pyspark/errors/exceptions/captured.py", line 185, in deco
    raise converted from None
pyspark.errors.exceptions.captured.IllegalArgumentException: path file:///tmp/profile.json#`my.share`.`my.schema`.`my.table` is not valid
```

This issue can be reproduced on the latest delta sharing client version.

The problem comes from path parsing implemented [here](https://github.com/delta-io/delta-sharing/blob/3481ef28452a07f6372ced1c880cab538b92d85d/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala#L1313-L1316), which splits the path on `.` characters, regardless of escaping.

The delta sharing API does not have any restrictions on share, schema, or table names, aside from mentioning they are case-insensitive.
So the expectation is that some sort of escaping could be applied to allow targeting shares with `.` in the name.

### Example JVM stacktrace

```
java.lang.IllegalArgumentException: path file:///tmp/profile.json#`my.share`.`my.schema`.`my.table` is not valid
	at io.delta.sharing.client.DeltaSharingRestClient$.parsePath(DeltaSharingClient.scala:970)
	at io.delta.sharing.spark.DeltaSharingDataSource.autoResolveBaseRelationForSnapshotQuery(DeltaSharingDataSource.scala:317)
	at io.delta.sharing.spark.DeltaSharingDataSource.createRelation(DeltaSharingDataSource.scala:237)
	at org.apache.spark.sql.execution.datasources.DataSource.resolveRelation(DataSource.scala:346)
	at org.apache.spark.sql.DataFrameReader.loadV1Source(DataFrameReader.scala:229)
	at org.apache.spark.sql.DataFrameReader.$anonfun$load$2(DataFrameReader.scala:211)
	at scala.Option.getOrElse(Option.scala:189)
	at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:211)
	at org.apache.spark.sql.DataFrameReader.load(DataFrameReader.scala:186)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:244)
	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:374)
	at py4j.Gateway.invoke(Gateway.java:282)
	at py4j.commands.AbstractCommand.invokeMethod(AbstractCommand.java:132)
	at py4j.commands.CallCommand.execute(CallCommand.java:79)
	at py4j.ClientServerConnection.waitForCommands(ClientServerConnection.java:182)
	at py4j.ClientServerConnection.run(ClientServerConnection.java:106)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

anxkhn:
Fixes #646.

- `DeltaSharingRestClient.parsePath` splits the `share.schema.table` identifier on every `.` via `tablePath.split("\\.", -1)` and then requires exactly three parts. As a result, any share, schema, or table whose name legitimately contains a dot cannot be loaded: for example `spark.read.format("deltaSharing").load("<profile>#`my.share`.`my.schema`.`my.table`")` throws `IllegalArgumentException: path ... is not valid` (issue #646), and there is no way to escape the dot.
- Parse the three-part identifier honoring backtick quoting instead of a naive split. Each part may be wrapped in backticks so its name can contain dots; a literal backtick inside a quoted part is escaped by doubling it. This mirrors how identifiers are already quoted elsewhere in this client (`RemoteDeltaCDFRelation.quoteIdentifier` builds ``s"`${part.replace("`", "``")}`"``), so the new parser is the inverse of an escaping scheme the codebase already uses, and it also matches how Spark/Delta quote identifiers.
- Unquoted input is split exactly like `split("\\.", -1)`, so all existing behavior and every existing validation error (wrong number of parts, empty part, missing profile) is preserved. Malformed quoting (an unterminated quote, or stray characters after a closing backtick) produces the same `path ... is not valid` error already used for bad input.

## Behavior

Given a profile file `<profile>`:

| Path | Before | After |
| --- | --- | --- |
| `<profile>#share.schema.table` | share=`share`, schema=`schema`, table=`table` | unchanged |
| ``<profile>#`my.share`.`my.schema`.`my.table` `` | `IllegalArgumentException` | share=`my.share`, schema=`my.schema`, table=`my.table` |
| ``<profile>#`my.share`.schema.table `` | `IllegalArgumentException` | share=`my.share`, schema=`schema`, table=`table` |
| ``<profile>#`a``b`.c.d `` | `IllegalArgumentException` | share=`` a`b ``, schema=`c`, table=`d` |

The backticks are optional: existing unquoted three-part paths keep working exactly as before.

## Test plan

- [x] `build/sbt "client/testOnly io.delta.sharing.client.DeltaSharingRestClientSuite -- -z parsePath"` - the extended `parsePath` test passes. New assertions cover quoted names containing dots, mixed quoted and unquoted parts, an escaped (doubled) backtick, and the malformed-quote error cases (unterminated quote, characters after a closing backtick, quoted empty name). `parsePath` is a plain unit test, so it runs without cloud credentials or a live server.
- [x] Verified this is a genuine regression test: reverting the one-line `parsePath` change back to `split("\\.", -1)` makes the new assertions fail with the exact issue #646 error, then passes again with the fix in place.
- [x] `build/sbt "client/compile"` and the `++2.12.18` cross-build both compile cleanly (Scala 2.13 and 2.12).
- [x] `build/sbt "client / Test / scalastyle"` (main and test) reports no errors or warnings.

## Notes

- Scope is the Scala client only. The Python client has the same limitation in its own path parsing; I have left it out to keep this change surgical and would be happy to follow up in a separate PR.
- The chosen escape character is the backtick, matching this client's existing `quoteIdentifier` and the escaping the reporter suggested in #646. Flagging it here in case the maintainers would prefer a different convention.

### DIFF
--- a/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
+++ b/client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
@@ -1621,7 +1621,9 @@ object DeltaSharingRestClient extends Logging {
       }
     }
 
-    val tableSplits = tablePath.split("\\.", -1)
+    val tableSplits = splitTablePath(tablePath).getOrElse {
+      throw new IllegalArgumentException(s"path $path is not valid")
+    }
     if (tableSplits.length != 3) {
       throw new IllegalArgumentException(s"path $path is not valid")
     }
@@ -1635,6 +1637,67 @@ object DeltaSharingRestClient extends Logging {
       schema = tableSplits(1),
       table = tableSplits(2)
     )
+  }
+
+  /**
+   * Split a `share.schema.table` identifier on unquoted dots. Each part may be quoted with
+   * backticks so that its name can contain dots, matching how Delta/Spark quotes identifiers:
+   * the part is wrapped in backticks and any backtick inside it is escaped by doubling it.
+   * The returned parts have their surrounding backticks stripped and doubled backticks unescaped.
+   *
+   * Returns None if the identifier is malformed (an unterminated quote, or a quoted part with
+   * unexpected characters outside its backticks); the caller turns this into the same
+   * "path is not valid" error used for the unquoted case. Unquoted input is split exactly like
+   * `split("\\.", -1)`, so existing behavior is preserved.
+   */
+  private def splitTablePath(tablePath: String): Option[Seq[String]] = {
+    val parts = scala.collection.mutable.ArrayBuffer.empty[String]
+    val current = new StringBuilder
+    // Tracks the state of the part currently being read: whether we are inside a backtick-quoted
+    // region, and whether a closing backtick has been seen (after which only a dot may follow).
+    var inQuotes = false
+    var quotedClosed = false
+    var i = 0
+    val len = tablePath.length
+    while (i < len) {
+      val c = tablePath.charAt(i)
+      if (inQuotes) {
+        if (c == '`') {
+          if (i + 1 < len && tablePath.charAt(i + 1) == '`') {
+            // An escaped backtick inside the quoted name.
+            current.append('`')
+            i += 1
+          } else {
+            inQuotes = false
+            quotedClosed = true
+          }
+        } else {
+          current.append(c)
+        }
+      } else if (c == '.') {
+        parts += current.toString
+        current.setLength(0)
+        quotedClosed = false
+      } else if (c == '`') {
+        // A backtick may only open a quote at the start of a part.
+        if (current.nonEmpty || quotedClosed) {
+          return None
+        }
+        inQuotes = true
+      } else if (quotedClosed) {
+        // Non-dot characters are not allowed after a closing backtick, e.g. `a`b.
+        return None
+      } else {
+        current.append(c)
+      }
+      i += 1
+    }
+    if (inQuotes) {
+      // Unterminated quote, e.g. `a.b.c.
+      return None
+    }
+    parts += current.toString
+    Some(parts.toSeq)
   }
 
   private def tryParseEndStreamAction(line: String): EndStreamAction = {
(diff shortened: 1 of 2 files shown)


anxkhn:
@linzhou-db when you have a chance, could you please take a look at this pr and let me know if any update is needed?

linzhou-db:
@anxkhn  the ping helped. 

Added @kevinjia-db  and @chakankardb  as reviewer

PR #999, opened 2026-08-04
Fix Get Query Info protocol documentation

vrinda-db:
## What changes are proposed?

Align the Get Query Info documentation with the implemented table-scoped endpoint and lowercase `pending`/`failed` status values. Update the URL parameters and all affected examples in the same section.

## How was this patch tested?

Rendered `PROTOCOL.md` with `marked`, verified the Get Query Info HTML table/code-fence/details tags are balanced, and checked that no stale unscoped endpoint or uppercase async status examples remain.


vrinda-db:
Superseded by https://github.com/delta-io/delta-sharing/pull/1000, which uses a branch in the upstream repository.

PR #1003, opened 2026-08-27
Log first k lines on unexpected snapshot getFiles actions

littlegrasscao:
## Summary
- Snapshot `getFiles` now calls `logFirstKLines` before throwing on an unexpected NDJSON action, matching CDF `getFiles` and `getMetadata`.
- This dumps the first 10 response lines in the error log so unexpected parquet-format query responses are easier to diagnose.

## Test plan
- [ ] Confirm snapshot `getFiles` still succeeds on a normal parquet query response (protocol + metadata + add files).
- [ ] Confirm an unexpected action line logs the first 10 lines via `logFirstKLines` then throws `IllegalStateException`.


client/src/main/scala/io/delta/sharing/client/DeltaSharingClient.scala
```diff
@@ -530,7 +531,15 @@ class DeltaSharingRestClient(
...
       } else {
-        throw new IllegalStateException(s"Unexpected Line:${line}" + getDsQueryIdForLogging)
+        logFirstKLines(
+          lines,
+          s"getFiles(versionAsOf-$versionAsOf, timestampAsOf-$timestampAsOf)",
+          10,
+          target
+        )
+        throw new IllegalStateException(
+          s"Unexpected Line:${line}. From the 3rd line expect file actions only. " +
```
chakankardb: Nit: perhaps remove "From the 3rd line..."; just say "Expecting file actions but got ...."

littlegrasscao: updated error message

chakankardb (approved):

